PR #4, opened 2019-10-17
Add substrate submodule

geigerzaehler:
Instead of using `substrate` as a git dependency we include it as a submodule. This makes it easier to modify the code and update it from upstream.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "substrate"]
+	path = substrate
+	url = https://github.com/paritytech/substrate.git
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "substrate"]
+	path = substrate
+	url = git@github.com:paritytech/substrate.git
```
kim: I have avoided generating ssh keys for the build agent user, and I'm not actually sure if the public key has to be pre-shared with GH even for public repos. Any concerns against using a `https://` URL?

geigerzaehler: Ah yes. I’ll use `https://` instead.

PR #9, opened 2019-10-21
Fix `clippy::all` warnings in `substrate`

rockbmb:
@geigerzaehler WIP, we may decide it's not at all worth it to keep `clippy::all` in CI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "substrate"]
+	path = substrate
+	url = https://github.com/rockbmb/substrate.git
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -80,5 +80,5 @@ gcr.io/opensourcecoin/radicle-registry       <none>                             
 ```
 
 The new image digest in this case will be
-sha256:264a6e976faeb63adda312cd3ecf2fb9ff71fd6f030128b3f012daf9c4cff05b, as it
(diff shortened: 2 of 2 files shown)


geigerzaehler:
I think it only makes sense to enable clippy for the `substrate` code base if they merge these fixes upstream and commit to use it upstream. Otherwise we are in a position where we need to maintain our fixed fork and fix their lints.

To check only our packages we should run `clippy` without the `--all` flag and select our packages explicitly with `-p`.

geigerzaehler:
> To check only our packages we should run `clippy` without the `--all` flag and select our packages explicitly with `-p`.

Turns out that this does not work. Investigating further...

rockbmb:
@geigerzaehler I forgot to mention this - as part of this PR I intend to submit these changes upstream, and request that `clippy` be added to their CI pipeline.

I saw that in this repo there's a `ci-test-lint` branch where you try out `clippy -D clippy::all`, so I thought I'd try adding that here as well.

geigerzaehler:
@rockbmb Can we delay merging this PR until the changes are merged upstream? For now I’d prefer to disable clippy instead of having to maintain a fork of substrate.

rockbmb:
@geigerzaehler ✔️

rockbmb:
#18 makes this redundant.

PR #14, opened 2019-10-22
Remove grandpa

geigerzaehler:
Completely remove grandpa consensus and only use Aura

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -23,7 +23,6 @@ use substrate_primitives::{crypto::key_types, ed25519, OpaqueMetadata};
 #[cfg(feature = "std")]
 use sr_version::NativeVersion;
 use sr_version::RuntimeVersion;
-use srml_grandpa::{fg_primitives, AuthorityId as GrandpaId, AuthorityWeight as GrandpaWeight};
 use substrate_client::{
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
```diff
@@ -202,7 +162,8 @@ impl srml_indices::Trait for Runtime {
 }
 
 parameter_types! {
-    pub const MinimumPeriod: u64 = SLOT_DURATION / 2;
+    /// Minimum time between blocks in milliseconds
+    pub const MinimumPeriod: u64 = 300;
```
rockbmb: Where did `SLOT_DURATION` go? It's probably a good idea to keep this constant from BABE, or at least name a new one.

Also, did you mean `3000` here? This results in a block at least every 0.3 seconds.

geigerzaehler: > Where did `SLOT_DURATION` go? It's probably a good idea to keep this constant from BABE, or at least name a new one.

Using `SLOT_DURATION` just adds another level of indirection. No reason to keep Babe stuff around if we replaced it.



> Also, did you mean `3000` here? This results in a block at least every 0.3 seconds.

300 is what I meant. We want fast feedback if we submit a transaction.

PR #20, opened 2019-10-24
Add basic client

geigerzaehler:
Add basic client with end to end tests for counter.


.circleci/config.yml
```diff
@@ -44,28 +44,37 @@ jobs:
```
rockbmb: In a previous PR I remember seeing this file's pipeline replaced with `true` so that it wouldn't do anything anymore, was I confusing things?

ci/run
```diff
@@ -43,9 +43,16 @@ export RUSTFLAGS="-D warnings"
 cargo build --all --all-targets --release
 
 echo "--- cargo test"
+echo "Starting radicle_registry_node"
+RUST_LOG=error ./target/release/radicle_registry_node \
+  --dev \
+  --base-path /tmp/radicle-registry \
+  &
+registry_node_pid=$!
```
rockbmb: Does this mean that appending a `&` to a command makes it return the PID of the process running it in the background?

client/examples/counter.rs
```diff
@@ -0,0 +1,21 @@
...
+use radicle_registry_client::{ed25519, Client, Error};
+use substrate_primitives::crypto::Pair;
+
+fn main() {
+    env_logger::init();
+    let mut runtime = tokio::runtime::Runtime::new().unwrap();
+    runtime.block_on(Compat::new(go().boxed())).unwrap();
+}
+
+async fn go() -> Result<(), Error> {
```
rockbmb: I didn't know this syntax was already supported, that's good.

client/src/base.rs
```diff
@@ -0,0 +1,54 @@
...
+        &self,
+        key_pair: &ed25519::Pair,
+        call: impl Into<Call>,
+    ) -> impl Future<Item = ExtrinsicSuccess, Error = Error> {
+        self.subxt_client
+            .xt(key_pair.clone(), None)
+            .and_then(move |xt_builder| xt_builder.set_system_call(call).submit_and_watch())
+    }
+
+    #[allow(dead_code)]
```
rockbmb: I see, this is indeed similar to the prototype. Is it necessary to have it even though it's `dead_code`?

client/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+            .map(|_| ())
+    }
+
+    pub fn free_balance(
+        &self,
+        account_id: &ed25519::Public,
+    ) -> impl Future<Item = Balance, Error = Error> {
+        self.base_client
+            .subxt_client
+            .free_balance(account_id.clone())
```
rockbmb: `AccountId = [u8; 20]`, such types implement `Copy` and as such do not need to be `clone()`d.
No need to fix this, `clippy` can be run later.

.circleci/config.yml
geigerzaehler: I only used `true` when I was developing the PR. Since we don’t have status updates yet, I’d like to keep CircleCI for now

ci/run
geigerzaehler: The command doesn’t quite “return” the PID but it is saved in `$!`

client/examples/counter.rs
geigerzaehler: It is in nightly

client/src/base.rs
geigerzaehler: I’ll remove it since it is not necessary anymore

client/src/lib.rs
geigerzaehler: We are not using our own definition of `AccountId` anymore. If you look at `runtime/src/lib.rs` you can see that `AccountId` is now a public key and does not implement `Copy` anymore.

ci/run
rockbmb: Nevermind - I confused `$?` with `$!`.

PR #21, opened 2019-10-25
Make CI base image smaller and add 'make' dependency

geigerzaehler:
* We make the CI base image smaller by not basing it on `rust:1.38.0` which includes the whole 1.38.0 toolchain but installing Rust from scratch. For this we copy the code from the `rust` dockerfiles. This reduces the image size from 3.15GB to 2.03GB.
* We add `make` to the list of dependencies. This is required by some new rust dependencies of the client.
* We simplify the documentation by pointing out that the digest is printed by `docker push`.

### DIFF
--- a/ci/base-image/Dockerfile
+++ b/ci/base-image/Dockerfile
@@ -1,36 +1,76 @@
-FROM rust:1.38.0-slim-buster
+# Image to run builds of the project
+#
+# Includes
+#
+# * Rustup with `nightly-2019-09-05` toolchain (see RUST_VERSION)
+# * Additional rustup components and the wasm32 target
+# * wasm-gc
+# * sccache v0.2.12
 
-# Install additional packages
+FROM debian:buster-slim
+
+SHELL ["/bin/bash", "-c"]
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH \
+    RUST_VERSION=nightly-2019-09-05 \
+    SCCACHE_DIR=/cache/sccache \
+    RUSTC_WRAPPER=sccache \
+    CARGO_HOME=/usr/local/cargo
+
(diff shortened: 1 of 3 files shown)


ci/base-image/Dockerfile
```diff
@@ -1,36 +1,76 @@
...
-# linting.
-RUN rustup default nightly-2019-09-05 \
-    && rustup update \
-    && rustup component add rustfmt \
+# Install sccache
+RUN set -euxo pipefail; \
+    sccache_version=0.2.12; \
+    sccache_base=sccache-$sccache_version-x86_64-unknown-linux-musl; \
+    wget --quiet https://github.com/mozilla/sccache/releases/download/$sccache_version/$sccache_base.tar.gz; \
+    echo "26fd04c1273952cc2a0f359a71c8a1857137f0ee3634058b3f4a63b69fc8eb7f  $sccache_base.tar.gz" | sha256sum -c -; \
```
rockbmb: I forgot to point it out to you, but last time you didn't have the `-c` switch on `sha256sum`, meaning it didn't do what it was expected to.

ci/base-image/Dockerfile
```diff
@@ -1,36 +1,76 @@
...
+
+SHELL ["/bin/bash", "-c"]
+
+ENV RUSTUP_HOME=/usr/local/rustup \
+    CARGO_HOME=/usr/local/cargo \
+    PATH=/usr/local/cargo/bin:$PATH \
+    RUST_VERSION=nightly-2019-09-05 \
+    SCCACHE_DIR=/cache/sccache \
+    RUSTC_WRAPPER=sccache \
+    CARGO_HOME=/usr/local/cargo
```
rockbmb: Is it ok to have this both here and below?

geigerzaehler: Yes. Is is supposed to take this value for `rustup-init`. But for the build we need to point it to the cache

PR #22, opened 2019-10-25
Add registry spec to repository

rockbmb:
Still in the process of updating this with the actual spec.

### DIFF
--- a/registry-spec/src/error.rs
+++ b/registry-spec/src/error.rs
@@ -4,9 +4,8 @@ pub enum TransferError {
     /// unit of currency.
     InvalidTransferAmount,
 
-    /// This type of error is only here tentatively since the validation of a
-    /// transfer's data may not necessarily occur in the Registry layer, meaning
-    /// it may not have to deal with this.
+    /// The transaction origin account's balance minus any transaction fee
+    /// was not greater than or equal to the amount being sent.
     InsufficientBalance,
 
     /// As mentioned in the whitepaper, the contracts associated with the
@@ -15,7 +14,7 @@ pub enum TransferError {
     ContractDenied,
 }
 
-pub enum ProjectValidationError {
+pub enum ValidationOfProjectRegistrationError {
     /// The origin of the `accept/reject_project` transaction is not
     /// in the set of root accounts.
     OriginNotRoot,
@@ -26,7 +25,7 @@ pub enum ProjectValidationError {
 
     /// The project in question has already been validated i.e. it has already
     /// been previously accepted/rejected.
-    ProjectAlreadyValidated,
+    ProjectVoteClosed,
 }
 
 /// Description of errors that may occur when registering a project in the
@@ -36,23 +35,14 @@ pub enum RegisterProjectError {
     /// The name with which the project was to be registered is invalid e.g.
     /// it is not valid UTF-8, it is longer than the protocol allows, or
     /// it has already been used.
-    InvalidName,
+    NameAlreadyExists,
 
-    /// The project's supplied domain was invalid e.g non-existant or just
+    /// The project's supplied domain was invalid e.g nonexistent or just
     /// invalid UTF-8.
     InvalidDomain,
 
     /// The project's supplied checkpoint did not exist.
     InvalidCheckpointId,
-}
-
-/// Representation of errors that may occur in `addkey` or `removekey`
-/// transactions.
-pub enum KeysetError {
-    /// Version 1.0 of the whitepaper does not mention what happens when
-    /// `addkey`/`removekey` are called with projects that have not yet been
-    /// added to the registry, so here that is tentatively treated as an error.
-    AccountNotInUse,
 }
 
 /// Errors that may happen when unregistering a project.
@@ -69,10 +59,10 @@ pub enum CheckpointError {
     /// The supplied parent contribution hash was not valid
     /// e.g. it was not empty in case of a project's first contribution, or was
     /// empty when it was not the first contribution.
-    InvalidParentCheckpoint,
+    ParentCheckpointDoesNotExist,
 
     /// The project state hash supplied with the checkpoint was not valid
-    /// e.g. it does not have t
+    /// e.g. it is improperly formed.
     InvalidCheckpointHash,
 
     /// The new project version supplied with this checkpoint was not valid
--- a/registry-spec/src/error.rs
+++ b/registry-spec/src/error.rs
@@ -1,14 +1,32 @@
 /// Description of errors that a transfer of Oscoin may raise.
 pub enum TransferError {
+    /// Amount to be transferred is not greater than or equal to 1 (one)
+    /// unit of currency.
+    InvalidTransferAmountError,
+
     /// This type of error is only here tentatively since the validation of a
     /// transfer's data may not necessarily occur in the Registry layer, meaning
-    /// it may not have to deal with this
+    /// it may not have to deal with this.
     InsufficientBalanceError,
 
     /// As mentioned in the whitepaper, the contracts associated with the
     /// sending and receiving addresses must authorize the transfer for it
     /// to be valid, otherwise it will result in this error.
     ContractDeniedError,
+}
+
+pub enum ProjectValidationError {
+    /// The origin of the `accept/reject_project` transaction is not
+    /// in the set of root accounts.
+    OriginNotRootError,
+
+    /// The hash of the transaction is invalid e.g. it does not correspond to
+    /// a `register_project` transaction, or has improper structure.
+    InvalidTransactionHashError,
+
+    /// The project in question has already been validated i.e. it has already
+    /// been previously accepted/rejected.
+    ProjectAlreadyValidatedError,
 }
 
 /// Description of errors that may occur when registering a project in the
--- a/registry-spec/src/error.rs
+++ b/registry-spec/src/error.rs
@@ -1,0 +1,67 @@
+/// Description of errors that a transfer of Oscoin may raise.
+pub enum TransferError {
+    /// This type of error is only here tentatively since the validation of a
+    /// transfer's data may not necessarily occur in the Ledger layer, meaning
+    /// it may not have to deal with this
+    InsufficientBalanceError,
+
+    /// As mentioned in the whitepaper, the contracts associated with the
+    /// sending and receiving addresses must authorize the transfer for it
+    /// to be valid, otherwise it will result in this error.
+    ContractDeniedError,
+}
+
+/// Description of errors that may occur when registering a project in the
+/// Oscoin ledger (`register` transaction). Not exhaustive, but should cover
+/// most common cases.
+pub enum RegisterProjectError {
+    /// The canonical source URL used to register the project is invalid.
+    ///
+    /// The 1.0 version of the whitepaper establishes only one condition for
+    /// the validity of the URL - the source code retrieved from it must
+    /// always hash to the `hash` field of the `Project` structure -
+    /// but this definition of validity can be tentatively extended to include:
+    ///
+    /// 1. it is a proper URL as defined in [RFC 3986](https:///tools.ietf.org/html/rfc3986#section-1.1.3), and
+    /// 2. it hosts the repository's page in a distributed version control
+    ///    system's website e.g. GitLab, BitBucket, SourceForge, GitHub, and
+    /// 3. it can be accessed without restrictions (^)
+    ///
+    /// (^) This part can be harder to define - if the URL permanently returns
+    /// `404`s *after* it has been inducted into the ledger, but not before,
+    /// is it still valid?
+    InvalidURLError,
+}
+
+/// Representation of errors that may occur in `addkey` or `removekey`
+/// transactions.
+pub enum KeysetError {
+    /// Version 1.0 of the whitepaper does not mention what happens when
+    /// `addkey`/`removekey` are called with projects that have not yet been
+    /// added to the ledger, so here that is tentatively treated as an error.
+    AccountIfNotInUseError,
+}
+
+/// Errors that may happen when unregistering a project.
+///
+/// Empty for now.
+pub enum UnregisterProjectError {}
+
+/// Errors that may occur when checkpointing a project.
+///
+/// Question:
+/// * Does an invalid dependency update list in a checkpoint invalidate it
+/// entirely?
+pub enum CheckpointError {
+    /// A dependency update is invalid if it adds a dependency the project
+    /// already uses.
+    UsedDependencyAddedError,
+
+    /// A dependency update is invalid if it removes a dependency the project
+    /// does not use.
+    UnusedDependencyRemovedError,
+
+    /// As the whitepaper says, a checkpoint is invalid if the dependency
+    /// update list containts duplicate dependencies.
+    DuplicateDependenciesError,
+}
--- a/registry-spec/src/error.rs
+++ b/registry-spec/src/error.rs
@@ -1,7 +1,7 @@
 /// Description of errors that a transfer of Oscoin may raise.
 pub enum TransferError {
     /// This type of error is only here tentatively since the validation of a
-    /// transfer's data may not necessarily occur in the Ledger layer, meaning
+    /// transfer's data may not necessarily occur in the Registry layer, meaning
     /// it may not have to deal with this
     InsufficientBalanceError,
 
@@ -12,7 +12,7 @@ pub enum TransferError {
 }
 
 /// Description of errors that may occur when registering a project in the
-/// Oscoin ledger (`register` transaction). Not exhaustive, but should cover
+/// Oscoin registry (`register` transaction). Not exhaustive, but should cover
 /// most common cases.
 pub enum RegisterProjectError {
     /// The canonical source URL used to register the project is invalid.
@@ -28,7 +28,7 @@ pub enum RegisterProjectError {
     /// 3. it can be accessed without restrictions (^)
     ///
     /// (^) This part can be harder to define - if the URL permanently returns
-    /// `404`s *after* it has been inducted into the ledger, but not before,
+    /// `404`s *after* it has been inducted into the registry, but not before,
     /// is it still valid?
     InvalidURLError,
 }
@@ -38,7 +38,7 @@ pub enum RegisterProjectError {
 pub enum KeysetError {
     /// Version 1.0 of the whitepaper does not mention what happens when
     /// `addkey`/`removekey` are called with projects that have not yet been
-    /// added to the ledger, so here that is tentatively treated as an error.
+    /// added to the registry, so here that is tentatively treated as an error.
     AccountIfNotInUseError,
 }
 
--- a/registry-spec/src/error.rs
+++ b/registry-spec/src/error.rs
@@ -33,22 +33,17 @@ pub enum ProjectValidationError {
 /// Oscoin registry (`register` transaction). Not exhaustive, but should cover
 /// most common cases.
 pub enum RegisterProjectError {
-    /// The canonical source URL used to register the project is invalid.
-    ///
-    /// The 1.0 version of the whitepaper establishes only one condition for
-    /// the validity of the URL - the source code retrieved from it must
-    /// always hash to the `hash` field of the `Project` structure -
-    /// but this definition of validity can be tentatively extended to include:
-    ///
-    /// 1. it is a proper URL as defined in [RFC 3986](https:///tools.ietf.org/html/rfc3986#section-1.1.3), and
-    /// 2. it hosts the repository's page in a distributed version control
-    ///    system's website e.g. GitLab, BitBucket, SourceForge, GitHub, and
-    /// 3. it can be accessed without restrictions (^)
-    ///
-    /// (^) This part can be harder to define - if the URL permanently returns
-    /// `404`s *after* it has been inducted into the registry, but not before,
-    /// is it still valid?
-    InvalidURLError,
+    /// The name with which the project was to be registered is invalid e.g.
+    /// it is not valid UTF-8, it is longer than the protocol allows, or
+    /// it has already been used.
+    InvalidNameError,
+
+    /// The project's supplied domain was invalid e.g non-existant or just
+    /// invalid UTF-8.
+    InvalidDomainError,
+
+    /// The project's supplied checkpoint did not exist.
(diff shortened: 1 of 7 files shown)


rockbmb:
The first two commits are just moving the spec from the prototype to here.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// registry.
+    fn unregister_project(
+        id: types::ProjectIdentifier,
+    ) -> Result<(), error::UnregisterProjectError>;
+
+    /// Checkpointing a project in Oscoin's registry.
+    fn checkpoint(
+        // Account identifier of the project to which a new checkpoint will be
+        // added.
+        project_account: types::ProjectId,
```
cloudhead: In the new model, checkpoints aren't associated 1-1 with projects, so this field can be removed.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        // Account to which Oscoin will be sent.
+        to_acc: types::AccountId,
+        // Amount of Oscoin to be sent.
+        amount: types::Oscoin,
+    ) -> Result<(), error::TransferError>;
+
+    fn accept_project(
+        // Hash of the `register_project` transaction for the `Project` in
+        // question.
+        t_hash: types::Hash,
```
geigerzaehler: Probably better to be explicit here and not assume that we have a uniform hash: Replace `types::Hash` with `types::TransactionHash`

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn transfer_oscoin(
+        // Account from which to send Oscoin.
+        from_acc: types::AccountId,
+        // Account to which Oscoin will be sent.
+        to_acc: types::AccountId,
+        // Amount of Oscoin to be sent.
+        amount: types::Oscoin,
+    ) -> Result<(), error::TransferError>;
+
+    fn accept_project(
```
geigerzaehler: Would be great to have documentation for this method.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// it may not have to deal with this.
+    InsufficientBalance,
+
+    /// As mentioned in the whitepaper, the contracts associated with the
+    /// sending and receiving addresses must authorize the transfer for it
+    /// to be valid, otherwise it will result in this error.
+    ContractDenied,
+}
+
+pub enum ProjectValidationError {
```
geigerzaehler: This name is a bit confusing. Maybe `ProjectRegistrationVoteError`?

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// in the set of root accounts.
+    OriginNotRoot,
+
+    /// The hash of the transaction is invalid e.g. it does not correspond to
+    /// a `register_project` transaction, or has improper structure.
+    InvalidTransactionHash,
+
+    /// The project in question has already been validated i.e. it has already
+    /// been previously accepted/rejected.
+    ProjectAlreadyValidated,
```
geigerzaehler: This is also a bit confusing. Maybe `VoteClosed`?

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
+/// Description of errors that a transfer of Oscoin may raise.
+pub enum TransferError {
+    /// Amount to be transferred is not greater than or equal to 1 (one)
+    /// unit of currency.
+    InvalidTransferAmount,
```
geigerzaehler: I don’t think these types of errors make sense to include here. They are validations of the transaction and independent of the chain state. They will be checked by the client before submitting the transaction. To capture these validations I would add them as pre-conditions to the method documentation.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+    InvalidCheckpointId,
+}
+
+/// Representation of errors that may occur in `addkey` or `removekey`
+/// transactions.
+pub enum KeysetError {
+    /// Version 1.0 of the whitepaper does not mention what happens when
+    /// `addkey`/`removekey` are called with projects that have not yet been
+    /// added to the registry, so here that is tentatively treated as an error.
+    AccountNotInUse,
```
geigerzaehler: We can remove this. We treat all accounts as existing.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+/// * Does an invalid dependency update list in a checkpoint invalidate it
+/// entirely?
+pub enum CheckpointError {
+    /// The supplied parent contribution hash was not valid
+    /// e.g. it was not empty in case of a project's first contribution, or was
+    /// empty when it was not the first contribution.
+    InvalidParentCheckpoint,
+
+    /// The project state hash supplied with the checkpoint was not valid
+    /// e.g. it does not have t
```
geigerzaehler: This line seems to be cut off

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        id: types::ProjectIdentifier,
+    ) -> Result<(), error::UnregisterProjectError>;
+
+    /// Checkpointing a project in Oscoin's registry.
+    fn checkpoint(
+        // Account identifier of the project to which a new checkpoint will be
+        // added.
+        project_account: types::ProjectId,
+        // Hash of the previous `checkpoint` associated with this project.
+        parent: types::CheckpointID,
```
geigerzaehler: Should be `Option<types::CheckpointId>` to allow for initial checkpoint creation.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Errors that may occur when checkpointing a project.
+///
+/// Question:
+/// * Does an invalid dependency update list in a checkpoint invalidate it
+/// entirely?
+pub enum CheckpointError {
+    /// The supplied parent contribution hash was not valid
+    /// e.g. it was not empty in case of a project's first contribution, or was
+    /// empty when it was not the first contribution.
+    InvalidParentCheckpoint,
```
geigerzaehler: The documentation and name do not make it clear when this error occurs. How about this
~~~rust
/// The supplied parent checkpoint does not exist.
ParentCheckpointDoesNotExist
~~~

registry-spec/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// A project's "smart" contract.
+///
+/// The actual type might not be representable as a regular data structure, or
+/// if it is, it may not be representable as part of a project's data
+/// structure, but it's kept here for visibility.
+pub struct Contract;
+
+/// A project's version at a given point in time.
+pub type Version = Vec<u8>;
```
geigerzaehler: Why is this a `Vec<u8>`? I’d prefer for this to be a string and document that it must be between 1 and 32 characters.

registry-spec/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+pub struct Oscoin;
+
+/// Representation of a contribution's author.
+pub type Author = PublicKey;
+
+/// ID of a project's current checkpoint.
+pub type CheckpointID = Hash;
+
+/// The `ProjectIdentifier` tuple serves to uniquely identify a project.
+pub type ProjectIdentifier = (ProjectName, ProjectDomain);
```
geigerzaehler: This supersedes the old `ProjectId` so we should rename this to `ProjectId`.

registry-spec/src/types.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// The name a `Project` has been registered with.
```
geigerzaehler: Please document that this must be between 1 and 32 characters.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
+/// Description of errors that a transfer of Oscoin may raise.
+pub enum TransferError {
+    /// Amount to be transferred is not greater than or equal to 1 (one)
+    /// unit of currency.
+    InvalidTransferAmount,
+
+    /// This type of error is only here tentatively since the validation of a
+    /// transfer's data may not necessarily occur in the Registry layer, meaning
+    /// it may not have to deal with this.
```
geigerzaehler: This is not correct. We’ll definitely handle this and this is a valid error case.

registry-spec/src/error.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// Description of errors that may occur when registering a project in the
+/// Oscoin registry (`register` transaction). Not exhaustive, but should cover
+/// most common cases.
+pub enum RegisterProjectError {
+    /// The name with which the project was to be registered is invalid e.g.
+    /// it is not valid UTF-8, it is longer than the protocol allows, or
+    /// it has already been used.
+    InvalidName,
```
geigerzaehler: As mentioned above we should distinguish between validation errors that are encoded by the types and handled client side and the registry errors. Let’s rename this to `NameAlreadyExists`.

registry-spec/src/types.rs
rockbmb: I know the spec isn't set in stone, but (for now) it does say each version is between 1 and 32 **bytes** - Making it a `String` is not really necessary, I believe, we don't need UTF-8.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    ) -> Result<(), error::KeysetError>;
+
+    /// Given a certain project, `removekey` removes a key from its set of
+    /// keys (c.f. section 4.4.1 of the whitepaper).
+    fn removekey(
+        id: types::ProjectId,
+        // Account identifier of the maintainer to be removed from the
+        // project's key set.
+        maintainer_key: types::AccountId,
+    ) -> Result<(), error::KeysetError>;
```
cloudhead: Both `addkey` and `removekey` are no longer needed, as we will be controlling access via the control solely.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Checkpointing a project in Oscoin's registry.
+    fn checkpoint(
+        // Account identifier of the project to which a new checkpoint will be
+        // added.
+        project_account: types::ProjectId,
+        // Hash of the previous `checkpoint` associated with this project.
+        parent: types::CheckpointID,
+        // New project hash - if the checkpoint succeeds, this will become its
+        // current hash.
```
cloudhead: In the current spec, creating a checkpoint doesn't affect any project until the project uses `set-checkpoint`

registry-spec/src/types.rs
rockbmb: I see, I left those two to coexist to see which we would drop. Is the old one to be removed, then?

registry-spec/src/error.rs
rockbmb: I understand your point, and I agree - I did not consider that if it becomes the client's responsibility to handle this, then this kind of error won't happen at this stage.

registry-spec/src/error.rs
rockbmb: Oversight, thanks.

registry-spec/src/error.rs
rockbmb: 👍

registry-spec/src/lib.rs
rockbmb: ✔️

registry-spec/src/error.rs
rockbmb: Will fix, I agree the name now is a little confusing.

registry-spec/src/types.rs
geigerzaehler: Yes. With the changed spec we also see that we don’t return the `ProjectId` from register project anymore. For convenience we should return the transaction hash so we can pass it on to the `accept` and `reject` methods.

registry-spec/src/types.rs
geigerzaehler: The spec says “a *string* between 1 and 32 bytes”. To me it is unclear what the meaning of the a simple byte string would be. @cloudhead?

registry-spec/src/types.rs
rockbmb: Thanks, oversight.

registry-spec/src/lib.rs
rockbmb: Alright - but still on this topic, how is this checkpoint associated with the project it's actually supposed to "checkpoint"? I presume it's the `parent` checkpoint id.

Also, unrelated: can anyone submit these transactions?

registry-spec/src/lib.rs
rockbmb: @cloudhead this is why I asked you on IRC about the hashes. I also think having a separate `TxHash` type would make sense.

Also, all of these registry transactions should now return `TxHash`, do you agree?

registry-spec/src/error.rs
cloudhead: Hmm but the transaction can fail if the preconditions are not met?

registry-spec/src/types.rs
cloudhead: :smile: - yeah I basically wanted to require ASCII here. We can use `String` anyway, since ASCII is valid UTF-8 - so when we parse the raw ascii input, it will be a valid `String`..

registry-spec/src/error.rs
cloudhead: Does `Vote` not imply a process with many voters? In this case it's more of an admin accepting/rejecting. However I can't think of a better alternative right now :thinking: - unless we split them, which could work: `RegistrationAcceptError` and `RegistrationRejectError`, since they are two transactions, it might be clearer.

registry-spec/src/error.rs
rockbmb: I do not quite understand, so let me ask a question: does this mean we want to disambiguate between `NameAlreadyExists` and `InvalidName` (invalid for structural reasons i.e. excessive length or not UTF-8), or remove the latter altogether because it will not exist at the registry level, but at the client level instead, therefore not needing to be included in this specification?

registry-spec/src/lib.rs
rockbmb: Oh, thanks for the notice - leftover from previous iteration of spec.

registry-spec/src/lib.rs
rockbmb: I haven't gotten to `set-checkpoint` yet, but you are nonetheless correct.

registry-spec/src/error.rs
geigerzaehler: Invalid transactions (transactions that don’t meet the preconditions) won’t be admitted to the ledger so they can never fail.

registry-spec/src/error.rs
geigerzaehler: Yes, these terms make more sense.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+/// the hash of the applied transaction if they succeed.
+pub trait RegistryTransactions {
+    /// The set of projects that are pending acceptance into the registry,
+    /// having been submitted with the `register_project` transaction.
+    type PendingProjectRegistrations = std::collections::HashSet<types::ProjectId>;
+
+    /// The set of root accounts, specified at genesis.
+    /// Per the spec, these accounts are the only ones authorized to perform
+    /// certain sets of actions e.g. accept or reject projects.
+    const ROOT_ACCOUNTS: std::collections::HashSet<types::AccountId>;
```
geigerzaehler: Not sure why this is on the `RegistryTransactions` trait. A `RegistyView` method would be more fitting.

registry-spec/src/error.rs
geigerzaehler: Again, the project name being a valid unicode string of limited size is a matter of validation and a precondition. As mentioned above we should focus on ledger errors that depend on the state. And there only `NameAlreadyExists` is such an error.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+/// A trait exposing the Oscoin registry transactions described in the
+/// whitepaper.
+///
+/// The methods here return `Result<types::TxHash, E>` for some error type `E` as they
+/// will be applying a modification on the Registry global state, and return
+/// the hash of the applied transaction if they succeed.
+pub trait RegistryTransactions {
+    /// The set of projects that are pending acceptance into the registry,
+    /// having been submitted with the `register_project` transaction.
+    type PendingProjectRegistrations = std::collections::HashSet<types::ProjectId>;
```
geigerzaehler: Not sure why this is on the `RegistryTransactions` trait and an associated type. A `RegistyView` method would be more fitting.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        // Account from which to send Oscoin.
+        from_acc: types::AccountId,
+        // Account to which Oscoin will be sent.
+        to_acc: types::AccountId,
+        // Amount of Oscoin to be sent.
+        amount: types::Oscoin,
+    ) -> Result<types::TxHash, error::TransferError>;
+
+    /// Accept a project that has been submitted for registration.
+    /// It is then removed from the pending registrations set.
```
geigerzaehler: Please also mention that it then can be retrieved with `get_project()`

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn reject_project(
+        // Hash of the `register_project` transaction for the `Project` in
+        // question.
+        t_hash: types::TxHash,
+    ) -> Result<types::TxHash, error::ValidationOfProjectRegistrationError>;
+
+    /// Registers a project on the Oscoin Registry and returns the new project’s ID.
+    ///
+    /// The transaction’s sender account becomes the initial maintainer of the project.
```
geigerzaehler: Only once the registration has been accepted.

registry-spec/src/lib.rs
rockbmb: That commit is still WIP. These concepts are quite useful in the spec so I moved them here as well. `RegistryView` sounds good.

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn get_registered_projects() -> std::collections::HashSet<types::ProjectId>;
+
+    /// The set of projects that are pending acceptance into the registry,
+    /// having been submitted with the `register_project` transaction.
+    fn get_pending_project_registrations() -> std::collections::HashSet<types::ProjectId>;
+
+    /// The set of root accounts, specified at genesis.
+    /// Per the spec, these accounts are the only ones authorized to perform
+    /// certain sets of actions e.g. accept or reject projects.
+    fn get_root_accounts() -> std::collections::HashSet<types::AccountId>;
```
geigerzaehler: I have a couple of meta-comments on the documentation style that would make the docs shorter, more concise, more precise, and leave less room for ambiguity.

* You don’t have have to refer to the spec. Keep the sentences short and drop the “Per the spec”.
* Try to avoid restating what we can glance from the types. “The set of root accounts” does not contain any information that is not present in the types.
* Try to be exhaustive. Instead of mentioning “certain sets of actions” and referring only to one example you should list all the actions. (Which in this case is only accepting and rejecting projects).
* Avoid adjectives like “certain” that don’t provide a clear definition or assume that the reader has some context beyond this spec.
* Present the most important information first. Here you mention that the accounts are specified at genesis before explaining what root accounts are. You can easily switch this around.

Taking all this into account I’d go with the following documentation.
~~~rust
/// Returns the set of accounts that are allowed to accept and reject project registration proposals.
///
/// The set is determined at genesis and cannot be changed.
~~~

registry-spec/src/lib.rs
```diff
@@ -0,0 +1,139 @@
+#![feature(associated_type_defaults)]
```
geigerzaehler: We can remove this now

registry-spec/src/types.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/// A project's "smart" contract.
+///
+/// The actual type might not be representable as a regular data structure, or
+/// if it is, it may not be representable as part of a project's data
+/// structure, but it's kept here for visibility.
+pub struct Contract {
+    pub ct: Box<dyn Fn(TxHash) -> ContractOutput>,
+}
```
geigerzaehler: Let’s leave this struct opaque: `pub struct Contract`. We still need to decide on a representation of contracts. Please also remove `ContractOutput`.

registry-spec/src/types.rs
```diff
@@ -0,0 +1,154 @@
...
+/// Numerical type representing Oscoin amounts. It is still to be decided, but
+/// it may be `u64`, `u128` or even a rational type so fractional amounts can
+/// be represented. Subject to discussion.
+pub struct Oscoin;
+
+/// Representation of a contribution's author.
+pub type Author = PublicKey;
+
+/// ID of a project's current checkpoint.
+pub type CheckpointId = TxHash;
```
geigerzaehler: This is not settled yet and we don’t want to rely on this. Let’s make `CheckpointId` and opaque type: `pub struct CheckpointId`

registry-spec/src/types.rs
```diff
@@ -0,0 +1,154 @@
...
+pub struct PublicKey;
+
+/// Type representing a GPG signature of data.
+pub struct Signature;
+
+/// Type for the hash digest used in the Oscoin Registry. Useful to represent
+/// commit hashes or public keys.
+pub struct Hash;
+
+/// Type representing a hashed transaction.
```
geigerzaehler:
```suggestion
/// The hash of a transaction. Uniquely identifies a transaction.
```

registry-spec/src/lib.rs
rockbmb: @geigerzaehler saying "certain actions" means the spec doesn't have to be rewritten everytime it's changed to include a new root account operation - right now there are two, in the future there may be e.g. 25 and I don't think we'll always remember to update every list that exhaustively mentions them.

I'll fix this now, though - I agree it does create a sense of ambiguity.

registry-spec/src/types.rs
rockbmb: The spec document is subject to change, yes, but reducing the above to `Fn(TxHash) -> ContractOutput>` is not far off from what a contract is described as.

rockbmb:
Merging this. Further updates to the spec in future PRs.

PR #24, opened 2019-10-28
Project registration

geigerzaehler:
Add project registration to runtime and client.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -1,0 +1,79 @@
+use alloc::prelude::v1::*;
+use alloc::vec;
+use codec::{Decode, Encode};
+use sr_primitives::weights::SimpleDispatchInfo;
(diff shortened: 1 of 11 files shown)


rockbmb:
Taking a look.

runtime/src/registry.rs
```diff
@@ -0,0 +1,79 @@
...
+        pub trait Store for Module<T: Trait> as Counter {
+            pub Projects: map H256 => Option<Project>;
+            pub ProjectIds: Vec<H256>;
+        }
+    }
+}
+
+pub use store::Store;
+
+decl_module! {
```
rockbmb: Are we going to have to do this for every single registry transaction?

geigerzaehler: No. We’ll have this for every module. In the beginning we’ll have only one module, I think.

PR #27, opened 2019-10-28
Remove CircleCI pipeline

rockbmb:
@geigerzaehler can you delete/remove the CircleCI service association on their end? I do not have permissions, it seems.


rockbmb:
I removed the webhook, which should be sufficient, but removing the organization should be the next step as well.

PR #28, opened 2019-10-30
Update substrate

geigerzaehler:
Update substrate

Also update the `jsonrpc-core-client` version used by `substrate-subxt`. This backports an upstream change that has already landed.

We also kill all tests from `substrate-subxt` because they do not compile with the substrate update. We can reinstantiate the tests once they are fixed upstream.

Also improved the update script to solve the following issues
* Error when a previous revision is not found
* Run `cargo update`


geigerzaehler:
Could you have a look at this @rockbmb?

rockbmb:
I started yesterday - a little big, I'll finish it ASAP.

PR #38, opened 2019-10-31
spec: Adjust type names and docs for more clarity

geigerzaehler:
* Rename `Oscoin` to `Balance`. The name of the currency is going to be `Rad` but we should focus on descriptive names and not own names.
* Rename `Proof` to `ProjectOwnershipProof` and make it opaque. It is not a general purpose proof and we don’t know yet what it’s structure is.
* Make `AccountId` opaque and update the documentation so that it becomes clear that it is a public key. Also remove `Address` which was only used to alias to `AccountId`.
* Prefix `PublicKey` and `Signature` with `Gpg`. This distinguishes GPG crypto used for contribution claiming from the rest of the registry crypto.
* Replace generic `HashLinkedList` with `ContributionList`. No need to abstract in the spec, especially if it is only used once. This also makes it easier to document.

### DIFF
--- a/registry-spec/src/lib.rs
+++ b/registry-spec/src/lib.rs
@@ -25,8 +25,7 @@ pub trait RegistryTransactions {
         from_acc: types::AccountId,
         // Account to which Oscoin will be sent.
         to_acc: types::AccountId,
-        // Amount of Oscoin to be sent.
-        amount: types::Oscoin,
+        amount: types::Balance,
     ) -> Result<types::TxHash, error::TransferError>;
 
     /// Registers a project on the Oscoin Registry and returns the new project’s ID.
@@ -43,15 +42,12 @@ pub trait RegistryTransactions {
     /// project is later accepted or rejected, the registration fee is deducted
     /// from the transaction sender's account - if no such available balance is
     /// found, it will result in an error.
-    ///
-    /// Preconditions:
-    /// * The ownership proof can only be up to 4096 bytes long.
     fn register_project(
         // Requested name of the project to be registered.
         project_id: types::ProjectId,
         project_checkpoint: types::CheckpointId,
         project_contract: types::Contract,
-        project_ownership_proof: types::Proof,
+        project_ownership_proof: types::ProjectOwnershipProof,
         project_meta: types::Meta,
     ) -> Result<types::TxHash, error::RegisterProjectError>;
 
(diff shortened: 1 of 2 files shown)


registry-spec/src/lib.rs
```diff
@@ -43,15 +42,12 @@ pub trait RegistryTransactions {
     /// project is later accepted or rejected, the registration fee is deducted
     /// from the transaction sender's account - if no such available balance is
     /// found, it will result in an error.
-    ///
-    /// Preconditions:
-    /// * The ownership proof can only be up to 4096 bytes long.
```
rockbmb: Why does this need to be removed?

I understand the spec is subject to change, but until the part that mentions this is removed, I believe this should be kept.

registry-spec/src/types.rs
```diff
@@ -86,34 +76,22 @@ pub struct Contract;
...
+///
+/// A contribution is valid if [sig] is a valid signature of [hash] for the public key [author].
 pub struct Contribution {
-    pub parent: Hash,
     pub hash: Hash,
-    // Note that `author` must be the public key that signed the contribution
-    // a data structure of this type must refer to.
-    pub author: Author,
-    pub sig: Signature,
+    /// Hash of the previous contribution’s [hash] or `None` if thi is the firt contribution.
```
rockbmb:
```suggestion
    /// Hash of the previous contribution’s [hash] or `None` if this is the first contribution.
```

registry-spec/src/lib.rs
geigerzaehler: Since the `ProjectOwnershipProof` is opaque it is unclear what the number of bytes refers to. Also this information should be encoded in the types so that we don’t have to mention it every time the type is used.

rockbmb: Not sure how to have vectors in Rust whose type encodes the "up to 4096 bytes" invariant.

geigerzaehler: > Not sure how to have vectors in Rust whose type encodes the "up to 4096 bytes" invariant.

We can build them ourselves with smart constructors.

rockbmb: Yes, `Vec::with_capacity`, however we won't be using constructors in the spec, so I understand why you removed that. Still, it'd be nice to have it documented somewhere - I don't think that part of the spec is changing anytime soon.

PR #39, opened 2019-10-31
Add clippy to Buildkite pipeline

rockbmb:


rockbmb:
@geigerzaehler it seems using `--exclude substrate-subxt` allows `clippy` to ignore the client library, whereas locally the behavior I got was a little different.

Can you take a quick look at this?

geigerzaehler:
> @geigerzaehler it seems using `--exclude substrate-subxt` allows `clippy` to ignore the client library, whereas locally the behavior I got was a little different.

I think clippy passes because all subxt issues were fixed. As I said I’m a bit reluctant to merge this with all the changes to subxt because it will make it harder to backport upstream patches.

rockbmb:
@geigerzaehler oh wait, I must have been distracted - yes, that is indeed the case...
Let me roll back those changes.

rockbmb:
Stale :moldy-bread:

PR #40, opened 2019-11-01
Add in-memory client

geigerzaehler:
* Add `radicle_registry_client_interface` package containing `Client` trait that provides an abstract client interface that is then implemented by `radicle_registry_client`.
* Add `MemoryClient` in the `radicle_registry_memory_client` package that provides an implementation of the `Client` interface backed by an in-memory instantiation of the registry ledger runtime.
* Re-write the runtime tests using the `MemoryClient`.

Follow-up issues to create
* Make `SyncClient` generic over `Client` trait
* Make runtime tests generic over client trait so that we can run the same test against a full node and an in-memory node.
* Make `MemoryClient` implement `Sync` so that it can be used by MVP
* Update MVP to use `MemoryClient`


memory-client/src/lib.rs
```diff
@@ -0,0 +1,125 @@
...
+        let test_ext = &mut self.test_ext.borrow_mut();
+        let result = test_ext.execute_with(f);
+        Box::new(future::result(result))
+    }
+}
+
+impl Client for MemoryClient {
+    type Error = Error;
+
+    fn transfer(
```
rockbmb: So we have to replicate every `Client` transaction for `MemoryClient`.

You abstracted the details into a trait, but it'll take a bit of work to remember to add every new feature to this client as well.

geigerzaehler: > So we have to replicate every `Client` transaction for `MemoryClient`.
> 
> You abstracted the details into a trait, but it'll take a bit of work to remember to add every new feature to this client as well.

Not sure what you are getting at. Yes, we’ll need to implement all functions twice. (At least until we have come up with a better solution.) But that is the hole point of this PR, providing two clients that adhere to the same interface. Plus the compiler will assist us in not forgetting to add methods.

rockbmb: >Not sure what you are getting at.
You said it yourself:
>Yes, we’ll need to implement all functions twice.

That was what I was trying to say. Not a fatal problem, in any case.

That

PR #41, opened 2019-11-04
Add `set_checkpoint` transaction to spec

rockbmb:


registry-spec/src/lib.rs
```diff
@@ -105,6 +105,27 @@ pub trait RegistryTransactions {
...
+    /// checkpoint.
+    fn set_checkpoint(
+        // Id of the project to be updated.
+        project_id: types::ProjectId,
+        // Id of the checkpoint to be associated to the above project.
+        checkpoint_id: types::CheckpointId,
+    ) -> Result<types::CheckpointId, error::SetCheckpointError>;
+
+    ///
+    ///
```
geigerzaehler: Documentation is missing

registry-spec/src/lib.rs
```diff
@@ -105,6 +105,27 @@ pub trait RegistryTransactions {
         // It is to be treated as a list i.e. processed from left to right.
         dependency_updates: Vec<types::DependencyUpdate>,
     ) -> Result<types::TxHash, error::CheckpointError>;
+
+    /// Set a project's checkpoint in the Oscoin registry.
```
geigerzaehler:
```suggestion
    /// Set a project's checkpoint in the Radicle registry.
```

registry-spec/src/lib.rs
```diff
@@ -105,6 +105,27 @@ pub trait RegistryTransactions {
...
+    ///
+    /// If succesful, the transaction returns the id of the project's new
+    /// checkpoint.
```
geigerzaehler: Why do we return the checkpoint ID? This transaction does not create checkpoints since you pass the checkpoint ID as a parameter.

registry-spec/src/lib.rs
```diff
@@ -105,6 +105,27 @@ pub trait RegistryTransactions {
...
+    fn set_checkpoint(
```
geigerzaehler: Who is allowed to execute checkpoints? What changes when you set a checkpoint. Right now you can’t observe the change? Can we document validating checkpoint ancestry?

registry-spec/src/error.rs
```diff
@@ -131,3 +131,33 @@ pub enum DependencyListError {
     /// Note that it does not have to refer to a project's existing version.
     InvalidProjectVersion,
 }
+
+/// Errors that may occur when setting a project's checkpoint.
+pub enum SetCheckpointError {
+    /// The supplied project id is invalid e.g. it is not present in the
+    /// the Oscoin registry because it is pending acceptance, or has already
+    /// been rejected.
+    InvalidProjectId,
```
geigerzaehler: Again, let‘s try to be more precise and comprehensive. Let’s rename this to `ProjectDoesNotExist`. There are no other `InvalidProjectId` errors for now,

registry-spec/src/error.rs
```diff
@@ -131,3 +131,33 @@ pub enum DependencyListError {
...
+
+    /// The supplied checkpoint id points to a checkpoint whose ancestry does
+    /// not contain the project's original checkpoint e.g. the checkpoint
```
geigerzaehler:
```suggestion
    /// not contain the project's original checkpoint i.e. the checkpoint
```

registry-spec/src/lib.rs
rockbmb: Oops, thanks.

registry-spec/src/lib.rs
rockbmb: You're right - should the transaction succeed, then the supplied checkpoint will be the project's new checkpoint so there's no need to return it again.

registry-spec/src/lib.rs
rockbmb: I believe I asked a similar question of permissions (I think my question was about who could call `withdraw_project`), but I may have forgotten to raise it publicly. I don't think that has been decided yet.

Regarding the other points, I'll expand on them in the comment.

geigerzaehler:
Don’t forget to rebase and squash ;)

PR #42, opened 2019-11-04
Document account keys

geigerzaehler:
Document what we use for account keys and how to create them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,37 @@ You’ll need to build the client with Rust Nightly.
 To build and view the client documentation run `./scripts/build-client-docs
 --open`.
 
+
+Account Keys
+------------
+
+We use Ed25519 keys for accounts. Key creation and handling functionality is
+provided by the `radicle_registry_client::ed25519` module. To use this module
+you will likely need to import the `radicle_registry_client::CryptoPair` and
+`radicle_registry_client::CryptoPublic` traits.
+
+You can create key pairs using [`CryptoPair::generate()`][api-pair-generate]
+```rust
+use radicle_registry_client::{ed25519, CryptoPair};
+let (key, seed) = ed25519::Pair::generate();
+```
+
+To create keys from human readable strings, use [`CryptoPair::from_string`][api-pair-from-string].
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -16,6 +16,32 @@ You’ll need to build the client with Rust Nightly.
 To build and view the client documentation run `./scripts/build-client-docs
 --open`.
 
+
+Account Keys
+------------
+
+We use Ed25519 keys for accounts. Key management functionality is provided by
```
xla: Does `management` imply anything beyond standardised creation of keys?

README.md
```diff
@@ -16,6 +16,32 @@ You’ll need to build the client with Rust Nightly.
...
+the `radicle_registry_client::ed25519` module. To use this module you will
+likely need to import the `radicle_registry_client::CryptoPair` and
```
xla: Is there a scenario where these traits don't need to be imported for the `ed25519` module to be usable?

README.md
```diff
@@ -16,6 +16,32 @@ You’ll need to build the client with Rust Nightly.
...
+`radicle_registry_client::CryptoPublic` traits.
+
+You can create key pairs using [`CryptoPair::generate()`][api-pair-generate]. To
+create keys from human readable strings, use
+```rust
+let alice = radicle_registry_client::ed25519::Pair::from_string("//Alice", None);
+```
+See [`CryptoPair::from_string`][api-pair-from-string].
+
+The `ed25519` module and the crypto traits are re-exports from
```
xla: Does this relate to the comment above?

README.md
geigerzaehler: You’ll also need to serialize keys probably. But “management” sounds to broad I changed it to “key creation and handling functionality”.

README.md
geigerzaehler: If you only use the types, say `ed25519::Public` then you don’t need the traits.

README.md
geigerzaehler: Not sure what you mean

PR #43, opened 2019-11-06
Add checkpoints

rockbmb:


client-interface/src/lib.rs
```diff
@@ -52,4 +52,13 @@ pub trait Client {
     fn get_project(&self, id: ProjectId) -> Response<Option<Project>, Self::Error>;
 
     fn list_projects(&self) -> Response<Vec<ProjectId>, Self::Error>;
+
+    fn create_checkpoint(
+        &self,
+        author: &ed25519::Pair,
+        project_id: ProjectId,
```
geigerzaehler: This should be an opaque project state hash, i.e. just a 32 byte blob that has no intrinsic meaning for now.

runtime/src/registry.rs
```diff
@@ -29,6 +35,21 @@ pub struct RegisterProjectParams {
...
+    pub parent: H256,
+    pub hash: H256,
+}
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq, Default)]
+pub struct Checkpoint {
+    pub parent: Option<CheckpointId>,
+    pub hash: H256,
+    pub version: Version,
+    pub contribs: Vec<Contribution>,
```
geigerzaehler: Could we please eliminate all the `contribs` and `hash` fields that are unused at the moment?

runtime/src/registry.rs
```diff
@@ -29,6 +35,21 @@ pub struct RegisterProjectParams {
...
+    pub checkpoint: CheckpointId,
+}
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Contribution {
+    pub parent: H256,
+    pub hash: H256,
+}
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq, Default)]
```
geigerzaehler: Could you please remove the `Default` instance? It is ill defined what this means and we don’t need it

runtime/src/registry.rs
```diff
@@ -42,6 +63,8 @@ pub mod store {
         pub trait Store for Module<T: Trait> as Counter {
             pub Projects: map H256 => Option<Project>;
             pub ProjectIds: Vec<H256>;
+            pub ProjectCheckpoints: map ProjectId => H256;
+            pub PreviousCheckpoints: map H256 => H256;
```
geigerzaehler: Why do we need this?

runtime/src/registry.rs
```diff
@@ -70,10 +94,27 @@ decl_module! {
...
+            Self::deposit_event(Event::CheckpointIndexed(checkpoint_id, prev_checkpoint_id));
+            Ok(())
+        }
     }
 }
 decl_event!(
     pub enum Event {
         ProjectRegistered(H256),
+        ProjectCheckpointAdded(H256, H256),
+        CheckpointIndexed(H256, H256),
```
geigerzaehler: What do we do with this event? Do we really need it?

runtime/src/registry.rs
```diff
@@ -70,10 +94,27 @@ decl_module! {
             Self::deposit_event(Event::ProjectRegistered(project_id));
             Ok(())
         }
+
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn create_checkpoint(
+            _origin,
+            project_id: ProjectId,
+            checkpoint_id: CheckpointId,
+            prev_checkpoint_id: CheckpointId,
```
geigerzaehler: Why is the previous checkpoint not optional?

runtime/src/registry.rs
rockbmb: This and the above are just the result of my misunderstanding, I'll remove these.

runtime/src/registry.rs
rockbmb: Removing it results in an error when the macro is processed:

```
the trait `std::default::Default` is not implemented for `registry::Checkpoint`
   |       in this macro invocation
```

When storing checkpoints in a storage map as values, wrapping them in an `Option` removed this issue.

geigerzaehler: > When storing checkpoints in a storage map as values, wrapping them in an `Option` removed this issue.

It is important to expose the value as `Option` so that we can easily check for existence.

geigerzaehler:
Could you also add some tests in `runtime/tests/main.rs`?

client-interface/src/lib.rs
```diff
@@ -52,4 +52,13 @@ pub trait Client {
     fn get_project(&self, id: ProjectId) -> Response<Option<Project>, Self::Error>;
 
     fn list_projects(&self) -> Response<Vec<ProjectId>, Self::Error>;
+
+    fn create_checkpoint(
+        &self,
+        author: &ed25519::Pair,
+        project_id: H256,
```
geigerzaehler: Please rename this parameter to `project_hash`

runtime/src/registry.rs
```diff
@@ -70,10 +85,28 @@ decl_module! {
...
         }
+
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn create_checkpoint(
+            _origin,
+            project_hash: H256,
+            checkpoint_id: CheckpointId,
+            prev_checkpoint_id: Option<CheckpointId>,
+        ) -> Result {
+            let checkpoint = Checkpoint {
```
geigerzaehler: Don’t forget `ensure_signed(origin)?`. We only want to accept properly signed transactions

runtime/src/registry.rs
```diff
@@ -70,10 +85,28 @@ decl_module! {
...
+
+            Self::deposit_event(Event::CheckpointIndexed(checkpoint_id));
+            Ok(())
+        }
     }
 }
 decl_event!(
     pub enum Event {
         ProjectRegistered(H256),
+        CheckpointIndexed(H256),
```
geigerzaehler: What do you think about renaming this to `CheckpointCreated`? It better relates to the transaction name and does not introduce another term (“indexed”).

runtime/src/registry.rs
```diff
@@ -42,6 +55,7 @@ pub mod store {
         pub trait Store for Module<T: Trait> as Counter {
             pub Projects: map H256 => Option<Project>;
             pub ProjectIds: Vec<H256>;
+            pub Checkpoints: map H256 => Option<Checkpoint>;
```
geigerzaehler: Please use `CheckpointId` as the key.

client/tests/end_to_end.rs
```diff
@@ -2,14 +2,18 @@
...
 #[test]
 fn register_project() {
     env_logger::init();
     let client = SyncClient::create().unwrap();
     let alice = ed25519::Pair::from_string("//Alice", None).unwrap();
 
+    let project_hash = H256::random();
+    let checkpoint_id = client
+        .create_checkpoint(&alice, project_hash, None)
+        .unwrap();
```
geigerzaehler: Can you please add the following assertions:
* The returned project checkpoint ID matches the provided checkpoint ID
* `client.get_checkpoint()` returns the correct checkpoint struct

runtime/tests/main.rs
```diff
@@ -4,14 +4,19 @@
...
 #[test]
 fn register_project() {
     let client = MemoryClient::new();
     let alice = key_pair_from_string("Alice");
 
+    let project_hash = H256::random();
+    let checkpoint_id = client
+        .create_checkpoint(&alice, project_hash, None)
+        .wait()
+        .unwrap();
```
geigerzaehler: Please add the assertions I made above. Consider splitting this into a different test case. Please also test checkpoint creation with a parent.

rockbmb:
@geigerzaehler haven't forgotten tests, will add before merger.

PR #45, opened 2019-11-08
Refactor project ID type

rockbmb:


runtime/src/registry.rs
```diff
@@ -40,8 +46,8 @@ pub mod store {
 
     decl_storage! {
         pub trait Store for Module<T: Trait> as Counter {
-            pub Projects: map H256 => Option<Project>;
-            pub ProjectIds: Vec<H256>;
+            pub Projects: map (String, String) => Option<Project>;
+            pub ProjectIds: Vec<(String, String)>;
```
geigerzaehler: Can you use `ProjectID` here, too?

rockbmb:
@geigerzaehler absolutely.

runtime/src/registry.rs
rockbmb: Since I saw an `H256`  where there could be `ProjectId` before this change, I believed type synonyms were not allowed in this declaration (perhaps the macro would not process them), but I was wrong.

PR #47, opened 2019-11-12
Extract transaction signing

geigerzaehler:
As a first step towards better transaction utilities we extract the transaction creation and signing logic from `subxt` into the `radicle_registry_client_common` crate.

### DIFF
--- a/client-common/src/lib.rs
+++ b/client-common/src/lib.rs
@@ -1,0 +1,136 @@
+//! This create provides code that is common to all client implementations.
+use parity_scale_codec::Encode;
+use radicle_registry_runtime::UncheckedExtrinsic;
+use sr_primitives::generic::{Era, SignedPayload};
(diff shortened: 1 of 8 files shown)


client-common/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+/// validation checks. See the `Checkable` implementation of [UncheckedExtrinsic] for how
+/// validation is performed.
+///
+/// `genesis_hash` is the genesis hash of the block chain this intrinsic is valid for.
+pub fn signed_extrinsic(
+    signer: &ed25519::Pair,
+    call: Call,
+    nonce: Index,
+    genesis_hash: Hash,
+) -> UncheckedExtrinsic {
```
rockbmb: I see, a successfully checked and submitted `UncheckedExtrinsic` becomes an `ExtrinsicSuccess`.

PR #48, opened 2019-11-13
Bump cache version to speed up builds

geigerzaehler:
We bump the cache version so a new cache is created and can cache recent
updates.


geigerzaehler:
Was already bumped for substrate update :man_facepalming:

PR #50, opened 2019-11-13
Add 'submit' method to 'Client' trait

geigerzaehler:
We add a `submit` method to the `Client` trait that is responsible for submitting any transaction. The `Client` transaction methods (`transfer`, `register_project`) are now implemented using of `submit` on the `Client` trait itself.

For `submit` we also introduce the `Call` enum that represents ledger calls.

### DIFF
--- a/client-common/src/lib.rs
+++ b/client-common/src/lib.rs
@@ -1,6 +1,7 @@
 //! This create provides code that is common to all client implementations.
 use parity_scale_codec::Encode;
 use radicle_registry_runtime::UncheckedExtrinsic;
+use radicle_registry_runtime::{balances, registry};
 use sr_primitives::generic::{Era, SignedPayload};
 use sr_primitives::traits::SignedExtension;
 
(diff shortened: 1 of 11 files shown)


client-common/src/lib.rs
```diff
@@ -36,6 +38,17 @@ pub fn signed_extrinsic(
...
 }
 
+/// Transforms a public ledger call into an internal runtime call.
+pub fn into_runtime_call(call: Call) -> RuntimeCall {
+    match call {
+        Call::RegisterProject(params) => registry::Call::register_project(params).into(),
+        Call::Transfer { recipient, balance } => {
+            balances::Call::transfer(recipient, balance).into()
+        }
+        Call::CreateCheckpoint(params) => registry::Call::create_checkpoint(params).into(),
```
rockbmb: Hope the merge conflict wasn't too hard to solve!

client-interface/src/lib.rs
```diff
@@ -8,43 +8,76 @@
...
+pub type TxHash = Hash;
+
+#[derive(Clone, Debug)]
+/// Message calls that can be sent to the ledger in a transaction.
+pub enum Call {
+    /// Registers a project with the given parameters.
+    RegisterProject(RegisterProjectParams),
+    /// Transfer `balance` from the author account to the recipient account.
+    Transfer {
+        recipient: AccountId,
```
rockbmb: Why not have a `TransferParams` as well?

client-interface/src/lib.rs
```diff
@@ -53,9 +86,22 @@ pub trait Client {
     fn create_checkpoint(
         &self,
         author: &ed25519::Pair,
-        project_id: H256,
-        prev_checkpoint: Option<CheckpointId>,
-    ) -> Response<CheckpointId, Error>;
+        project_hash: H256,
+        previous_checkpoint: Option<CheckpointId>,
+    ) -> Response<CheckpointId, Error> {
+        let checkpoint_id = CheckpointId::random();
```
rockbmb: This is still required, I see. Only when the tx can be signed without being submitted can this be removed and the actual tx hash used.

client-interface/src/lib.rs
geigerzaehler: Good point. Will fix.

PR #54, opened 2019-11-14
Add 'get_transaction_extra' method to 'Client' trait

geigerzaehler:
We add the `get_transaction_extra` method to the `Client` trait. This method was internal to the clients before and is part of the signing interface for clients.

### DIFF
--- a/client-interface/src/lib.rs
+++ b/client-interface/src/lib.rs
@@ -14,7 +14,7 @@ pub use radicle_registry_runtime::{
     registry::{
         Checkpoint, CheckpointId, CreateCheckpointParams, Project, ProjectId, RegisterProjectParams,
     },
-    AccountId, Balance,
+    AccountId, Balance, Index,
 };
 pub use substrate_primitives::crypto::{Pair as CryptoPair, Public as CryptoPublic};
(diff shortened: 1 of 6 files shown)


client-interface/src/lib.rs
```diff
@@ -26,6 +26,13 @@ pub type Error = substrate_subxt::Error;
 /// The hash of a transaction. Uniquely identifies a transaction.
 pub type TxHash = Hash;
 
+#[derive(Clone)]
```
rockbmb: Can't this be `Copy` too?

client-interface/src/lib.rs
```diff
@@ -26,6 +26,13 @@ pub type Error = substrate_subxt::Error;
...
+/// All data that is necessary to build the [SignedPayload] for a extrinsic.
+pub struct TransactionExtra {
+    pub nonce: Index,
+    pub genesis_hash: Hash,
```
rockbmb: Could you explain why this is needed? I cannot understand.

geigerzaehler: Transactions are signed against the genesis hash of the chain they apply to to prevent replay attacks against other chains, e.g. a testnet/mainnet.

PR #60, opened 2019-11-15
Project id validation

### LINKED ISSUE #46: Validate user-provided project ID
After #45, a project's ID is now a name/domain tuple, where both are `String`s.

However, their length is not validated - it should not me more than 32 characters.

This should happen after the user provides them and before they are added to a block via the `register_project` transaction, in order to prevent problems with names that are too large.

rockbmb:
Closes #46 .

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -1,11 +1,14 @@
+use alloc::format;
 use alloc::prelude::v1::*;
 use alloc::vec;
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, Error as CodecError, Input};
 use sr_primitives::weights::SimpleDispatchInfo;
 use srml_support::{
-    decl_event, decl_module, decl_storage, dispatch::Result, storage::StorageMap as _,
-    storage::StorageValue as _,
+    decl_event, decl_module, decl_storage, dispatch::Result as DispatchResult,
+    storage::StorageMap as _, storage::StorageValue as _,
 };
+
+use sr_std::str::FromStr;
 
 use substrate_primitives::H256;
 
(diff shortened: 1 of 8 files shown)


runtime/src/registry.rs
```diff
@@ -1,27 +1,62 @@
...
 use srml_system as system;
 use srml_system::ensure_signed;
 
 use crate::AccountId;
 
+/// Type to represent project names and domains.
+///
+/// Since their lengths are limited to 32 characters, a smart constructor is
+/// provided to check validity.
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
```
geigerzaehler: We need a custom `Decode` implementation that ensures that the decoded string is only 32 characters long. Otherwise it is possible to inject invalid data through the APIs.

runtime/src/registry.rs
```diff
@@ -1,27 +1,62 @@
...
+pub struct String32(String);
+
+impl String32 {
+    pub fn new() -> Self {
+        String32(String::with_capacity(32))
+    }
```
geigerzaehler: We don’t need this method. Please remove it.

runtime/src/registry.rs
```diff
@@ -1,27 +1,62 @@
...
+
+    pub fn from_string(s: String) -> Result<Self, String> {
```
geigerzaehler: Please add documentation that the only possible error can occur when the string is longer than 32 bytes. Please also add tests.

rockbmb: Ah, I accidentally removed the tests that I added in the first commit. I'll put them back.

runtime/src/registry.rs
rockbmb: Good point.

runtime/src/registry.rs
rockbmb: You're right.

runtime/src/registry.rs
rockbmb: > 
> 
> Please add documentation that the only possible error can occur when the string is longer than 32 bytes. Please also add tests.
>32 bytes

I believe here you meant 32 characters, since it's UTF-8.

runtime/src/registry.rs
```diff
@@ -1,27 +1,71 @@
...
+
+    /// This function only raises an error if the `String` it is passed is
+    /// longer than 32 characters.
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        String32::from_string(s.to_string())
+    }
+}
+
+impl Decode for String32 {
+    fn decode<I: Input>(input: &mut I) -> Result<Self, CodecError> {
```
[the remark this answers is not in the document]
rockbmb: Of course.

runtime/src/registry.rs
```diff
@@ -1,27 +1,71 @@
...
+        let decoded: String = String::from_utf8_lossy(&Vec::decode(input)?).into();
```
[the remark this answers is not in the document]
rockbmb: I'll try this out 👍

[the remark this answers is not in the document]
rockbmb: Yes, this does indeed make more sense - to use `impl Decode for String` instead of replicating it here.

PR #61, opened 2019-11-15
Return events from 'Client::submit'

geigerzaehler:
We introduce the `TransactionApplied` type that is returned from `Client::submit`. This gives client users the change to inspect events emitted by the submitted transaction.

### DIFF
--- a/client-interface/src/lib.rs
+++ b/client-interface/src/lib.rs
@@ -12,9 +12,10 @@ use radicle_registry_runtime::Hash;
 
 pub use radicle_registry_runtime::{
     registry::{
-        Checkpoint, CheckpointId, CreateCheckpointParams, Project, ProjectId, RegisterProjectParams,
+        Checkpoint, CheckpointId, CreateCheckpointParams, Event as RegistryEvent, Project,
+        ProjectId, RegisterProjectParams,
     },
-    AccountId, Balance, Index,
+    AccountId, Balance, Event, Index,
 };
 pub use substrate_primitives::crypto::{Pair as CryptoPair, Public as CryptoPublic};
 pub use substrate_primitives::{ed25519, H256};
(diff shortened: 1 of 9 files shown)


client-interface/src/lib.rs
```diff
@@ -39,6 +40,18 @@ pub struct TransferParams {
     pub balance: Balance,
 }
 
+/// Result of a transaction being included in a block.
+///
+/// Returned after submitting an transaction to the blockchain.
+#[derive(Clone, Debug)]
+pub struct TransactionApplied {
+    pub tx_hash: TxHash,
```
rockbmb: Naming's a little confusing - is it the hash of the transaction, or of the block the transaction is in?

rockbmb: Nevermind - this comment is wrong, my bad

client/tests/end_to_end.rs
```diff
@@ -19,19 +20,24 @@ fn register_project() {
         .wait()
         .unwrap();
     let project_id = ("NAME".to_string(), "DOMAIN".to_string());
-    client
-        .register_project(
+    let tx_applied = client
```
rockbmb: I just noticed you only did this here, and not with every other transaction. Is it intentional and one-off e.g. you want to test the events, but only here (see the `assert_eq!` with `tx_applied.events[0]` below), or does every transaction have to be done this way now?

geigerzaehler: I added the tests to ensure that the mechanism works. Since the mechanism is the same for every transaction it suffices to test this only once. (It would probably be better to have a separate test case for this). There’s no need to add this everywhere now. We’ll soon have a better solution to extract events that we can then use to test the domain logic.

rockbmb: Ok, that answers my question, thanks!

PR #62, opened 2019-11-17
Add `set_checkpoint` transaction to client

rockbmb:


runtime/src/registry.rs
```diff
@@ -116,11 +124,52 @@ decl_module! {
...
+            let mut ancestor_id = params.new_checkpoint_id;
+            // At the end of this loop, the value of `checkpoint_id` will be
+            // the ID of the first ancestor of the checkpoint in
+            // `params: SetCheckpointParams`.
+            while let Some(cp) = store::Checkpoints::get(ancestor_id) {
+                match cp.parent {
+                    None => break,
+                    Some(cp_id) => ancestor_id = cp_id,
+                }
+            };
```
geigerzaehler: How about extracting this piece of code? It is self contained, yet not trivial. And we could document the runtime complexity.

runtime/src/registry.rs
```diff
@@ -116,11 +124,52 @@ decl_module! {
...
+            let new_project = match opt_project {
+                None => return Err("The provided project ID is not associated with any project."),
+                Some(prj) =>
+                    Project {
+                        id: prj.id,
+                        description: prj.description,
+                        img_url: prj.img_url,
+                        members: prj.members,
+                        current_cp: params.new_checkpoint_id,
+                    }
```
geigerzaehler:
```rust
Some(project) => { 
  project.current_cp = params.new_checkpoint_id;
  project
}
```
The doesn’t require us to change the code when other `Project` fields are changed.

runtime/src/registry.rs
```diff
@@ -116,11 +124,52 @@ decl_module! {
...
+            // the ID of the first ancestor of the checkpoint in
+            // `params: SetCheckpointParams`.
+            while let Some(cp) = store::Checkpoints::get(ancestor_id) {
+                match cp.parent {
+                    None => break,
+                    Some(cp_id) => ancestor_id = cp_id,
+                }
+            };
+            if Some(ancestor_id) != store::InitialCheckpoints::get(params.project_id.clone()) {
+                return Err("The provided checkpoint ID is not a descendant of the project's first checkpoint.")
```
geigerzaehler: I think we should first check that the initial checkpoint (i.e. the project) exists before we find the root ancestor.

runtime/src/registry.rs
```diff
@@ -116,11 +124,52 @@ decl_module! {
...
+            Self::deposit_event(Event::CheckpointChanged(new_project.id, params.new_checkpoint_id));
+            Ok(())
+        }
     }
 }
 decl_event!(
     pub enum Event {
         ProjectRegistered(ProjectId),
         CheckpointCreated(CheckpointId),
+        CheckpointChanged(ProjectId, CheckpointId),
```
geigerzaehler: Let’s keep the event names consistent
```suggestion
        CheckpointSet(ProjectId, CheckpointId),
```

runtime/tests/main.rs
```diff
@@ -101,6 +104,118 @@ fn create_checkpoint() {
...
+        .unwrap();
+
+    let project_hash2 = H256::random();
+    let garbage = CheckpointId::random();
+    let new_checkpoint_id = client
+        .create_checkpoint(&david, project_hash2, Some(garbage))
+        .wait()
+        .unwrap();
+
+    let res = panic::catch_unwind(|| {
```
geigerzaehler: Could you please add a note that we need to do this because we cannot extract errors from the transactions yet?

runtime/tests/main.rs
```diff
@@ -101,6 +104,118 @@ fn create_checkpoint() {
...
+            &charles,
+            RegisterProjectParams {
+                id: project_id.clone(),
+                description: "DESCRIPTION".to_string(),
+                img_url: "IMG_URL".to_string(),
+                checkpoint_id,
+            },
+        )
+        .wait()
+        .unwrap();
```
geigerzaehler: It’s probably a good idea to extract project registration into a helper function.
```rust
fn create_project_with_checkpoint(client: &MemoryClient, author: &ed25519::Pair) -> Project
```

runtime/tests/main.rs
```diff
@@ -101,6 +104,118 @@ fn create_checkpoint() {
...
+    let checkpoint_ = Checkpoint {
+        parent: None,
+        hash: project_hash1,
+    };
+    let checkpoint = client
+        .get_checkpoint(checkpoint_id)
+        .wait()
+        .unwrap()
+        .unwrap();
+    assert_eq!(checkpoint_, checkpoint);
```
geigerzaehler: No need to check `get_checkpoint` here again. This is already covered in `create_checkpoint` and bloats this test unnecessarily.

runtime/tests/main.rs
```diff
@@ -101,6 +104,118 @@ fn create_checkpoint() {
...
+        )
+        .wait()
+        .unwrap();
+
+    let new_checkpoint_ = Checkpoint {
+        parent: Some(checkpoint_id),
+        hash: project_hash2,
+    };
+    let new_checkpoint = client
+        .get_checkpoint(new_checkpoint_id)
```
geigerzaehler: Why are we checking `get_checkpoint` again? Shouldn’t we be checking that `get_project` includes the new checkpoint?

rockbmb: Both work - in the end what should be done is a test to see if the new checkpoint has been set. Whether that's done via `get_checkpoint` or via `get_project` does not matter, I believe.

runtime/tests/main.rs
rockbmb: Good point.

runtime/tests/main.rs
rockbmb: Good point.

runtime/src/registry.rs
rockbmb: Makes sense.

runtime/src/registry.rs
rockbmb: I don't think the `StorageMap` methods have such information documented, so placeholders will have to be used for e.g. the performance of `store::Checkpoints::get`.

runtime/tests/main.rs
geigerzaehler: > in the end what should be done is a test to see if the new checkpoint has been set. Whether that's done via `get_checkpoint` or via `get_project` does not matter, I believe.

You’re right, we want to check the effect of the `set_checkpoint` transaction. But this transaction does not have any effect on the result of `get_checkpoint` since the transaction does not modify the checkpoint store. Conversely, if we introduce an error into the `set_checkpoint` transaction, say we make it a noop, then this test will still pass because we don’t assert that the project has been changed.

runtime/src/registry.rs
geigerzaehler: That is ok. I think it is just important to explain that the number of storage requests is linear in the size of the checkpoint ancestry.

runtime/tests/main.rs
rockbmb: You are correct - this test would only check the behavior of `create_checkpoint`. I understand that now, thanks for raising this.

runtime/src/registry.rs
rockbmb: Oh, great - I thought I had been missing something. At the start of the season I spent some time looking at Substrate's storage utilities, and I was surprised that there is not (yet) any mention of performance.

runtime/src/registry.rs
rockbmb: I tried doing this before submitting the PR, but unless `project` is declared to be mutable it won't work.

I understand your point, though - similar to Haskell's record substitution syntax `Project {.., current_cp = params.new_checkpoint_id}`.

runtime/src/registry.rs
rockbmb: I attempted to make this its own function, but having it inside the `decl_event!` macro made it impossible to return a `CheckpointId`, and hard to modify a `&mut` created elsewhere since there are limitations to what functions there can do.

geigerzaehler: I don’t quite get what you where trying to do, putting it in the `decl_event!` macro. I was suggesting this.
```rust
fn get_root_checkpoint(checkpoint_id: CheckpointId) -> CheckpointId {
    let mut ancestor_id = checkpoint_id;
    while let Some(cp) = store::Checkpoints::get(ancestor_id) {
        match cp.parent {
            None => break,
            Some(cp_id) => ancestor_id = cp_id,
        }
    }

    ancestor_id
}
```

rockbmb: >I don’t quite get what you where trying to do, putting it in the decl_event!

I didn't get it either, until I realized I meant to write `decl_module!`.

runtime/src/registry.rs
```diff
@@ -116,11 +124,58 @@ decl_module! {
             Self::deposit_event(Event::CheckpointCreated(params.checkpoint_id));
             Ok(())
         }
+
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn set_checkpoint(
+            origin,
+            params: SetCheckpointParams,
+        ) -> Result {
+            ensure_signed(origin)?;
```
geigerzaehler: You can fix this in a separate PR, but we only want project members to be able to create checkpoints.

rockbmb: You mean create checkpoints, or set? Because this comment is in the latter transaction.

geigerzaehler: Sorry, I meant “set checkpoints”

rockbmb: Alright, thanks for clarifying! 👌

runtime/src/registry.rs
```diff
@@ -116,11 +146,41 @@ decl_module! {
...
+
+            let opt_project = store::Projects::get(params.project_id.clone());
+            let new_project = match opt_project {
+                None => return Err("The provided project ID is not associated with any project."),
+                Some(prj) => {
+                    Project {
+                        current_cp: params.new_checkpoint_id,
+                        ..prj
```
rockbmb: @geigerzaehler seems like this is the way to do it.

runtime/src/registry.rs
rockbmb: Will do this in a separate PR.

rockbmb:
Should be proper now 👍

PR #65, opened 2019-11-18
Implement project accounts

geigerzaehler:
Every project is now associated with an account. When a project is registered we create a random account ID. We also implement a function that allows the project owner to transfer money from the project account to another user.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -3,19 +3,22 @@ use alloc::prelude::v1::*;
 use alloc::vec;
 use codec::{Decode, Encode, Error as CodecError, Input};
 use paint_support::{
-    decl_event, decl_module, decl_storage, dispatch::Result as DispatchResult,
-    storage::StorageMap as _, storage::StorageValue as _,
+    decl_event, decl_module, decl_storage,
+    dispatch::Result as DispatchResult,
+    storage::StorageMap as _,
+    storage::StorageValue as _,
+    traits::{Currency, ExistenceRequirement, Randomness as _},
 };
 use sr_primitives::weights::SimpleDispatchInfo;
 
 use sr_std::str::FromStr;
 
-use substrate_primitives::H256;
+use substrate_primitives::{crypto::UncheckedFrom, H256};
 
 use paint_system as system;
 use paint_system::ensure_signed;
 
-use crate::AccountId;
+use crate::{AccountId, Balance, Hash};
 
 /// Type to represent project names and domains.
 ///
@@ -88,6 +91,7 @@ pub type Version = String;
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
 pub struct Project {
     pub id: ProjectId,
+    pub account_id: AccountId,
     pub description: String,
     pub img_url: String,
     pub members: Vec<AccountId>,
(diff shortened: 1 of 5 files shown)


runtime/src/registry.rs
```diff
@@ -97,10 +114,21 @@ decl_module! {
...
-            Self::deposit_event(Event::ProjectRegistered(project_id.clone()));
+            Self::deposit_event(Event::ProjectRegistered(project_id, account_id));
             Ok(())
         }
 
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn transfer_from_project(origin, params: TransferFromProjectParams) -> Result {
+            let sender = ensure_signed(origin)?;
+            let project = store::Projects::get(params.project).ok_or("Project does not exist")?;
+            let is_member = project.members.iter().any(|member| *member == sender);
```
rockbmb: Can't you just do `project.members.contains(&sender)` here?

runtime/tests/main.rs
```diff
@@ -107,6 +104,109 @@ fn create_checkpoint() {
...
+        .wait()
+        .unwrap();
+
+    assert_eq!(
+        client.free_balance(&project.account_id).wait().unwrap(),
+        2000
+    );
+}
+
+fn create_project_with_checkpoint(client: &MemoryClient, author: &ed25519::Pair) -> Project {
```
rockbmb: When you rebase on master I'm not sure which version it'll keep, but I introduced this in #62 as you requested.

runtime/src/registry.rs
```diff
@@ -86,8 +99,12 @@ decl_module! {
         pub fn register_project(origin, params: RegisterProjectParams) -> Result {
             let sender = ensure_signed(origin)?;
             let project_id = params.id.clone();
+            let account_id = AccountId::unchecked_from(
+                srml_randomness_collective_flip::Module::<T>::random(b"project-account-id")
```
rockbmb: Is it possible, or sensible, to use the actual project ID here?

runtime/tests/main.rs
geigerzaehler: Yes, I’ll use your version.

runtime/src/registry.rs
geigerzaehler: This does not make sense since the sender should not be able to influence the random number generation.

runtime/tests/main.rs
geigerzaehler: Turns out that I needed my version because I need the whole project struct.

geigerzaehler:
@rockbmb Could you have another look. There have been substantial changes due to the rebase.

runtime/src/registry.rs
```diff
@@ -186,10 +203,21 @@ decl_module! {
...
 
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn transfer_from_project(origin, params: TransferFromProjectParams) -> DispatchResult {
+            let sender = ensure_signed(origin)?;
+            let project = store::Projects::get(params.project).ok_or("Project does not exist")?;
+            let is_member = project.members.contains(&sender);
+            if !is_member {
+                Err("Sender is not a project member")?;
+            }
+            <crate::Balances as Currency<_>>::transfer(&project.account_id, &params.recipient, params.value)
```
rockbmb: What is the meaning of this: `crate::Balances as Currency<_>`, is it an explicit cast?

geigerzaehler: Not quite. The transfer method on Balances is overloaded, there are two definitions of `Balances::transfer`. To distinguish them the trait needs to be provided.

PR #67, opened 2019-11-18
Return call result from 'Client::submit'

geigerzaehler:
We change `TransactionApplied` returned by `Client::submit` so that it includes the result of the call submitted.

This is accomplished by changing the call argument to `submit` from an enum to a trait. This new `Call` trait is implemented for all runtime calls, i.e. the variants that made up the `Call` enum before. The `Call` trait has an associated `Result` type that specifies the result of a runtime call.

Currently the result type for all calls is `Result<(), &'static str>` which maps one to one to the `DispatchResult` of a method in the runtime definition. In the future we’ll provide proper return values.

### DIFF
--- a/client-interface/src/call.rs
+++ b/client-interface/src/call.rs
@@ -1,0 +1,130 @@
+//! Defines [Call] trait and implementations for all transaction parameters.
+
+use radicle_registry_runtime::{balances, registry, Call as RuntimeCall, Event};
+use sr_primitives::DispatchError;
+
+/// Indicates that parsing the events into the approriate call result failed.
+type EventParseError = String;
+
+/// Trait implemented for every runtime call.
+///
+/// For every [RuntimeCall] that is exposed to the user we implement [Call] for the parameters
+/// struct of the runtime call.
+pub trait Call: Send + 'static {
+    /// Result of executing the call in the runtime that is presented to the client user.
+    type Result: Send + 'static;
+
+    /// Parse all runtime events emitted by the call and return the appropriate call result.
+    ///
+    /// Returns an error if the event list is not well formed. For example if an expected event is
+    /// missing.
+    fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError>;
+
+    fn into_runtime_call(self) -> RuntimeCall;
+}
+
+impl Call for registry::RegisterProjectParams {
+    type Result = Result<(), Option<&'static str>>;
+
+    fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError> {
+        let dispatch_result = get_dispatch_result(&events)?;
+        match dispatch_result {
+            Ok(()) => find_event(&events, "ProjectRegistered", |event| match event {
+                Event::registry(registry::Event::ProjectRegistered(_project_id)) => Some(Ok(())),
+                _ => None,
+            }),
+            Err(dispatch_error) => Ok(Err(dispatch_error.message)),
+        }
+    }
+
+    fn into_runtime_call(self) -> RuntimeCall {
+        registry::Call::register_project(self).into()
+    }
+}
+
+impl Call for registry::CreateCheckpointParams {
+    type Result = Result<(), Option<&'static str>>;
+
+    fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError> {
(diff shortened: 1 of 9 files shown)


client-interface/src/call.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    fn into_runtime_call(self) -> RuntimeCall {
+        registry::Call::register_project(self).into()
+    }
+}
+
+impl Call for registry::CreateCheckpointParams {
+    type Result = Result<(), Option<&'static str>>;
```
rockbmb: Is it possible to have `type Result = Result<CheckpointId, Option<&'static str>>;` so that we can finally return the created checkpoint's ID with the `create_checkpoint` tx?

memory-client/src/lib.rs
```diff
@@ -77,7 +77,15 @@ impl MemoryClient {
     pub fn new() -> Self {
         let genesis_config = GenesisConfig {
             srml_aura: None,
-            srml_balances: None,
+            srml_balances: Some(BalancesConfig {
+                balances: vec![(
+                    ed25519::Pair::from_string("//Alice", None)
```
rockbmb: What are the slashes in `"//Alice"` for?

runtime/tests/main.rs
```diff
@@ -107,6 +103,27 @@ fn create_checkpoint() {
...
+            &alice,
+            TransferParams {
+                recipient: bob,
+                balance: balance_alice + 1,
+            },
+        )
+        .wait()
+        .unwrap();
+
+    assert_eq!(tx_applied.result, Err(None))
```
rockbmb: Right, now that I see this, does it mean transaction errors are obtainable/processable, and that we can start using them to e.g. capture data validation failures?

client-interface/src/call.rs
geigerzaehler: Yes, this is now possible. We just need to generate the checkpoint ID in the runtime.

memory-client/src/lib.rs
geigerzaehler: You can learn more about it [here](https://crates.parity.io/substrate_primitives/crypto/trait.Pair.html#method.from_string_with_seed)

runtime/tests/main.rs
geigerzaehler: Almost. What does not yet work is getting the actual error message so we can distinguish errors

client-interface/src/call.rs
rockbmb: Hold on, I phrased my question poorly.

What I wanted to know is if we can finally return the *hash* of the created checkpoint's `create_checkpoint` transaction.

Also, what different will it make to generate the ID in the runtime instead of the client as it is currently done?

runtime/tests/main.rs
rockbmb: Will you do this in this PR, or is it beyond its scope?

client-interface/src/call.rs
geigerzaehler: This code does not allow us to use the transaction hash. We also don’t want the transaction author to determine the checkpoint ID—this would allow for collisions of IDs and be a security concern.

runtime/tests/main.rs
geigerzaehler: This is an issue with substrate so we have to postpone the work. I’ll create an issue for this.

geigerzaehler:
@rockbmb Could you have another look. There have been substantial changes due to the rebase.

rockbmb:
Alright, I'll do it!

runtime/tests/main.rs
```diff
@@ -202,23 +196,46 @@ fn fail_to_set_checkpoint() {
...
+            &david,
+            SetCheckpointParams {
+                project_id: project_id.clone(),
+                new_checkpoint_id,
+            },
+        )
+        .wait()
+        .unwrap();
+
+    assert_eq!(tx_applied.result, Err(None));
```
rockbmb: I see - in the future, when the error issue is sorted out upstream, we'll be able to directly discern the cause of the error in cases like these (and check that failure occured for the expected reasons).

Maybe even define our own `Error` enum and extend `substrate`'s with it.

geigerzaehler: Yes, indeed.

PR #72, opened 2019-11-19
Fix comments

rockbmb:


runtime/src/registry.rs
```diff
@@ -131,7 +131,7 @@ pub mod store {
     decl_storage! {
         pub trait Store for Module<T: Trait> as Counter {
             pub Projects: map ProjectId => Option<Project>;
-            pub InitialCheckpoints: map ProjectId => Option<CheckpointId>;
+            pub RootCheckpoints: map ProjectId => Option<CheckpointId>;
```
geigerzaehler: I would keep the name `InitialCheckpoints`. It is not necessary for the first project checkpoint to be a root checkpoint (i.e. not to have a parent).

rockbmb: Hm, is it not the other way around? E.g. `register_project` can only be called with a checkpoint that does not have a predecessor, which then becomes the project's first (or "root", in the spec) checkpoint.

rockbmb: This ties into #73, which is a result of not checking if the supplied checkpoint is root or not (my bad).

cloudhead: So let's take an example case: P1's current checkpoint K_100 has 99 ancestors, going back to a root checkpoint that has no parent. This root checkpoint happens to have been the project's initial checkpoint. Now P2 wants to fork from P1 at checkoint K_97 - this should be valid, because they share the history from K_0 to K_97. So P2 registers, using K_97 as the initial checkpoint. This means P2's initial checkpoint has a parent, and is not the root of the chain.

cloudhead: And hence, any future checkpoint from P2 needs to have K_97 in its ancestry.

rockbmb: I get it - from the perspective of P2, K_97 is a root checkpoint.

It doesn't matter what the storage map is called, but this does clear up the issue.

cloudhead: Yeah, dunno if it's confusing to call it "root" if it can have a parent - @geigerzaehler implied that "root" == "orphan", which makes sense. Maybe "initial" or "first" checkpoint are better names, and we keep "root" to mean the genesis/parentless checkpoint?

rockbmb: Ok, I can agree to that.

PR #75, opened 2019-11-21
Refactor checkpoint validation

rockbmb:
When creating a checkpoint, validate the supplied parent checkpoint.
A test for this has been added to `runtime/tests/main.rs`.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -116,7 +116,7 @@ pub struct Checkpoint {
 pub struct CreateCheckpointParams {
     pub checkpoint_id: CheckpointId,
     pub project_hash: H256,
-    pub previous_checkpoint: Option<CheckpointId>,
+    pub previous_checkpoint_id: Option<CheckpointId>,
 }
 
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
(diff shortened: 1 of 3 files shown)


runtime/tests/main.rs
```diff
@@ -67,6 +67,54 @@ fn register_project() {
...
+        )
+        .wait()
+        .unwrap();
+
+    assert_eq!(registration_2.result, Err(None));
+
+    let project = client.get_project(project_id).wait().unwrap().unwrap();
+
+    assert_eq!("DESCRIPTION", project.description);
+    assert_eq!("IMG_URL", project.img_url)
```
geigerzaehler: We don’t need to check that the original project was correctly registered. This is already tested by project registration and is not significant for this test.

runtime/src/registry.rs
```diff
@@ -233,8 +233,19 @@ decl_module! {
             params: CreateCheckpointParams,
         ) -> DispatchResult {
             ensure_signed(origin)?;
+
+            match params.previous_checkpoint_id {
+                None => {}
+                Some(cp_id) => {
+                    match store::Checkpoints::get(cp_id) {
+                        None => return Err("The previous checkpoint's ID that was supplied does not exist."),
```
geigerzaehler: Making it simpler.
```suggestion
                        None => return Err("Parent checkpoint does not exist"),
```

PR #76, opened 2019-11-22
Test duplicate project registrations

rockbmb:


runtime/tests/main.rs
```diff
@@ -67,6 +67,54 @@ fn register_project() {
...
+        )
+        .wait()
+        .unwrap();
+
+    assert_eq!(registration_2.result, Err(None));
+
+    let project = client.get_project(project_id).wait().unwrap().unwrap();
+
+    assert_eq!("DESCRIPTION", project.description);
+    assert_eq!("IMG_URL", project.img_url)
```
geigerzaehler: We don’t need to check that the original project was correctly registered. This is already tested by project registration and is not significant for this test.

geigerzaehler:
Please avoid unrelated updates to `Cargo.lock`

runtime/tests/main.rs
rockbmb: The idea here is to test that even after the invalid second registration, the data indexed by the project has not been modified - in case a duplicate registration gave an error, but still modified on-chain data.

geigerzaehler: Good point!

PR #77, opened 2019-11-23
Checkpoint ancestry

### LINKED ISSUE #73: Fix checkpoint update for projects with non-root initial checkpoint
Assume we register a project with a checkpoint that is not a root checkpoint, that is the checkpoint has a parent. A subsequent `set_checkpoint` transaction for the project with a descendant checkpoint of the initial checkpoint will fail.

rockbmb:
Closes #73.

Validation of checkpoint ancestry has been reworked, meaning that forks of projects are now able to create and set their own checkpoints without this bug causing that to fail.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -165,31 +165,36 @@ pub mod store {
 
 pub use store::Store;
 
-/// Given a checkpoint, return its oldest ancestor present in storage.
-///
-/// Note that this doesn't necessarily have to be the initial checkpoint of
-/// that checkpoint's project - even though this would mean the computed
-/// checkpoint, and every other checkpoint along the way, is invalid because
-/// their ancestry does not include the initial checkpoint.
-///
-/// Such an error can be handled by comparing the result of this function
-/// with the checkpoint used to register the project, available in
-/// `store::InitialCheckpoints`.
-fn get_initial_checkpoint(checkpoint_id: CheckpointId) -> CheckpointId {
+/// Returns true iff `checkpoint_id` descends from `initial_cp_id`.
+fn descends_from_initial_checkpoint(
+    checkpoint_id: CheckpointId,
+    initial_cp_id: CheckpointId,
+) -> bool {
+    if checkpoint_id == initial_cp_id {
+        return true;
+    };
+
+    let mut ancestor_id = checkpoint_id;
+
     // The number of storage requests made in this loop grows linearly
     // with the size of the checkpoint's ancestry.
     //
     // The loop's total runtime will also depend on the performance of
     // each `store::StorageMap::get` request.
-    let mut ancestor_id = checkpoint_id;
     while let Some(cp) = store::Checkpoints::get(ancestor_id) {
         match cp.parent {
-            None => break,
-            Some(cp_id) => ancestor_id = cp_id,
+            None => return false,
+            Some(cp_id) => {
+                if cp_id == initial_cp_id {
+                    return true;
+                } else {
+                    ancestor_id = cp_id;
+                }
+            }
         }
     }
 
-    ancestor_id
+    false
 }
 
 decl_module! {
@@ -296,8 +301,11 @@ decl_module! {
                 }
             };
 
-            let ancestor_id = get_initial_checkpoint(params.new_checkpoint_id);
-            if Some(ancestor_id) != store::InitialCheckpoints::get(params.project_id.clone()) {
+            let initial_cp = match store::InitialCheckpoints::get(params.project_id.clone()) {
+                None => return Err("A registered project must necessarily have an initial checkpoint."),
+                Some(cp) => cp,
+            };
+            if !descends_from_initial_checkpoint(params.new_checkpoint_id, initial_cp) {
                 return Err("The provided checkpoint ID is not a descendant of the project's initial checkpoint.")
             }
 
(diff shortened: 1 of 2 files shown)


rockbmb:
The removed `fail_to_set_checkpoint` test is added again in #75 since the cause for the false negative has been addressed there.

runtime/tests/main.rs
```diff
@@ -158,22 +158,22 @@ fn set_checkpoint() {
 }
 
 #[test]
-fn fail_to_set_checkpoint() {
```
geigerzaehler: Why are you deleting this test?

runtime/src/registry.rs
```diff
@@ -157,31 +157,40 @@ pub mod store {
...
-/// with the checkpoint used to register the project, available in
-/// `store::InitialCheckpoints`.
-fn get_initial_checkpoint(checkpoint_id: CheckpointId) -> CheckpointId {
+/// Given a checkpoint, check whether it descends from its project's initial
+/// checkpoint.
+fn descends_from_initial_checkpoint(
+    checkpoint_id: CheckpointId,
+    initial_cp_id: CheckpointId,
+) -> bool {
+    let mut res = true;
```
geigerzaehler: No need to have this mutable variable. Just return the literal result instead of using this or breaking.

runtime/src/registry.rs
```diff
@@ -157,31 +157,40 @@ pub mod store {
...
-/// Such an error can be handled by comparing the result of this function
-/// with the checkpoint used to register the project, available in
-/// `store::InitialCheckpoints`.
-fn get_initial_checkpoint(checkpoint_id: CheckpointId) -> CheckpointId {
+/// Given a checkpoint, check whether it descends from its project's initial
+/// checkpoint.
+fn descends_from_initial_checkpoint(
+    checkpoint_id: CheckpointId,
+    initial_cp_id: CheckpointId,
+) -> bool {
```
geigerzaehler: We could go for some clearer language that does not use the concepts “project” and “initial checkpoint”.
```rust
/// Returns `true` iff `checkpoint` is a descendant of `ancestor_candidate`.
fn checkpoint_is_descendant(checkpoint, ancestor_candiadate) -> bool
```

runtime/tests/main.rs
rockbmb: I was just about to add a comment to the PR: see #85.

This test fails because `set_checkpoint` does not validate the provided checkpoint - this is probably beyond the scope of the PR.

runtime/src/registry.rs
```diff
@@ -157,31 +157,35 @@ pub mod store {
...
+                    return true;
+                } else {
+                    ancestor_id = cp_id;
+                }
+            }
         }
     }
 
-    ancestor_id
+    true
```
geigerzaehler: This is possible attack vector. Since we don’t check that the checkpoint `checkpoint_id` exists it is possible to use a dummy checkpoint ID as `params.checkpoint_id`. Then the code will never enter the body of the `while` loop and we end up here and return `true`. This means we’ll accept the invalid checkpoint. Please fix this. Return `false` here and also check that the checkpoint exists. Maybe we could require a `checkpoint` instead of a `checkpoint_id` argument.

runtime/tests/main.rs
```diff
@@ -182,49 +182,66 @@ fn fail_to_set_checkpoint() {
...
+        id: fork_id.clone(),
+        description: project.description,
+        img_url: project.img_url,
+        checkpoint_id: checkpoints[2],
+    };
 
-    let project_hash2 = H256::random();
-    let new_checkpoint_id = client
-        .create_checkpoint(&eve, project_hash2, Some(project.current_cp))
+    client.register_project(&grace, params).wait().unwrap();
```
geigerzaehler: Is it really necessary to create another project in this test? What we want to test here is that we can set a new checkpoint that is an ancestor but not a direct child of the initial checkpoint.

runtime/src/registry.rs
rockbmb: Even though #85 would disallow phony `checkpoint_id`s to make it this far, you are correct - I will fix this.

runtime/tests/main.rs
rockbmb: >we can set a new checkpoint that is an ancestor but not a direct child

Perhaps you meant "we can set a new checkpoint that is a descendant but not a direct child"

runtime/src/registry.rs
```diff
@@ -286,8 +291,9 @@ decl_module! {
                 }
             };
 
-            let ancestor_id = get_initial_checkpoint(params.new_checkpoint_id);
-            if Some(ancestor_id) != store::InitialCheckpoints::get(params.project_id.clone()) {
+            let initial_cp =
+                store::InitialCheckpoints::get(params.project_id.clone()).expect("A registered project must necessarily have an initial checkpoint.");
```
geigerzaehler: I’m not sure it is a good idea to panic here. Although we only panic if the data is inconsistent this might be the case if we make an error in the implementation. I don’t now how substrate handles panics. It might prevent blocks from being evaluated and stall the chain.

You can leave it like it is for now but please open an issue to make sure we investigate.

rockbmb: I do recall seeing something, back when we started investigating Substrate, about **not** panicking in the runtime.

Honestly, I'll just fix it now, it's not a big issue.

PR #78, opened 2019-11-24
Checkpoint hash change

### LINKED ISSUE #53: Use Transaction Hash as checkpoint ID
At the moment the `create_checkpoint` method generates the checkpoint ID locally and submits it to the blockchain. However, the checkpoint ID should not be determined by the user. It should be deterministically generated and unique. There are two options to achieve this
* Use the transaction hash of the `create_checkpoint` transaction as the checkpoint ID. For this will need to make the transaction hash available to the runtime.
* Use substrates random number generator to generate the checkpoint ID and emit it as an event.

rockbmb:
Closes #53.

Checkpoint IDs are now a hash of the checkpoint's contents.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -19,7 +19,8 @@ use substrate_primitives::{crypto::UncheckedFrom, H256};
 use paint_system as system;
 use paint_system::ensure_signed;
 
-use crate::{AccountId, Balance, Hash};
+use crate::{AccountId, Balance, Hash, Hashing};
+use sr_primitives::traits::Hash as _;
 
 /// Type to represent project names and domains.
 ///
@@ -122,7 +123,6 @@ pub struct Checkpoint {
 
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
 pub struct CreateCheckpointParams {
-    pub checkpoint_id: CheckpointId,
     pub project_hash: H256,
     pub previous_checkpoint_id: Option<CheckpointId>,
 }
@@ -265,9 +265,10 @@ decl_module! {
                 parent: params.previous_checkpoint_id,
                 hash: params.project_hash,
             };
-            store::Checkpoints::insert(params.checkpoint_id, checkpoint);
+            let checkpoint_id = Hashing::hash_of(&checkpoint);
+            store::Checkpoints::insert(checkpoint_id, checkpoint);
 
-            Self::deposit_event(Event::CheckpointCreated(params.checkpoint_id));
+            Self::deposit_event(Event::CheckpointCreated(checkpoint_id));
             Ok(())
         }
 
--- a/runtime/tests/main.rs
+++ b/runtime/tests/main.rs
@@ -205,7 +205,6 @@ fn create_checkpoint_without_parent() {
     let client = MemoryClient::new();
     let bob = key_pair_from_string("Bob");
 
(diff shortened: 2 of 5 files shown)


runtime/src/registry.rs
```diff
@@ -224,6 +225,9 @@ decl_module! {
                 parent: params.previous_checkpoint,
                 hash: params.project_hash,
             };
+            if params.checkpoint_id != Hashing::hash_of(&checkpoint) {
+                return Err("Checkpoint Id computed by client is invalid.");
+            }
```
geigerzaehler: Instead of checking this here let’s just remove the `checkpoint_id` from the parameters and compute it locally.

runtime/tests/main.rs
```diff
@@ -125,6 +125,33 @@ fn create_checkpoint() {
...
+            CreateCheckpointParams {
+                checkpoint_id,
+                project_hash,
+                previous_checkpoint_id,
+            },
+        )
+        .wait()
+        .unwrap();
+
+    assert_eq!(tx_applied.result, Err(None))
```
geigerzaehler: This should not error

runtime/src/registry.rs
rockbmb: The problem then becomes, how do we return it from the registry runtime, so that it can be returned by the transaction? `DispatchResult = Result<(), &'static str>`.

Since we have to do it in the client because of the above, it should probably be checked.

geigerzaehler: We don’t need to return it from the transaction. We can compute it locally in the client. But we should also emit it as part of an event so that the client can pick it up later.

runtime/tests/main.rs
rockbmb: It should - `let previous_checkpoint_id = Some(CheckpointId::random());` means the provided parent checkpoint does not exist, which makes the `create_checkpoint` transaction invalid.

geigerzaehler: But this validation is not implemented yet, right? The test should be added to the PR that implements the change. Here we should test that the checkpoint hash is properly computed.

rockbmb: >The test should be added to the PR that implements the change.

Agreed, this was my oversight. Must have made a mistake while undoing the rebases.

runtime/src/registry.rs
rockbmb: Ok, we could compute it both in the client and in the node's runtime.

However, I'm still concerned about not checking client-produced data - do you think there's no reason to worry here?

geigerzaehler: > However, I'm still concerned about not checking client-produced data - do you think there's no reason to worry here?

I was suggesting removing the `checkopint_id` field from `CreateCheckpointParams`. So there is no need to check it because we don’t use the client produced data at all.

rockbmb: Then how will `Client::create:checkpoint` return a `CheckpointId` if

```
self.submit(
                author,
                CreateCheckpointParams {
...
```
cannot return anything?

geigerzaehler: The checkpoint ID is available as a constructor argument to `Event::CheckpointCreated`. We can use the `result_from_events` implementation of `CreateCheckpointParams` to extract that data. We just need to change the associated `Result` type.

rockbmb: Hm, I considered that, but wasn't sure it'd be viable.

It'll work for now.

PR #79, opened 2019-11-25
Implemen std::fmt::Display for String32

xla:
When handling the String32 type it's sometimes required to get the underlying String it wraps in order hand over to clients. As advises this is done by implementing the Display trait from std::fmt.

***
@geigerzaehler Working of your `consolidate-client-packages` branch and needed `to_string` on `String32`. Let me know if you want this change in a different form and would like to see tests for it.


rockbmb:
@xla you can merge this now, if you still need it.

runtime/src/registry.rs
```diff
@@ -50,6 +51,12 @@ impl FromStr for String32 {
     }
 }
 
+impl fmt::Display for String32 {
```
geigerzaehler: Could you add `#[cfg(feature = "std")]` so that we only compile this for the client?

xla: Is that true even when we use the `sr_std` crate?

rockbmb:
Did you mean to close this?

xla:
@rockbmb No, but the base branch vanished and I can't edit it....dumb UI. Gonna open a new one.

PR #80, opened 2019-11-25
Validate registration checkpoint

rockbmb:
When registering a project, the provided checkpoint was not validated. This is now done, and a test for this feature has been added to `runtime/tests/main.rs`.

### DIFF
--- a/runtime/tests/main.rs
+++ b/runtime/tests/main.rs
@@ -116,6 +116,33 @@ fn register_project_with_duplicate_id() {
 }
 
 #[test]
+fn register_project_with_bad_checkpoint() {
+    let client = MemoryClient::new();
+    let alice = key_pair_from_string("Alice");
+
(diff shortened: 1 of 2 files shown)


runtime/tests/main.rs
```diff
@@ -67,6 +67,77 @@ fn register_project() {
...
+    let params = RegisterProjectParams {
+        id: project_id,
+        description: "DESCRIPTION".to_string(),
+        img_url: "IMG_URL".to_string(),
+        checkpoint_id,
+    };
+
+    let tx_applied = client.submit(&alice, params).wait().unwrap();
+
+    assert_eq!(tx_applied.result, Err(None));
```
geigerzaehler: Please also assert that `client.get_project(project_id)` returns `None`

rockbmb: Good idea.

PR #83, opened 2019-11-26
Restructure client

### LINKED ISSUE #66: Reorganize client packages
Currently, we have two client (node and in-memory) and there are four packages that contain client logic: `client`, `memory-client`, `client-interface`, and `client-common`. Both clients share a substantial amount of code. This code is currently spread between `client-interface`, `client-common` or is simply duplicated in the client implementations. To address this issue I suggest the following.

We have one `Client` with multiple backends. The common client code would be part of the `Client` whereas the code that differs among the node and in-memory client would be part of a backend. The `Client` would be generic over a backend trait object. We would also collect all client related code in one package.

geigerzaehler:
This is a major restructuring of the client code to a new architecture.

Instead of having two clients that implement the `Client` trait we only have one `Client` with two backends—one for the remote node the other for the emulator (previously `MemoryClient`). Fixes #66

### Breaking changes
* `MemoryClient` was removed. Please use `Client::new_emulator()` now.

### Follow-up
* Turn `ClientWithExecutor` into a backend, if possible
* Eliminate `Client` trait and only provide `Client` structure to consumers.


geigerzaehler:
@xla: Just an FYI so you know what the external API changes are

geigerzaehler:
> Could you also update README.md and DEVELOPING.md ?

Done. Could you have another look @rockbmb?

PR #84, opened 2019-11-26
Implemen std::fmt::Display for String32 

xla:
When handling the String32 type it's sometimes required to get the underlying String it wraps in order hand over to clients. As advises this is done by implementing the Display trait from std::fmt.


runtime/src/registry.rs
```diff
@@ -50,6 +51,12 @@ impl FromStr for String32 {
     }
 }
 
+impl fmt::Display for String32 {
```
xla: @geigerzaehler You suggested putting this behind a config flag `#[cfg(feature = "std")]`, is that necessary even for `sr_std`?

geigerzaehler: No, it is not strictly necessary and it would also work without the flag. But we want to generate as little code as possible for the WASM target since the code lives on chain.

PR #90, opened 2019-11-29
Refactor test structure

### LINKED ISSUE #89: Split up tests
The [`runtime/tests/main.rs`](https://github.com/radicle-dev/radicle-registry/blob/master/runtime/tests/main.rs) file is getting too large, and it is becoming difficult to discern how many and what things are being tested there.

Ideally, tests concerning the same behavior (e.g. transaction or data type `impl`ementation) are moved to their own module for clarity.

rockbmb:
Closes #89.

Refactors tests so that each behavior being individually tested has its own module e.g. a module per registry transaction.


runtime/tests/string32.rs
```diff
@@ -0,0 +1,33 @@
...
+#[test]
+fn encode_then_decode() {
+    let string = String32::from_string(String::from("ôítÏйгますいщαφδвы")).unwrap();
+
+    let encoded = string.encode();
+
+    let decoded = <String32>::decode(&mut &encoded[..]).unwrap();
+
+    assert_eq!(string, decoded)
+}
```
geigerzaehler: Instead of creating a seperate file I’d actually prefer to move the code and the tests in to `runtime/src/string32`. If tests are closely related to one piece of code and not too long they should be kept together.

runtime/tests/common.rs
```diff
@@ -0,0 +1,38 @@
+/// Miscellaneous helpers used throughout other tests.
+use futures::prelude::*;
+
+use radicle_registry_client::*;
+
+#[allow(dead_code)]
```
geigerzaehler: At the moment this file is treated as a test file. If you move it to `runtime/tests/common/mod.rs` then it is treated like a module and you can remove the `allow` attribute (I hope). See [“Submodules in integration tests”](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests) from the Rust book for info.

runtime/tests/checkpoing_setting.rs
```diff
@@ -0,0 +1,103 @@
+/// Runtime tests implemented with [MemoryClient].
+///
+/// High-level runtime tests that only use [MemoryClient] and treat the runtime as a black box.
+///
+/// The tests in this module concern checkpoint creation.
```
geigerzaehler: ...and setting project checkpoints.

rockbmb: Thanks, Ctrl+C + Ctrl+V distraction.

runtime/tests/common.rs
rockbmb: I tried that module structure, but it wasn't working. Now it is, and I can't remember what I did wrong.

In any case, even with it, the `allow` is needed.

runtime/src/string32.rs
```diff
@@ -0,0 +1,84 @@
...
+        if decoded.len() > 32 {
+            Err(From::from("String32 length was more than 32 characters."))
+        } else {
+            Ok(String32(decoded))
+        }
+    }
+}
+
+#[test]
+fn long_string32() {
```
geigerzaehler: Please also add `#[cfg(test)]` so that we only compile this for the test target

runtime/tests/common.rs
geigerzaehler: Too bad :(

runtime/src/lib.rs
```diff
@@ -64,6 +64,9 @@ pub type EventRecord = paint_system::EventRecord<Event, Hash>;
 
 pub mod registry;
 
+pub mod string32;
```
geigerzaehler: No need to export the whole module
```suggestion
mod string32;
```

rockbmb: This also silently exports trait instances, then, right?

PR #96, opened 2019-12-09
Replace ClientWithExecutor with backend

geigerzaehler:
Follow-up of #83

We replace `ClientWithExecutor` with `Client::create_with_executor`.
This exposes less surface to users, makes the APIs more consistent and
requires less implementation overhead.

Breaking changes:
* Replace `ClientWithExecutor::create()` with `Client::create_with_executor()`


rockbmb:
I'm guessing next is the refactoring of the `Client` right?

PR #106, opened 2019-12-14
Update substrate node to use GRANDPA finality logic

rockbmb:
Previously, the `radicle-registry` node did not use the GRANDPA finality logic that is present in the [node template](https://github.com/paritytech/substrate/blob/40a16efefc070faf5a25442bc3ae1d0ea2478eee/bin/node-template/src/service.rs#L79) from Substrate it was adapted from.

This PR enables this functionality.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -78,3 +78,11 @@ rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 git = "https://github.com/paritytech/substrate"
 rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 package = "substrate-transaction-pool"
+
+[dependencies.substrate-finality-grandpa]
+git = "https://github.com/paritytech/substrate"
+rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
+
+[dependencies.substrate-finality-grandpa-primitives]
+git = "https://github.com/paritytech/substrate"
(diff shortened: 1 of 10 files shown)


node/Cargo.toml
```diff
@@ -78,3 +78,13 @@ rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 git = "https://github.com/paritytech/substrate"
 rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 package = "substrate-transaction-pool"
+
+[dependencies.grandpa]
```
geigerzaehler: I’d like to avoid package name aliases because they are very confusing. Same applies to the other packages.
```suggestion
[dependencies.substrate-finality-grandpa]
```

node/Cargo.toml
```diff
@@ -78,3 +78,13 @@ rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 git = "https://github.com/paritytech/substrate"
 rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
 package = "substrate-transaction-pool"
+
+[dependencies.substrate-finality-grandpa]
+git = "https://github.com/paritytech/substrate"
+rev = "dc4a8329b9c47790c25429a5df272b3648275c1b"
+package = "substrate-finality-grandpa"
```
geigerzaehler: You don’t need the `package` attribute if it is the same as the dependency name.

PR #108, opened 2019-12-16
Update substrate to 91ddc4f

rockbmb:
This PR is a Substrate revision update to commit `91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462`.


node/Cargo.toml
```diff
@@ -20,9 +20,9 @@ log = "0.4.8"
 parking_lot = "0.9.0"
 tokio = "0.1.22"
 trie-root = "0.15.2"
-# substrate-cli requires we fix this version
-structopt = "=0.3.4"
-structopt-derive = "=0.3.4"
+# sc-cli requires we fix these versions.
+structopt = "=0.3.7"
+structopt-derive = "=0.4.0"
```
geigerzaehler: Since you upgraded the versions the comment does not hold anymore. Please remove the comment and the explicit dependencies

client/src/call.rs
```diff
@@ -117,11 +115,11 @@ impl Call for crate::TransferParams {
 }
 
 impl Call for registry::TransferFromProjectParams {
-    type Result = Result<(), Option<&'static str>>;
+    type Result = Result<(), DispatchError>;
 
     fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError> {
         let dispatch_result = get_dispatch_result(&events)?;
-        Ok(dispatch_result.map_err(|dispatch_error| dispatch_error.message))
+        Ok(dispatch_result)
```
geigerzaehler: Can be simplified to just `get_dispatch_result(&events)`

node/src/service.rs
```diff
@@ -19,15 +19,14 @@ use aura_primitives::sr25519::AuthorityPair as AuraPair;
...
 use std::sync::Arc;
 use std::time::Duration;
-use substrate_client::LongestChain;
-use substrate_executor::native_executor_instance;
-pub use substrate_executor::NativeExecutor;
-use substrate_finality_grandpa::{self, FinalityProofProvider as GrandpaFinalityProofProvider};
-use substrate_service::{
-    error::Error as ServiceError, AbstractService, Configuration, ServiceBuilder,
-};
+use transaction_pool::{self};
```
geigerzaehler: Can be removed altogether. We only use this crate with explicit qualification.

node/Cargo.toml
```diff
@@ -34,72 +34,73 @@ version = "3.1.3"
...
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
 
-[dependencies.substrate-finality-grandpa-primitives]
+[dependencies.consensus-common]
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
+package = "sp-consensus"
```
geigerzaehler: In the new node template of this substrate version they stopped using package aliases. We don’t need to change the existing aliases but we should avoid adding new ones.

runtime/Cargo.toml
```diff
@@ -47,112 +48,114 @@ version = "1.0.101"
...
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
 default_features = false
 
 [dependencies.frame-executive]
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
+package = "frame-executive"
```
geigerzaehler: Not needed

runtime/Cargo.toml
```diff
@@ -47,112 +48,114 @@ version = "1.0.101"
...
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
 default-features = false
 
 [dependencies.pallet-sudo]
 git = "https://github.com/paritytech/substrate"
-rev = "1903f36f7bafea45eababfda9011a239b3aaab6c"
+rev = "91ddc4f1ffb3a961c0a39b28a6c6de1c4326d462"
+package = "pallet-sudo"
```
geigerzaehler: Not needed

runtime/Cargo.toml
```diff
@@ -12,27 +12,28 @@ repository = "https://github.com/radicle-dev/radicle-registry"
...
-    "substrate-offchain-primitives/std",
-    "substrate-primitives/std",
-    "substrate-session/std",
+    "sp-api/std",
+    "sp-block-builder/std",
+    "sp-consensus-aura/std",
+    "sp-offchain/std",
+    "sp-core/std",
+    "sp-session/std",
+    "sp-transaction-pool/std",
```
geigerzaehler: Personally, I like to sort this alphabetically so they’re easier to scan and it’s easier to find an item.

runtime/src/registry.rs
```diff
@@ -158,17 +159,17 @@ decl_module! {
             let sender = ensure_signed(origin)?;
 
             if store::Checkpoints::get(params.checkpoint_id).is_none() {
-                return Err("The checkpoint provided to register the project does not exist.")
+                return Err(DispatchError::Other("The checkpoint provided to register the project does not exist."))
```
geigerzaehler: I think it would actually make sense to use the `DispatchError::Module` variant here. This is more space efficient and allows us to give codes to errors. But we can also postpone this to the error revamp.

node/Cargo.toml
geigerzaehler: Please also add an issue to remove the package aliases.

runtime/Cargo.toml
geigerzaehler: Please also ensure that nothing is missing here by compiling the runtime standalone. E.g. `cargo check -p radicle-registry-runtime`.

geigerzaehler:
Please also add an entry to the changelog explaining the breaking change in the result types.

node/Cargo.toml
rockbmb: Good idea - I noticed that too. Package aliases are mostly gone now.

runtime/Cargo.toml
rockbmb: Is there any tool to format `.toml` files? I just threw this list into `ghci` and pretty-printed it, but it'd be nice if there was a better way to do it.

runtime/src/registry.rs
rockbmb: @geigerzaehler I should have documented this - yes, that was the idea, especially since we had spoken yesterday of error handling.

`DispatchError` is now an `enum`, so not all variants have a `message` field anymore. The idea was that we fix this as part of the larger error overhauling.

runtime/Cargo.toml
geigerzaehler: I’m not aware of any. I usually do sorting via my editor

PR #111, opened 2019-12-17
Remove underscores from package names

### LINKED ISSUE #102: Replace underscores with dashes in package names
It is more common in the Rust community to use dashes instead of underscores in package names. We should adhere to this.

rockbmb:
Closes #102.


geigerzaehler:
Don’t forget the change log either

rockbmb:
@geigerzaehler right, thanks for the reminder.

PR #113, opened 2019-12-19
Use async/.await for backend

geigerzaehler:
We use `async`/`.await` and `futures` v0.3 for all the backends.


scripts/run-dev-node
```diff
@@ -3,5 +3,5 @@
 set -euo pipefail
 
 export RUST_LOG="${RUST_LOG:-info},substrate_consensus_slots=error,substrate=error"
-radicle-registry-node=./target/release/radicle-registry-node
```
rockbmb: Did I miss this in #102 ?

geigerzaehler: Yes

PR #115, opened 2019-12-19
ci: Upload node binary to bintray on builds

geigerzaehler:
We upload the node binaries to bintray on every build. The download URL is documented in the README

### DIFF
--- a/ci/run
+++ b/ci/run
@@ -15,7 +15,7 @@ fi
 TIMEFORMAT='elapsed time: %R (user: %U, system: %S)'
 
 echo "--- Load cache"
-declare -i cache_version=13
+declare -i cache_version=14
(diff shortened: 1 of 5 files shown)


ci/run
```diff
@@ -87,3 +87,40 @@ else
   echo "$cache_key" > "$target_cache_key_file"
   echo "Size of $target_cache is $(du -sh "$target_cache" | cut -f 1)"
 fi
+
+# Upload the node binary to bintray using the commit hash as the version.
+function upload_to_bintray () {
+  declare -r api_url="https://bintray.com/api/v1"
```
kim: Bash zealot drive-by comment: `local` accepts the same flags as `declare`, but restricts the variable visibility to the function.

geigerzaehler: TIL

PR #116, opened 2019-12-19
Revamped CI caching

geigerzaehler:
We rewrote the caching logic for CI. The caching behavior is the same but now we don’t need to manually update the cache version and clean the cache.

### DIFF
--- a/ci/run
+++ b/ci/run
@@ -14,34 +14,16 @@ fi
 
 TIMEFORMAT='elapsed time: %R (user: %U, system: %S)'
 
-echo "--- Load cache"
-declare -i cache_version=14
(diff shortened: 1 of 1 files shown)


rockbmb:
Ah, right - Kim's point reminded me of something. I would also prefer to see `cargo clean` instead of manually removing cache directories.

kim:
See also oscoin/ci#29

geigerzaehler:
@kim: You’re totally right. This was way to complex and linking target works. (The only problem we have now is that the cache may grow indefinitely since we only add to it. But this should be fine). I implemented the necessary changes.

PR #119, opened 2019-12-20
Repeatable end-to-end tests

geigerzaehler:
Before, end-to-end test could only be run once since we were always using the same project ID. Now we generate a random project ID.

This change also includes a renaming of the `futures` crate to `futures01` for the runtime tests.


client/tests/common/mod.rs
```diff
@@ -0,0 +1,2 @@
+// TODO Very poor man’s code sharing. This should be moved into a shared package
```
rockbmb: How so? I think it's the other way around, a separate package for this is overkill.

Unless it's because we cannot use this in the client, in which case I would agree.

PR #120, opened 2019-12-20
Move subxt code into RemoteBackend

geigerzaehler:
We implement `RemoteNode::submit` with our own code instead of relying so much on `subxt` code.

This will enable us to ditch `substrate-subxt` and provides the necessary code to implement the transaction signing utilities.

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -38,8 +38,12 @@ pub use crate::transaction::{Transaction, TransactionExtra};
 #[doc(inline)]
 pub type Error = substrate_subxt::Error;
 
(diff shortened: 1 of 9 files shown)


client/src/interface.rs
```diff
@@ -38,8 +38,12 @@ pub use crate::transaction::{Transaction, TransactionExtra};
 #[doc(inline)]
 pub type Error = substrate_subxt::Error;
 
+/// The hash of a block. Uniquely identifies a transaction.
```
rockbmb: A transaction, or a block? Should it not be the latter?

PR #123, opened 2020-01-06
Eliminate subxt

### LINKED ISSUE #109: Inline errors intto `radicle_registry_client`
Currently the `Error` type exported by `radicle_registry_client` is a type alias for `substrate_subxt::Error`. To reduce the dependency on the `subxt` library and to be able to properly document and extend the errors we should introduce our own error enum. This enum should implement `From<substrate_subxt::Error>`.

geigerzaehler:
After #120 it is now possible to remove subxt.

* We create `client/src/error.rs` that defines the error enum that
  replaces `substrate_subxt::Error`. Note that we don’t need all the
  error cases.
* In `RemoteNode` we replace `subxt_client` with the underyling RPC
  clients.

Closes #109


ci/run
```diff
@@ -34,6 +34,7 @@ time cargo fmt --all -- --check
 echo "--- cargo build"
 export BUILD_DUMMY_WASM_BINARY=0
 export RUSTFLAGS="-D warnings"
+cargo clippy --workspace --all-targets --release -- -D clippy::all
```
rockbmb: The saga has finally concluded.

PR #124, opened 2020-01-06
Prepare package metadata for crate publish

geigerzaehler:
We update the `Cargo.toml` files so the packages are ready for publishing.

We also remove the `[package]` section from the workspace root. It is not needed.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -1,9 +1,13 @@
 [package]
-name = "radicle-registry-cli"
-version = "0.1.0"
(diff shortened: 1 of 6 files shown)


cli/Cargo.toml
```diff
@@ -1,9 +1,13 @@
 [package]
-name = "radicle-registry-cli"
-version = "0.1.0"
-authors = ["Thomas Scholtes <thomas@monadic.xyz>"]
```
rockbmb: Should we add ~the README.md~ the readme field?

geigerzaehler: Good point. #128

PR #126, opened 2020-01-06
Move test utils into own crate

### LINKED ISSUE #121: Move common testing tools into separate crate
PR #119 raised the fact that common testing utilities are not very well implemented because they are not usable, by default, outside of the `runtime` crate.

This issue is about moving them into a separate crate so that tests in e.g. `client` can use them without [`include!`](https://github.com/radicle-dev/radicle-registry/blob/master/client/tests/common/mod.rs) manipulation.

rockbmb:
Depends on #144, closes #121.

### DIFF
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -1,4 +1,20 @@
-/// Miscellaneous helpers used throughout other tests.
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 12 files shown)


test-utils/src/lib.rs
```diff
@@ -1,14 +1,30 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// Miscellaneous helpers used throughout Registry tests.
 use futures01::prelude::*;
 use rand::distributions::Alphanumeric;
 use rand::Rng;
 
+use sp_core::{ed25519, H256};
```
geigerzaehler: Why do we need this now?

test-utils/src/lib.rs
```diff
@@ -1,14 +1,30 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// Miscellaneous helpers used throughout Registry tests.
```
geigerzaehler: You probably meant this to be a module doc: `//!`.

PR #127, opened 2020-01-06
Update Cargo.lock

rockbmb:
It has been a couple of weeks since the last dependency update, so here it is.


geigerzaehler:
From the tests it looks like the update broke something.

rockbmb:
@geigerzaehler sadly it didn't get as far as to run tests - compile time failure in our substrate dependency checkout.

rockbmb:
Fixed by both #134 and #108.

PR #129, opened 2020-01-07
Separate transaction submission results

geigerzaehler:
`Client::submit` and  `Client::submit_transaction` now return a wrapped future. This allows consumers to distinguish the accepted and applied states of a transaction.

To implement this we change the backend interface and implementation accordingly.

We also add tests for invalid transactions.

### DIFF
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -14,8 +14,9 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 //! [backend::Backend] implementation for a remote full node
-use futures01::prelude::Stream as _;
+use futures01::stream::Stream as _;
 use futures03::compat::{Future01CompatExt as _, Stream01CompatExt as _};
+use futures03::future::BoxFuture;
 use futures03::prelude::*;
 use jsonrpc_core_client::RpcChannel;
 use parity_scale_codec::{Decode, Encode as _};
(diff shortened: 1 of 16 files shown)


client/src/backend/remote_node.rs
```diff
@@ -84,19 +85,35 @@ impl RemoteNode {
...
-                Some(tx_status) => match tx_status {
-                    TxStatus::Future | TxStatus::Ready | TxStatus::Broadcast(_) => continue,
-                    TxStatus::Finalized(block_hash) => return Ok(block_hash),
-                    TxStatus::Usurped(_) => return Err("Extrinsic Usurped".into()),
-                    TxStatus::Dropped => return Err("Extrinsic Dropped".into()),
-                    TxStatus::Invalid => return Err("Extrinsic Invalid".into()),
-                },
-            }
+        let opt_tx_status = tx_status_stream.try_next().await?;
+        match opt_tx_status {
```
rockbmb: Why does the first tx need to be handled outside of the `loop`?

geigerzaehler: We just want to listen to the first event which says whether the transaction was accepted or not.

geigerzaehler:
> when these functions return a result it means that the transaction has been accepted, when resolving the future inside it means the transaction has been applied?

Yes indeed. Do you have a suggestion to make the documentation clearer.

> the method chaining for this gets quite long, are there patterns/helpers to make the interaction nicer?

We have some helpers in the tests but otherwise no. Let’s see what the usage patterns in the proxy are and if it would make sense to provide some of them as convenience functions in the client.

PR #134, opened 2020-01-08
Update substrate to 1903f36f7

geigerzaehler:
The main changes here are renaming the `paint-` packages to `frame-`

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -14,11 +14,15 @@ radicle-registry-runtime = { version = "0.0.0", path = "../runtime" }
 
 derive_more = "0.15.0"
 exit-future = "0.1.4"
-futures = "0.1.29"
+futures = "0.3.1"
+futures01 = { package = "futures", version = "0.1" }
 log = "0.4.8"
 parking_lot = "0.9.0"
 tokio = "0.1.22"
 trie-root = "0.15.2"
+# substrate-cli requires we fix this version
+structopt = "=0.3.4"
+structopt-derive = "=0.3.4"
 
 [dependencies.codec]
 package = "parity-scale-codec"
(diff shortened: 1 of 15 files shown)


node/Cargo.toml
```diff
@@ -14,11 +14,15 @@ radicle-registry-runtime = { version = "0.0.0", path = "../runtime" }
...
 derive_more = "0.15.0"
 exit-future = "0.1.4"
-futures = "0.1.29"
+futures = "0.3.1"
+futures01 = { package = "futures", version = "0.1" }
 log = "0.4.8"
 parking_lot = "0.9.0"
 tokio = "0.1.22"
 trie-root = "0.15.2"
+# substrace-cli requires we fix this version
```
rockbmb:
```suggestion
# substrate-cli requires we fix this version
```

node/src/cli.rs
```diff
@@ -124,7 +138,7 @@ where
 // handles ctrl-c
 pub struct Exit;
 impl IntoExit for Exit {
-    type Exit = future::MapErr<oneshot::Receiver<()>, fn(oneshot::Canceled) -> ()>;
+    type Exit = Map<oneshot::Receiver<()>, fn(Result<(), oneshot::Canceled>) -> ()>;
```
rockbmb: Just a heads-up - `clippy::all` is now required by CI, and this typedef will raise a lint warning.
```suggestion
    type Fut = oneshot::Receiver<()>;
    type Func = fn(Result<(), oneshot::Canceled>) -> ();
    type Exit = Map<Fut, Func>;
```

node/src/service.rs
```diff
@@ -59,11 +58,14 @@ macro_rules! new_full_start {
         .with_select_chain(|_config, backend| {
             Ok(substrate_client::LongestChain::new(backend.clone()))
         })?
-        .with_transaction_pool(|config, client| {
-            Ok(transaction_pool::txpool::Pool::new(
-                config,
-                transaction_pool::FullChainApi::new(client),
-            ))
+        .with_transaction_pool(|config, client, _fetcher| {
```
rockbmb: It was a good idea to re-enable GRANDPA logic in this module in a previous PR, otherwise this file would have been annoying to rework.

node/src/cli.rs
geigerzaehler: I added an `#[allow(..)]`

PR #144, opened 2020-01-10
Move primitive types to own crate

### LINKED ISSUE #133: Move primitive datatypes into separate crate
In order to decouple the runtime and the client's mutual dependency, the primitive datatypes present in the runtime should be moved to a separate package.

Needed by #121, related to #95.

rockbmb:
Closes #133.

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -19,16 +19,11 @@
 //! methods to get the ledger state.
 use futures01::prelude::*;
 
+pub use radicle_registry_core::*;
+
 pub use radicle_registry_runtime::Hash;
 
-pub use radicle_registry_runtime::{
-    registry::{
-        Checkpoint, CheckpointId, CreateCheckpointParams, DispatchError, Event as RegistryEvent,
-        Project, ProjectDomain, ProjectId, ProjectName, RegisterProjectParams, SetCheckpointParams,
-        TransferFromProjectParams,
-    },
-    AccountId, Balance, Event, Hashing, Index, String32,
-};
+pub use radicle_registry_runtime::{registry::Event as RegistryEvent, Event};
 pub use sp_core::crypto::{Pair as CryptoPair, Public as CryptoPublic};
 pub use sp_core::{ed25519, H256};
 
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -1,0 +1,80 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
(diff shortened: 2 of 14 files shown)


client/src/interface.rs
```diff
@@ -19,15 +19,17 @@
 //! methods to get the ledger state.
 use futures01::prelude::*;
 
+pub use radicle_registry_core::{
+    AccountId, Balance, Checkpoint, CheckpointId, CreateCheckpointParams, Project, ProjectDomain,
+    ProjectId, ProjectName, RegisterProjectParams, SetCheckpointParams, String32,
+    TransferFromProjectParams,
+};
```
geigerzaehler: Would it make sense to re-export everything with `pub use radicle_registry_core::*`?

core/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub project_id: ProjectId,
+    pub new_checkpoint_id: CheckpointId,
+}
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct TransferFromProjectParams {
+    pub project: ProjectId,
+    pub recipient: AccountId,
+    pub value: Balance,
+}
```
geigerzaehler: How do you feel about grouping all the `XxxParams` objects in a `message` module?

client/src/interface.rs
```diff
@@ -19,15 +19,17 @@
...
 
 pub use radicle_registry_runtime::{
-    registry::{
-        Checkpoint, CheckpointId, CreateCheckpointParams, DispatchError, Event as RegistryEvent,
-        Project, ProjectDomain, ProjectId, ProjectName, RegisterProjectParams, SetCheckpointParams,
-        TransferFromProjectParams,
-    },
-    AccountId, Balance, Event, Hashing, Index, String32,
+    registry::{DispatchError, Event as RegistryEvent},
+    Event, Hashing, Index,
```
geigerzaehler: Can’t we move these into `core` too so that we don’t depend on runtime?

rockbmb: Sadly no, because this `Event` is defined in `runtime/lib.rs` and moving it means moving all of the `impl pallet_*Trait for Runtime {...` as well, at least from what I can tell - there's a lot of `type Event = Event;` in that module which is likely to be needed by a macro (`construct_runtime!`?)

core/src/lib.rs
rockbmb: Makes sense!

rockbmb:
> 
> 
> This makes the code much cleaner 🧹. Could you please add an item for the new package to DEVELOPING and quickly explain the heuristic for what code to put into the package?

> DEVELOPING
Did you perhaps mean CHANGELOG?

client/src/interface.rs
geigerzaehler: How about `DispatchError`, `Hashing`, and `Index`?

PR #158, opened 2020-01-14
Deploy new node versions to devnet

geigerzaehler:
Deploy new node versions to devnet


ci/deploy
```diff
@@ -0,0 +1,27 @@
...
+set -euo pipefail
+
+if [[ ${CI_VERBOSE:-} =~ ^1|true$ ]]; then
+  set -x
+fi
+
+image="gcr.io/opensourcecoin/radicle-registry/node:$BUILDKITE_COMMIT"
+
+function kubectl() {
+  ./kubectl \
```
rockbmb: Could you add a small section to `DEVELOPING` that explains the need for this command line tool in order to work with the devnet?

geigerzaehler: I added a section about continuous deployment to the developer manual but I’m omitting the details (e.g. `kubectl`): If you’re working with the devnet you’re working in the `radicle-registry-infrastructure` repo where this is documented. For this repo you don’t actually need `kubectl` because it is all managed by CI.

PR #160, opened 2020-01-15
Drop description and img_url fields

### LINKED ISSUE #137: Eliminate description and img_url fields from project registry
We are not using the `description` and `img_url` fields for a project. We should remove them everywhere.

/cc @xla

NunoAlexandre:
Closes #137 

These fields were initially included on a common-sense basis but they
turn out to neither be useful neither optimal to keep in a blockchain.
Hereby those two fields are removed from the Project model as well as
from the RegisterProjectParams message a node can receive.

### DIFF
--- a/runtime/tests/registration.rs
+++ b/runtime/tests/registration.rs
@@ -33,8 +33,6 @@ fn register_project() {
         .unwrap()
         .unwrap();
     assert_eq!(project.id, params.clone().id);
-    assert_eq!(project.description, params.description);
-    assert_eq!(project.img_url, params.img_url);
     assert_eq!(project.current_cp, checkpoint_id);
 
     assert_eq!(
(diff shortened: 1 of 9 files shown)


NunoAlexandre:
@geigerzaehler by dropping the mentioned fields everywhere in the project, the `register_project` test fails on getting a project using the client. I am investigating why that's happening.

runtime/tests/registration.rs
```diff
@@ -87,18 +85,16 @@ fn register_project_with_duplicate_id() {
...
     );
 
     assert_eq!(registration_2.result, Err(DispatchError::Other("")));
 
+    // TODO(nuno) delete line below if no further assert is required.
     let project = client.get_project(params.id).wait().unwrap().unwrap();
 
-    assert_eq!(params.description, project.description);
-    assert_eq!(params.img_url, project.img_url)
+    // TODO(nuno): Add assert_eq on id?
```
geigerzaehler: The idea for the assertions was to make sure that the project data was not overwritten. Since `description` and `img_url` are not available anymore we should use the checkpoint to assert this.

NunoAlexandre: Thanks, Thomas, it's now addressed.

NunoAlexandre:
@geigerzaehler could you have another look? The fields have been removed, the project builds, the tests pass (after recompiling the nodes), the code is correctly formatted, unused imports are removed, and the changelog is updated.

runtime/tests/registration.rs
```diff
@@ -83,22 +81,13 @@ fn register_project_with_duplicate_id() {
...
-    );
+    let registration_2 = submit_ok(&client, &alice, RegisterProjectParams { ..params.clone() });
 
     assert_eq!(registration_2.result, Err(DispatchError::Other("")));
 
     let project = client.get_project(params.id).wait().unwrap().unwrap();
 
-    assert_eq!(params.description, project.description);
-    assert_eq!(params.img_url, project.img_url)
+    assert_eq!(params.checkpoint_id, project.current_cp)
```
geigerzaehler: This assertion does guard against what we intended. Since the second project was created _with the same checkpoint_ we would expect the assertion to pass even if the project was overwritten.

NunoAlexandre:
@xla pinging you for this breaking change (described in the Changelog).

PR #162, opened 2020-01-15
cli: Set author key via env variable and document

geigerzaehler:
The author key for the cli can now be set through the `RAD_AUTHOR_KEY` env variable. We also improve the documentation of the option.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -20,12 +20,21 @@ use structopt::StructOpt;
 mod commands;
 use commands::*;
 
-#[derive(StructOpt, Debug, Clone)]
+#[derive(StructOpt, Clone)]
 #[structopt(max_term_width = 80)]
 struct Args {
-    /// The key pair that is used to sign transaction is generated from this seed.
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -42,17 +51,18 @@ struct Args {
...
             .unwrap();
         CommandContext {
-            author_key_pair,
+            author_key_pair: self.author_key.clone(),
             client,
         }
     }
+
+    fn parse_author_key(s: &str) -> Result<ed25519::Pair, String> {
+        ed25519::Pair::from_string(s, None).map_err(|err| format!("{:?}", err))
```
NunoAlexandre: why do we need the `map_err(|err| format!("{:?}", err))` block? Is it to convert Error to String?

geigerzaehler: Yes, we use the `std::fmt::Debug` implementation of the error to convert it into a string.

NunoAlexandre: clear, thanks.

PR #164, opened 2020-01-16
Use futures-0.3 in client code

### LINKED ISSUE #138: Use futures 0.3 for the client interface
The `Client` and `ClientT` methods should return v0.3 futures instead of v0.1 futures.

rockbmb:
Closes #138.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -12,7 +12,9 @@ repository = "https://github.com/radicle-dev/radicle-registry"
 [dependencies]
 radicle-registry-client = { version = "0.0.0", path = "../client" }
 
-futures01 = { package = "futures", version = "0.1" }
+async-std = { version = "1.4", features = ["attributes"] }
+async-trait = "0.1"
+futures = "0.3"
 hex = "0.4.0"
 lazy_static = "1.4"
 pretty_env_logger = "0.3.1"
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -13,7 +13,6 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use futures01::prelude::*;
 use radicle_registry_client::*;
 use structopt::StructOpt;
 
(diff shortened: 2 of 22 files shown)


cli/src/main.rs
```diff
@@ -66,13 +66,13 @@ enum Command {
 fn main() {
     pretty_env_logger::init();
     let args = Args::from_args();
-    let command_context = args.command_context();
+    let command_context = block_on(args.command_context());
```
geigerzaehler: Instead of using `block_on` multiple times in `main` you could use define `main` as `async` and use `#[async_std::main]`.

cli/Cargo.toml
```diff
@@ -12,7 +12,8 @@ repository = "https://github.com/radicle-dev/radicle-registry"
 [dependencies]
 radicle-registry-client = { path = "../client" }
 
-futures01 = { package = "futures", version = "0.1" }
+async-trait = "0.1"
+futures03 = { package = "futures", version = "0.3" }
```
geigerzaehler: If we’re using only version 0.3 we should just call this package `futures`.

client/examples/project_registration.rs
```diff
@@ -1,22 +1,22 @@
...
 use radicle_registry_client::*;
 
-fn main() {
+#[async_std::main]
+async fn main() {
     env_logger::init();
     let mut runtime = tokio::runtime::Runtime::new().unwrap();
     runtime.block_on(Compat::new(go().boxed())).unwrap();
-    runtime.shutdown_now().wait().unwrap();
+    runtime.shutdown_now().compat().await.unwrap();
```
geigerzaehler: If we use `async_std::main` we don’t need to spawn a tokio runtime. just call `go().await.unwrap()`

geigerzaehler: Argh... This might not actually work because the websocket library we use requires a tokio 0.1 runtime. You may want to skip this and tackle it later.

rockbmb: Yes - I ran into that issue myself.

This'll have to be a separate issue - move from `tokio-0.1` to `tokio-0.2` to finally remove all remnants of `futures-0.1` from the codebase. I'll create an issue.

PR #167, opened 2020-01-16
Add metadata field to Project

### LINKED ISSUE #156: Add metadata field to project
We add a `metadata` field of type `Vec<8>` to the project. The field is stored in the project state. We limit the size of the metadata to 128bytes.

NunoAlexandre:
Closes #156 

This PR adds a new `metadata` field to Project. This fields is a `Vec<u8>`, that can be anything serialized as an array of bytes. What is and how it's (de)serialized is out of the Registry's scope.

### Note

@geigerzaehler on #156 you mention this field should host at most 128 bytes. I'd thus make it `Vec<[u8; 128]>`. To avoid duplication, I'd prefer to have at least a type alias. Just want to double-check to know how you guys go about this. Also, how will the Registry act when given metadata with more than 128 bytes? Will it take the first 128 or refuse the request?

### TODO 

- [x] Add metadata field
- [x] Fix tests
- [x] Create `Bytes128` type
- [x] Replace `Vec<u8>` with `Bytes128`
- [x] Validate valid `metadata`

### DIFF
--- a/core/src/bytes128.rs
+++ b/core/src/bytes128.rs
@@ -16,22 +16,19 @@
 /// `Bytes128` type, and its validation tests.
 use alloc::format;
 use alloc::prelude::v1::*;
+use core::convert::TryFrom;
 use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
-use std::convert::TryFrom;
 
-/// This type is used to represent project metadata fields.
-///
-/// Radicle limits the size of the project metadata field to 128 bytes.
-/// To guarantee that at the type-level, a smart constructor is provided to check validity.
+/// Byte vector that is limited to 128 bytes.
 #[derive(Encode, Clone, Debug, Eq, PartialEq)]
 pub struct Bytes128(Vec<u8>);
-
-/// TODO review
-/// 1. From and To vec traits
 
 impl Bytes128 {
     const MAXIMUM_SUPPORTED_LENGTH: usize = 128;
 
+    /// Smart constructor that attempts to build a Bytes128
+    /// from a Vec<u8> with an arbitrary size. It fails if the
+    /// input vector is larger than Bytes128::MAXIMUM_SUPPORTED_LENGTH.
     pub fn from_vec(vector: Vec<u8>) -> Result<Self, String> {
         if vector.len() > Self::MAXIMUM_SUPPORTED_LENGTH {
             Err(format!(
@@ -46,27 +43,36 @@ impl Bytes128 {
 }
 
 impl TryFrom<Vec<u8>> for Bytes128 {
-    type Error = &'static str;
+    type Error = String;
 
     fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
         Bytes128::from_vec(value)
     }
 }
 
+impl From<Bytes128> for Vec<u8> {
+    fn from(value: Bytes128) -> Self {
+        value.0
+    }
+}
+
 /// Bytes128 random functions useful for unit testing.
-/// Note that since these fuctions make use of rand,
-/// we need to guard with the std feature to
-/// be able to compile it for wasm.
+///
+/// Note that since these fuctions make use of rand, we need to guard
+/// with the std feature to be able to compile it for wasm.
 #[cfg(feature = "std")]
 impl Bytes128 {
+    /// Generate a random Bytes128 vector with as many bytes as its limit.
     pub fn random() -> Self {
         Self::from_vec(Self::random_vector(Self::MAXIMUM_SUPPORTED_LENGTH)).unwrap()
     }
 
+    /// Generate a random Bytes128 vector with as many bytes as specified with 'size'.
     pub fn random_with_size(size: usize) -> Result<Self, String> {
         Bytes128::from_vec(Self::random_vector(size))
     }
 
+    /// Generate a random Vec<u8> with as many bytes as specified with 'size'.
     fn random_vector(size: usize) -> Vec<u8> {
         (0..size).map(|_| rand::random::<u8>()).collect()
     }
@@ -75,11 +81,8 @@ impl Bytes128 {
 impl Decode for Bytes128 {
     fn decode<I: Input>(input: &mut I) -> Result<Self, CodecError> {
         let decoded: Vec<u8> = Vec::decode(input)?;
-        Bytes128::from_vec(decoded).or_else(|_| {
-            Err(CodecError::from(
-                "Bytes128 length was more than 128 characters.",
-            ))
-        })
+        Bytes128::from_vec(decoded)
+            .or_else(|_| Err(CodecError::from("Failed to decode an inordinate Bytes128.")))
     }
 }
 
--- a/core/src/bytes128.rs
+++ b/core/src/bytes128.rs
@@ -1,0 +1,108 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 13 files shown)


runtime/tests/registration.rs
```diff
@@ -85,10 +86,9 @@ fn register_project_with_duplicate_id() {
 
     assert_eq!(registration_2.result, Err(DispatchError::Other("")));
 
-    let _project = client.get_project(params.id).wait().unwrap().unwrap();
+    let project = client.get_project(params.id).wait().unwrap().unwrap();
 
-    // TODO(nuno): assert that the project data is left untouched.
-    // This can be done again when the metadata field is added.
+    assert_eq!(params.metadata, project.metadata)
```
geigerzaehler: When registering the second project we’re reusing the parameters (including the metadata) of the first project. So even if the second registration overrides the first one we will not notice because the metadata is the same. To fix this `registration2` should use a different metadata value.

xla:
@NunoAlexandre From a consumers perspective it should throw an explicit error for there not to be subtle errors, i.e. a client puts in more than 128bytes expecting to read it out and then later on the deserialise breaks. This is esp. important given that this data is meant to be immutable.

NunoAlexandre:
@xla awesome, we are on the same page as for the approach then. Thanks!

geigerzaehler:
I though about the 128 byte limit a bit more and came to the conclusion that using the same approach as with [`String32`](https://github.com/radicle-dev/radicle-registry/blob/master/core/src/string32.rs#L29-L42) is probably best: Let’s implement a `Bytes128` with a smart constructor, `impl Bytes128 for Vec<u8>` and a custom decoder implementation

Advantages
* The client is forced to check that it constructs the right values early. It is impossible to construct an invalid transaction with Rust.
* Nodes will reject long values early (at the decoding step), this holds true for the transaction and block validation

Disadvantages
* Using a decoding error for something that sounds more like a validation might be confusing
* Approach does not scale indefinitely. This is a concern that we can handle later

runtime/tests/registration.rs
NunoAlexandre: good catch. fixed now 👍

NunoAlexandre:
@geigerzaehler I gave this some thought and having a wrapper type seems the way to go. We want to be able to say that x is a `Bytes128` and having that type of information guaranteeing the correctness of that data. Now, I don't follow the decoding part. Perhaps because I'm missing some context. 

I'd imagine that when a request sent to Registry involving a metadata field needs to be deserialized. During this deserialization or parsing step, we either can or can not parse the input data into our internal representation, i.e., Either<ParsingError, Value>. Within this specific problem, we would have a smart constructor in `Bytes128` that tries to build itself from an array of bytes (`Option<Bytes128>`). When that array exceeds the 128 elements, it'd return `None`.  In that case, the outer parsing should fail with a message indicating the metadata can not be processed and thus the request is rejected (like a 422 http error). 

In sum, I see this as a parsing matter, not decoding. You seem to agree though, so could you expand on that?

geigerzaehler:
The `String32` struct in `core` can serve as a template for how to implement `Bytes128`. Regarding the decoding: The problem is that we cannot split the process into two decoding and parsing since this is handled by substrate. The only thing that we can control is the decoding function which has roughly the signature `fn decode (&[u8]) -> Result<Bytes128, CodecError>`. There’s no option involved. Again it might help to take a look at `String32`

NunoAlexandre:
@geigerzaehler 

> since this is handled by substrate

That's the part I wasn't considering, it's clear now. As for the rest, _is goed_ 👍

NunoAlexandre:
@geigerzaehler all things we have discussed yesterday are addressed and thus the PR is ready to go. Please have another look.

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// This type is used to represent project metadata fields.
+///
+/// Radicle limits the size of the project metadata field to 128 bytes.
+/// To guarantee that at the type-level, a smart constructor is provided to check validity.
+#[derive(Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Bytes128(Vec<u8>);
+
+/// TODO review
+/// 1. From and To vec traits
```
geigerzaehler: You can delete these lines

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl Bytes128 {
+    const MAXIMUM_SUPPORTED_LENGTH: usize = 128;
+
+    pub fn from_vec(vector: Vec<u8>) -> Result<Self, String> {
```
geigerzaehler: Could you add some documentation? Basically saying that this fails if the vector has more than 128 items.

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    #[test]
+    fn decode_inordinate_vector_fails() {
+        // Encode a malformed bytes128 and verify that it fails to decode.
+        // Note that we use Bytes128(vec) instead of Bytes128::from_vec().
+        let inordinate_bytes128 = Bytes128(random_vector(129));
+        let encoded = inordinate_bytes128.encode();
+        let decoding_result = <Bytes128>::decode(&mut &encoded[..]);
+
+        assert!(decoding_result.is_ok())
```
geigerzaehler:
```suggestion
        assert!(decoding_result.is_err())
```

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+/// `Bytes128` type, and its validation tests.
+use alloc::format;
+use alloc::prelude::v1::*;
+use core::convert::TryFrom;
+use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
+
+/// This type is used to represent project metadata fields.
+///
+/// Radicle limits the size of the project metadata field to 128 bytes.
+/// To guarantee that at the type-level, a smart constructor is provided to check validity.
```
geigerzaehler: How do you feel about documenting this in a manner that does not require the reader to know about project data? I.e. being more abstract.

```
/// Byte vector that is limited to 128 bytes.
```

It would also be great to give a small code example that constructs and extracts `Bytes128`. With Rust, code examples are also tested: https://doc.rust-lang.org/rustdoc/documentation-tests.html

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+#[derive(Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Bytes128(Vec<u8>);
```
geigerzaehler: Could you also add a `From<Bytes128> for Vec<u8>` implementation?

core/src/bytes128.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+
+/// Bytes128 random functions useful for unit testing.
+/// Note that since these fuctions make use of rand,
+/// we need to guard with the std feature to
+/// be able to compile it for wasm.
+#[cfg(feature = "std")]
+impl Bytes128 {
+    pub fn random() -> Self {
```
geigerzaehler: Could you add some documentation for this method and `random_with_size`?

core/src/bytes128.rs
NunoAlexandre: Done

core/src/bytes128.rs
NunoAlexandre: Fixed already 👍

core/src/bytes128.rs
NunoAlexandre: I am all for it, actually. I only mentioned project data because string32 does too, so I didn't want to cause friction there. Will change!

core/src/bytes128.rs
NunoAlexandre: ✔️

core/src/bytes128.rs
NunoAlexandre: Done 👍

NunoAlexandre:
Thanks, @geigerzaehler! This was a good one to get in touch and learn a good deal of things. Thanks for your guidance 🙆‍♂

NunoAlexandre:
@xla ping to the Application team as for this breaking change.

PR #169, opened 2020-01-16
Update substrate to 4c4ee3d

geigerzaehler:


geigerzaehler:
> Although to note that `compat`s appeared in `node/src/service.rs`, it may make sense in the future to move the node to futures-0.3 as well

This change does remove future 0.1 from the node. It is now completely future 0.3 based.

PR #170, opened 2020-01-16
ci: Do not pollute /cache in base image

geigerzaehler:
To address an issue first raised in https://github.com/oscoin/ci/pull/41 we change the CI base image so that it does not write data to the `/cache` directory

To make it easier for other images to use our image as a base image we also don’t set `RUSTC_WRAPPER` and `CARGO_HOME` in the image. We set them in the CI script.

We also take the opportunity to update the rust toolchain.


NunoAlexandre:
> To make it easier for other images to use our image as a base image we also don’t set RUSTC_WRAPPER and CARGO_HOME in the image. We set them in the CI script.

That's a good move.

ci/base-image/Dockerfile
```diff
@@ -13,11 +13,7 @@ SHELL ["/bin/bash", "-c"]
 
 ENV RUSTUP_HOME=/usr/local/rustup \
     CARGO_HOME=/usr/local/cargo \
-    PATH=/usr/local/cargo/bin:$PATH \
```
NunoAlexandre: nice 👌

runtime/src/lib.rs
```diff
@@ -15,8 +15,9 @@
 
 //! The Substrate Node Template runtime. This can be compiled with `#[no_std]`, ready for Wasm.
 
-// The `impl_runtime_apis` macro produces code that clippy complains about.
+// The `impl_runtime_apis` and `construct_runtime` macros produce code that clippy complains about.
```
NunoAlexandre: This comment serves to explain why the lines below are necessary, right? I'd rather make that clearer.

PR #171, opened 2020-01-16
cli: Add project-transfer command

geigerzaehler:
We add the `project-transfer` command to the cli to transfer funds from a project to an arbitrary account.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -215,6 +215,56 @@ impl CommandT for Transfer {
 }
 
 #[derive(StructOpt, Debug, Clone)]
+/// Transfer funds from a project to a recipient. The author needs to be the owner of the project
+pub struct TransferProjectFunds {
+    /// Name of the project in the .rad domain
+    #[structopt(value_name = "project")]
+    project_name: ProjectName,
+
+    /// Recipient Account in SS58 address format
+    #[structopt(parse(try_from_str = parse_account_id))]
+    recipient: AccountId,
+    funds: Balance,
+}
+
+#[async_trait::async_trait]
+impl CommandT for TransferProjectFunds {
+    async fn run(&self, command_context: &CommandContext) -> Result<(), ()> {
+        let client = &command_context.client;
+        let transfer_fut = client
+            .sign_and_submit_message(
+                &command_context.author_key_pair,
+                message::TransferFromProject {
+                    project: (self.project_name.clone(), RAD_DOMAIN.clone()),
+                    recipient: self.recipient,
+                    value: self.funds,
+                },
+            )
+            .await
+            .unwrap();
+        println!("transferring funds...");
+        let project_registered = transfer_fut.await.unwrap();
+        match project_registered.result {
+            Ok(()) => {
+                println!(
+                    "transferred {} RAD from {}.{} to {} in block {}",
+                    self.funds,
+                    self.project_name,
+                    RAD_DOMAIN.clone(),
+                    self.recipient,
+                    project_registered.block,
+                );
+                Ok(())
+            }
+            Err(_) => {
+                println!("transaction failed in block {}", project_registered.block);
+                Err(())
+            }
+        }
+    }
+}
+
+#[derive(StructOpt, Debug, Clone)]
 /// Show the balance of an account
 pub struct ShowBalance {
     #[structopt(
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -169,7 +169,7 @@ impl CommandT for ShowGenesisHash {
 }
 
 #[derive(StructOpt, Debug, Clone)]
-/// Transfer some funds to recipient
+/// Transfer funds to recipient
 pub struct Transfer {
     #[structopt(parse(try_from_str = parse_account_id))]
     /// Recipient Account in SS58 address format
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -66,12 +66,12 @@ impl Args {
 
 #[derive(StructOpt, Debug, Clone)]
 enum Command {
-    RegisterProject(RegisterProject),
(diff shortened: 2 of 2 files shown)


cli/src/commands.rs
```diff
@@ -214,6 +214,58 @@ impl CommandT for Transfer {
     }
 }
 
+#[derive(StructOpt, Debug, Clone)]
+/// Transfer some funds from a project to recipient. The author needs to be the owner of the
```
NunoAlexandre: 1. nit: I'd rather have the two sentences in separate lines:

```
/// Transfer some funds from a project to a recipient.
// The author needs to be the owner of the
```

2. You miss the article 'a': to **a** recipient.

3. I'd drop `some` in `some funds`, but leave that up to you.

4. Biased from previous experiences, `ProjectTransfer` could sound like transferring/moving a project to another organization or something alike. `TransferProjectFunds` or `ProjectFundsTransfer`? Since this is a command, I'd prefer the first alternative, as it also follows the imperative naming in, for instance, `ShowBalance`.

cli/src/commands.rs
```diff
@@ -214,6 +214,58 @@ impl CommandT for Transfer {
...
+/// project
+pub struct ProjectTransfer {
+    /// Name of the project in the .rad domain
+    #[structopt(value_name = "project")]
+    project_name: ProjectName,
+
+    /// Recipient Account in SS58 address format
+    #[structopt(parse(try_from_str = parse_account_id))]
```
NunoAlexandre: ❓ what exactly is this doing?

cli/src/commands.rs
```diff
@@ -214,6 +214,58 @@ impl CommandT for Transfer {
...
+                    self.funds,
+                    self.project_name,
+                    RAD_DOMAIN.clone(),
+                    self.recipient,
+                    project_registered.block,
+                );
+                Ok(())
+            }
+            Err(_) => {
+                println!("transaction failed in block {}", project_registered.block,);
```
NunoAlexandre: unnecessary comma before the last parenthesis?

cli/src/main.rs
```diff
@@ -71,6 +71,7 @@ enum Command {
     ShowProject(ShowProject),
     ShowGenesisHash(ShowGenesisHash),
     Transfer(Transfer),
+    ProjectTransfer(ProjectTransfer),
```
NunoAlexandre: could we have these sorted alphabetically?

cli/src/main.rs
```diff
@@ -86,6 +87,7 @@ async fn main() {
         Command::ShowProject(cmd) => cmd.run(&command_context).await,
         Command::ShowGenesisHash(cmd) => cmd.run(&command_context).await,
         Command::Transfer(cmd) => cmd.run(&command_context).await,
+        Command::ProjectTransfer(cmd) => cmd.run(&command_context).await,
```
NunoAlexandre: Same here.

cli/src/commands.rs
```diff
@@ -214,6 +214,58 @@ impl CommandT for Transfer {
...
+    #[structopt(parse(try_from_str = parse_account_id))]
+    recipient: AccountId,
+    funds: Balance,
+}
+
+#[async_trait::async_trait]
+impl CommandT for ProjectTransfer {
+    async fn run(&self, command_context: &CommandContext) -> Result<(), ()> {
+        let client = &command_context.client;
+
```
NunoAlexandre: unnecessary empty line

cli/src/commands.rs
geigerzaehler: This defines a custom parser to get the account ID from a String from the command line. See the [`structopt` docs](https://docs.rs/structopt/0.3.8/structopt/index.html#custom-string-parsers) for more info.

cli/src/commands.rs
geigerzaehler: Nice catch. Not sure why `cargo fmt` didn’t take care of this.

NunoAlexandre: `cargo fmt` doesn't seem to remove hanging commands.

geigerzaehler:
> Looks good overall. I'm surprised to see `unwrap` used to often, and that it seems to be common practise in Rust.

I think using `unwrap` is often fine in the early development. Especially with the CLI we don’t benefit from the improved control flow if we were to pass the errors, just the output would be nicer. I’ll create a ticket though to improve this.

NunoAlexandre:
@geigerzaehler I agree that in the CLI we don't really benefit from being too safe, as long as the UX we provide in the case of invalid data or of an error is good.

cli/src/commands.rs
```diff
@@ -215,9 +215,8 @@ impl CommandT for Transfer {
 }
 
 #[derive(StructOpt, Debug, Clone)]
-/// Transfer some funds from a project to recipient. The author needs to be the owner of the
-/// project
-pub struct ProjectTransfer {
+/// Transfer funds from a project to a recipient. The author needs to be the owner of the project
```
NunoAlexandre: Sweet 👌

geigerzaehler:
> As for merging, I'd suggest you keep both commits since one introduces functional changes (a new feature) and the second mostly improves readability.

I try to avoid touching the same code in two subsequent commits in one PR since we’ll never really see the intermediate state and it will make it confusing to review later. Then the only separate change would really be the documentation change on `Transfer` which I think is small enough to include in that commit, too.

NunoAlexandre:
The renaming is indeed fine to merge, but the change in sorting the commands alphabetically not really.

geigerzaehler:
> The renaming is indeed fine to merge, but the change in sorting the commands alphabetically not really.

:+1: I split it into separate commits.

NunoAlexandre:
@geigerzaehler, awesome!

Issue #132, opened 2020-01-07
Rename “call” to “message” in client


geigerzaehler:
Replace the term “call” with “message”. This includes function names, struct names, variables names, modules, etc. ~~We also move the message structs in their own module. I.e. `CreateCheckpointParams` becomes `message::CreateCheckpoint`.~~ (Already done previously)

NunoAlexandre:
@geigerzaehler you appear to suggest to wrap the structs defined in `messages.rs` within a `mod`. Before just jumping into that solution, I'd like to know if there is a reason to not have defined these as an Enum? It seems more applicable and gives the namespace for free, so I wonder if there's a reason to have gone with separate structs?

geigerzaehler:
> I'd like to know if there is a reason to not have defined these as an Enum? It seems more applicable and gives the namespace for free, so I wonder if there's a reason to have gone with separate structs?

The reason is that we are using the message types as arguments to the runtime transaction handlers 
https://github.com/radicle-dev/radicle-registry/blob/350ad37a48abaca3a7b94ed9ca1020c2928821e7/runtime/src/registry.rs#L105
and that we have different result types for each transaction message
https://github.com/radicle-dev/radicle-registry/blob/350ad37a48abaca3a7b94ed9ca1020c2928821e7/client/src/call.rs#L63-L82

For this we need the message types to be first class structs.

NunoAlexandre:
Right, I see. thanks.

NunoAlexandre:
@geigerzaehler the scope of this issue seems to be within the `client` executable. However, the `runtime` also has the notion of `Call`. Shouldn't we apply the change in terminology there as well?

geigerzaehler:
> However, the `runtime` also has the notion of `Call`. Shouldn't we apply the change in terminology there as well?

The `Call` in `runtime` is substrate terminology and pretty much baked into substrate. So we should not change there.

PR #172, opened 2020-01-17
Rename call to message

NunoAlexandre:
Addresses #132 

## TODO

- [x] Rename client `call` module to `message`
- [x] Rename all `call` terminology to `message` in client
- [x] Qualify `messages` usages. I.e: `messages::RegisterProject` instead of just `RegisterProject`.
- [x] Drop `Params` suffix from messages
- [x] Qualify `messages` usages. I.e: `message::RegisterProject` instead of just `RegisterProject`.
- [x] Rename core/messages to core/message
- [x] Export core/message from client/message


NunoAlexandre:
@geigerzaehler reposting my question here:

the scope of this issue seems to be within the `client` executable. However, the `runtime` also has the notion of `Call`. Shouldn't we apply the change in terminology there as well? In any case, I am starting to question the added value of this change. Or if going from 'call' to 'transaction' wouldn't be better 🤔

xla:
@NunoAlexandre Transaction and the payload(s)/message(s) they carry are distinct. Overloading the terminology could lead to a lot of subtle problems. A transaction is always a signed tuple of fixed things (nonce, genesis of the chain and the payload/actual operation). In essence messages are the domain operations and transactions are the container for propagation and persistence.

NunoAlexandre:
@xla thanks, that makes perfect sense. On the `runtime` target there is `RuntimeCall` and I wonder if that, alongside with other `Call` types and terminology should follow the renaming. I need to give a step back and see the bigger picture.

NunoAlexandre:
@geigerzaehler I'd appreciate your review. Before, I recommend reading the comments above.

Without a full understanding of all Registry, it's difficult for me to pinpoint every single 'call' and say securely if it must be renamed to 'message' or not, but I'd convinced I got it (close to) right, so please review cautiously.

client/src/message.rs
```diff
@@ -39,7 +39,7 @@ pub trait Call: Send + 'static {
     fn into_runtime_call(self) -> RuntimeCall;
 }
 
-impl Call for RegisterProjectParams {
+impl Message for RegisterProjectParams {
```
geigerzaehler: One piece of information present in the issue that was not tackled is removing the `Prams` suffix from all the message types. I think it would also be nice to not export the message type (e.g. `RegisterProject`) on the top level but from the `message` module so that they are always used qualified like `message::RegisterProject`. I think this would add a lot of clarity to the code.

client/src/message.rs
```diff
@@ -13,24 +13,24 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Defines [Call] trait and implementations for all transaction parameters.
+//! Defines [Message] trait and implementations for all transaction parameters.
```
geigerzaehler: We can be a bit more precise here:
```suggestion
//! Defines [Message] trait and implementations for all messages in `radicle_registry_core::messages`.
```

client/src/message.rs
NunoAlexandre: noted. just came to face the `cli` `Command` enum that has the exact same names as the messages without the `Params` prefix, which collides. it's fixed by - as you suggest and I appreciate - qualifying the references.

xla:
@NunoAlexandre Are we sticking with the plural for the mod name `messages` instead of `message`?

NunoAlexandre:
@xla your points reveals an ambiguity issue. We have core/messages (plural) and client/message (singular, previously `call`). This issue involves dropping `Params` as suffix from all core/message**s**. Since those are globally exposed (::*), such a drop would result in an unresolvable collision. So I've qualified all core/messages references, now dropping the `Params` suffix, and next qualifying all the `client/message` references.

Does it make sense? the (core) `messages` vs  (client) `message` issue is something we should think better about, anyway. It will surely get in the way more often than not. @geigerzaehler thoughts?

NunoAlexandre:
>  I think it would also be nice to not export the message type (e.g. RegisterProject) on the top level but from the message module so that they are always used qualified like message::RegisterProject

@geigerzaehler I think I know to get what you were saying here. Tell me if wrong, I understand you are suggesting to export the core/messages module in the client/message module so to access the core/messages types from client/messages. I understand the idea but it seems counter-intuitive to have `messages` in `core` so to be shared across all executables but then expect it to be imported from the `client`. I like the idea but not sure if that's the best way to go, since nothing but agreed policy stops us from do `messages::` instead of `message::`.

Note that now, after https://github.com/radicle-dev/radicle-registry/pull/172/commits/57ddde80fa975eb8a111542ac44ce6eacb1675b0, all references moved to be qualified with `messages` (plural, from `core`). Explained [here](https://github.com/radicle-dev/radicle-registry/pull/172#issuecomment-575589527) why.

geigerzaehler:
I have no strong opinion on `messages` vs `message` as the module name. It’s up to you, @NunoAlexandre. We should just be consistent.

xla:
Still not convinced that `messages` is sensible. Is so that we can support wildcard imports in the cli? there should be no harm in referring to the message types qualified: `core::message::Foo`,`client::message::Bar`. It's rather not ergonomic that one is called `message` and the other is `messages`. Again I'm yet to find precedence for plural naming in Rust modules in the wild.

geigerzaehler:
Nuno addressed your concerns, @xla. Everything is now called `message` and imported qualified as `radicle_registry_client::message`. Same for `core`.

NunoAlexandre:
@xla as @geigerzaehler points out, our quick talk on the singular vs plural was clear and I have changed the code so that the singular mode is used both in `core` and in `client`.

xla:
Oh nevermind, disregard my comment. Sorry for the noise.

NunoAlexandre:
No worries!

PR #175, opened 2020-01-17
Use storage prefix APIs for project listing

geigerzaehler:
Instead of storing the project lists in a dedicated place in the chain state we use the storage prefix APIs to get the list of all projects.

We keep the private `Client::fetch_value` method alive since we’ll definitely need it in the future.

### DIFF
--- a/client/src/backend/mod.rs
+++ b/client/src/backend/mod.rs
@@ -59,6 +59,13 @@ pub trait Backend {
         block_hash: Option<BlockHash>,
     ) -> Result<Option<Vec<u8>>, Error>;
 
+    /// Fetch all keys with the given prefix from the state storage at the given block.
+    async fn fetch_keys(
+        &self,
+        prefix: &[u8],
+        block_hash: Option<BlockHash>,
+    ) -> Result<Vec<Vec<u8>>, Error>;
+
     /// Get the genesis hash of the blockchain. This must be obtained on backend creation.
     fn get_genesis_hash(&self) -> Hash;
 }
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -36,6 +36,7 @@ use std::sync::Arc;
 use parity_scale_codec::{Decode, FullCodec};
 
 use frame_support::storage::generator::{StorageMap, StorageValue};
+use frame_support::storage::StoragePrefixedMap;
 use radicle_registry_runtime::{balances, registry, Runtime};
 
 mod backend;
@@ -92,6 +93,7 @@ impl Client {
     /// ```ignore
     /// client.fetch_value::<frame_balance::TotalIssuance<Runtime>, _>();
     /// ```
+    #[allow(dead_code)]
     async fn fetch_value<S: StorageValue<Value>, Value: FullCodec + Send + 'static>(
         &self,
     ) -> Result<S::Query, Error>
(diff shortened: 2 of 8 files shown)


runtime/src/registry.rs
```diff
@@ -45,16 +43,46 @@ pub mod store {
...
+        /// let project_id = (
+        ///     String32::from_str("name").unwrap(),
+        ///     String32::from_str("domain").unwrap()
+        /// );
+        ///
+        /// let key = store::Projects::storage_map_final_key(project_id.clone());
+        /// let extracted_project_id = store::Projects::id_from_key(&key).unwrap();
+        /// assert_eq!(project_id, extracted_project_id)
+        /// ```
+        pub fn id_from_key(key: &[u8]) -> Option<ProjectId> {
```
gnunicorn: Why eat up the `Err` that `decode` might throw, giving you more information and wrap this into an `Option` instead?

client/src/lib.rs
```diff
@@ -222,7 +224,15 @@ impl ClientT for Client {
     }
 
     async fn list_projects(&self) -> Result<Vec<ProjectId>, Error> {
-        self.fetch_value::<registry::store::ProjectIds, _>().await
+        let project_prefix = registry::store::Projects::final_prefix();
+        let keys = self.backend.fetch_keys(&project_prefix, None).await?;
+        let mut project_ids = Vec::with_capacity(keys.len());
+        for key in keys {
+            let project_id = registry::store::Projects::id_from_key(&key)
+                .ok_or_else(|| Error::Other("Failed to decode ProjectID from key".to_string()))?;
```
gnunicorn: do you want to `Err` if any of the given keys doesn't map or do you just want to filter for the keys that do map to `id`? You are currently doing the first, if you want to do the second, you could instead do:

```rust
Ok(keys.iter().map(|key| registry::store::Projects::id_from_key(&key)).collect::<Vec_>())
```

runtime/src/registry.rs
```diff
@@ -45,16 +43,46 @@ pub mod store {
 
     decl_storage! {
         pub trait Store for Module<T: Trait> as Counter {
-            pub Projects: map ProjectId => Option<Project>;
+            // We use the twox_64_concat hasher so that the ProjectId can be extracted from the key
+            pub Projects: map hasher(twox_64_concat) ProjectId => Option<Project>;
```
gnunicorn: Please switch to `blake2_128_concat`–`twox`  is not user-input-safe.

client/src/lib.rs
geigerzaehler: I want the former. If this fails there is something wrong with the runtime state so nothing that we can recover from. I switched to `.expect` so we panic.

client/src/backend/mod.rs
```diff
@@ -59,6 +59,13 @@ pub trait Backend {
         block_hash: Option<BlockHash>,
     ) -> Result<Option<Vec<u8>>, Error>;
 
+    /// Fetch all keys with the given prefix from the state storage at the given block.
+    async fn fetch_keys(
+        &self,
+        prefix: &[u8],
+        block_hash: Option<BlockHash>,
```
rockbmb: This might be a moot question, but does this mean we want
* all keys with a given prefix that were present in the chain state at a given block, or
* all keys with a given prefix that were added at a certain block?

I know it's the former since I read the docs for this, but just making sure.

runtime/src/registry.rs
```diff
@@ -45,16 +43,48 @@ pub mod store {
...
+        /// );
+        ///
+        /// let key = store::Projects::storage_map_final_key(project_id.clone());
+        /// let extracted_project_id = store::Projects::id_from_key(&key).unwrap();
+        /// assert_eq!(project_id, extracted_project_id)
+        /// ```
+        pub fn id_from_key(key: &[u8]) -> Result<ProjectId, parity_scale_codec::Error> {
+            use parity_scale_codec::Decode;
+
+            let project_prefix = Self::final_prefix();
```
rockbmb: Where does `Self::final_prefix()` come from? A macro, right?

client/src/backend/mod.rs
geigerzaehler: Yes, it’s the former.

runtime/src/registry.rs
geigerzaehler: That method is available through the [`StoragePrefixedMap`](https://substrate.dev/rustdocs/master/frame_support/storage/trait.StoragePrefixedMap.html#method.final_prefix) implementation of `Projects`. And the implementation is generated by the `decl_storage!` macro.

PR #177, opened 2020-01-20
ci: Check Wasm build with clippy

geigerzaehler:
Check the Wasm build with `clippy` on CI. We also document the checks.


DEVELOPING.md
```diff
@@ -65,8 +66,23 @@ Packages
   it should probably go into `core`.
 
 
-Running tests
--------------
+Running checks and tests
+------------------------
+
+We check our code with clippy and `cargo fmt`
```
NunoAlexandre: I'd like to read _why_ we one type of check is not sufficient or a link to where that is explained.

PR #180, opened 2020-01-21
Cleanup `core` dependencies

geigerzaehler:
We remove all dependencies from `core` that are not needed. Some of them were just used for type aliases that we can also get from other dependencies.

Also regrouped the imports in the registry runtime code.

### DIFF
--- a/core/src/string32.rs
+++ b/core/src/string32.rs
@@ -18,9 +18,6 @@ use alloc::format;
 use alloc::prelude::v1::*;
 use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
 
-use sp_std::fmt;
-use sp_std::str::FromStr;
-
 /// Type to represent project names and domains.
 ///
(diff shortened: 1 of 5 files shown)


core/src/string32.rs
```diff
@@ -52,8 +49,8 @@ impl FromStr for String32 {
 }
 
 #[cfg(feature = "std")]
-impl fmt::Display for String32 {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+impl core::fmt::Display for String32 {
```
NunoAlexandre: Do we still need `#[cfg(feature = "std")]` now that we switch to `core::fmt`?

geigerzaehler: It would also work without the feature flag. However, we don’t want to include this code when we compile it to wasm to keep the code size small.

NunoAlexandre: Ok got it.

PR #182, opened 2020-01-21
Refactor registry errors to use DispatchError::Module

rockbmb:
This PR changes uses of `DispatchError::Other` in the registry's runtime to `DispatchError::Module` for a better description of errors that occur in invalid transactions.

Part of a larger effort to refactor how error handling is done on the client.

- [x] Refactor Registry transaction error handling to not use `DispatchError::Other(&'static str)`.
- [x] Create `enum` with Registry transaction failures (with statically generated messages and discriminant values for ease of use when converting to `DispatchError`).
- [ ] Refactor error handling on the client side to reflect the previous changes and allow "richer" errors.
- [ ] ?

I'm still not sure what shape it'll take, but `crate::Client::Error::InvalidTransaction` can be changed to receive the `DispatchError` produced by the a failed transaction in the registry.

### DIFF
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -1,0 +1,60 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::DispatchError;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+/// Errors describing failed Registry transactions.
+pub enum RegistryError {
+    InexistentCheckpointId = 0,
+    DuplicateProjectId,
+    InexistentProjectId,
+    InsufficientSenderPermissions,
+    InexistentParentCheckpoint,
+    InexistentInitialProjectCheckpoint,
+    InvalidCheckpointAncestry,
+}
+
+impl From<RegistryError> for &'static str {
+    fn from(error: RegistryError) -> &'static str {
+        match error {
+            RegistryError::InexistentCheckpointId => "The provided checkpoint does not exist",
+            RegistryError::DuplicateProjectId => "A project with a similar ID already exists.",
+            RegistryError::InexistentProjectId => "Project does not exist",
+            RegistryError::InsufficientSenderPermissions => "Sender is not a project member",
+            RegistryError::InexistentParentCheckpoint => "Parent checkpoint does not exist",
+            RegistryError::InexistentInitialProjectCheckpoint => {
+                "A registered project must have an initial checkpoint."
+            }
+            RegistryError::InvalidCheckpointAncestry => {
+                "The provided checkpoint is not a descendant of the project's initial checkpoint."
+            }
+        }
+    }
+}
+
+impl From<RegistryError> for DispatchError {
+    fn from(error: RegistryError) -> Self {
+        // This is the index with which the registry runtime module is declared
+        // in the Radicle Registry runtime - see the `construct_runtime`
+        // declaration in the `runtime` crate.
+        let registry_index = 7;
+        DispatchError::Module {
+            index: registry_index,
+            error: error as u8,
+            message: None,
+        }
+    }
+}
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -109,7 +109,7 @@ impl RemoteNode {
             Some(tx_status) => match tx_status {
                 TxStatus::Future | TxStatus::Ready | TxStatus::Broadcast(_) => (),
                 TxStatus::InBlock(_block_hash) => {
-                    return Err("Invalid tx status \"Finalized\"".into())
+                    return Err("Invalid tx status \"InBlock\"".into())
                 }
                 TxStatus::Usurped(_) => return Err("Extrinsic Usurped".into()),
                 TxStatus::Dropped => return Err("Extrinsic Dropped".into()),
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -39,6 +39,9 @@ pub use string32::String32;
 mod project_domain;
 pub use project_domain::ProjectDomain;
 
+mod error;
+pub use error::RegistryError;
+
 /// Index of a transaction in the chain.
 pub type Index = u32;
 
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -135,14 +135,15 @@ decl_module! {
             let sender = ensure_signed(origin)?;
 
             if store::Checkpoints::get(message.checkpoint_id).is_none() {
-                return Err(DispatchError::Other("The checkpoint provided to register the project does not exist."))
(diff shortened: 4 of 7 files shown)


geigerzaehler:
This is looking good already. A couple of comments
* We should probably have an enum with all the error cases. We can cast this into e `u8` error code easily by using [discriminant values](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations).
* We should probably statically generate the message from that error enum so to not specifiy it in the module methods.

rockbmb:
@geigerzaehler 
* Yes, that is the idea! I'm still considering how this might tie into `Client::Error::InvalidTransaction` - should it take the `enum` variant produced by the `(u8, u8)` tuple that uniquely identifies that error, or the other way around?
* That sounds like a good idea - how would you suggest this be done?

geigerzaehler:
> I'm still not sure what shape it'll take, but `crate::Client::Error::InvalidTransaction` can be changed to receive the `DispatchError` produced by the a failed transaction in the registry.

`InvalidTransaction` is probably not the right place. We’re dealing with ledger errors which are exposed through `TransactionApplied.result`. So we need to adjust the associated `Message::Result` types and `result_from_events` implementations to unwrap the dispatch error.

core/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+                "A registered project must necessarily have an initial checkpoint.",
+            RegistryError::InvalidCheckpointAncestry          =>
+                "The provided checkpoint ID is not a descendant of the project's initial checkpoint.",
+        }
+    }
+}
+
+// Due to a bug with Substrate, the `DispatchError::Module` variant that
+// results from a failed transaction appears with a `None` message, while
+// in actuality it should be `Some($MESSAGE)`.
```
geigerzaehler: This is not a bug but a conscious decision. The message is not included in the serialized version of `DispatchError` to save space. We can always recover the message through the error code. So I think it would make sense to remove the `error_msg` argument from `into_dispatch_error` and never include the message.

core/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::DispatchError;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+/// Errors describing failed Registry transactions.
+pub enum RegistryError {
+    // Implementation note: right now this `enum` is fairly small.
```
NunoAlexandre: I find this first line comment noisy.

core/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+    InexistentParentCheckpoint,
+    InexistentInitialProjectCheckpoint,
+    InvalidCheckpointAncestry,
+}
+
+impl From<RegistryError> for &'static str {
+    fn from(error: RegistryError) -> &'static str {
+        match error {
+            RegistryError::InexistentCheckpointId             =>
+                "The provided checkpoint does not exist",
```
NunoAlexandre: Haskeller alert!

I'd suggest doing, instead:

``` rust
RegistryError::InexistentCheckpointId
    => "The provided checkpoint does not exist",
```

That way, the names of the errors can grow without requiring us to keep moving all the `=>` along and it's easier on the eyes.

core/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+
+// Due to a bug with Substrate, the `DispatchError::Module` variant that
+// results from a failed transaction appears with a `None` message, while
+// in actuality it should be `Some($MESSAGE)`.
+//
+// Until this behavior is fixed, this enum is used in `into_dispatch_error` to
+// create `DispatchError::Module`s with different messages.
+pub enum RegistryErrorMessage {
+    Message,
+    NoMessage,
```
NunoAlexandre: Not sure I understand this. Isn't this a boolean as is? Or would use `Option<Message>` better?

NunoAlexandre: Also, note that you say "this behavior" but there is no context in those very docs as to what "this" is referring to.

rockbmb: It is a boolean, but I just thought it would be clearer what it does if it were an enum.

core/src/error.rs
rockbmb: Aargh, I did not know this had changed - however, was there not a point in the past where it was actually true the message was irrecoverable because of a bug?

core/src/error.rs
NunoAlexandre: Yeah, I get the idea, you don't want to go down the blind boolean road. I'd appreciate a better doc then or a better name. Is there a context as to why this feature exists / where/why sometimes we want to consider the error msg and other times not?

rockbmb: I thought I had written an explanation, turns out I did not. It's used in tests since the message doesn't show up there. I'll fix this (if it turns out we still need this change when the PR is done).

NunoAlexandre: Ah, I see! It could ease it to make a `Environment` enum which is either `Tests` or `Normal` or so. Then the logic is transparent: if running tests, no message, otherwise there is. Leave that up to you ✌️

core/src/error.rs
geigerzaehler: I don’t think we need the comment at all. These notes get outdated very quickly and I think we can trust the next developer to make the right call here.

core/src/error.rs
geigerzaehler: > however, was there not a point in the past where it was actually true the message was irrecoverable because of a bug?

Maybe we thought that it was a bug. But it is definitely intended

core/src/error.rs
NunoAlexandre: Subscribed

core/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::DispatchError;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+/// Errors describing failed Registry transactions.
+pub enum RegistryError {
+    InexistentCheckpointId = 0,
```
NunoAlexandre: @rockbmb why `= 0`?

core/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+            RegistryError::DuplicateProjectId
+                => "A project with the supplied ID already exists.",
+            RegistryError::InexistentProjectId
+                => "Project does not exist",
+            RegistryError::InsufficientSenderPermissions
+                => "Sender is not a project member",
+            RegistryError::InexistentParentCheckpoint
+                => "Parent checkpoint does not exist",
+            RegistryError::InexistentInitialProjectCheckpoint
+                => "A registered project must necessarily have an initial checkpoint.",
```
NunoAlexandre:
```suggestion
                => "A registered project must have an initial checkpoint.",
```

core/src/error.rs
rockbmb: @NunoAlexandre 

This is from a previous comment by @geigerzaehler :
>     * We should probably have an enum with all the error cases. We can cast this into e `u8` error code easily by using [discriminant values](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations).

There's more info in the link, but simply put, the `= 0` says that variants in this enum can be costlessly converted to integers (e.g. `error as u8`), and that their value is their index in the enum declaration, starting at 0 (another number can be chosen, but due to Rust rules, an enum cannot have more than 255 variants).

This is to tie into the `DispatchError::Module` variant that has a field named `error: u8`, which is where this is used - since each variant has a unique `u8` value, this means each registry error can be uniquely converted to a dispatch error (and back, but that's not needed in this PR).

Hope the answer was informative.

NunoAlexandre: Thanks, @rockbmb. That makes sense, your explanation is clear! I find it weird the way it's done though. [See here how Swift does this](https://stackoverflow.com/questions/35225449/using-enum-with-integers-in-swift), for instance.

PR #183, opened 2020-01-21
Document String32 limit in bytes

geigerzaehler:
We document that the `String32` limit is calculated in bytes and not in characters. We also improve the overall documentation of String32 and add doc tests.


rockbmb:
👍

PR #187, opened 2020-01-22
Create new ProjectDomain type

NunoAlexandre:
This new type wraps a String32 with a smart contract that ensures the
project domain is not only at most 32 bytes in length but also that it
is a supported domain by Radicle at this time. For now, we only support
the "rad" domain.

- [x] Create `ProjectDomain` wrapper type
- [x] Implement all useful conversion implementations
- [x] Provide a better error type 
- [x] Add unit tests
- [x] Replace current `ProjectDomain` alias with this new one
- [x] Handle failing `ProjectDomain` and report errors accordingly

## Questions

Q1: Does the domain support need to be case-insensitive?

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -15,13 +15,7 @@
 
 //! Defines [CommandT] trait, structs for all commands and their [CommandT] implementations.
 use radicle_registry_client::*;
-use std::str::FromStr;
 use structopt::StructOpt;
-
-lazy_static::lazy_static! {
-    static ref RAD_DOMAIN: String32 =
-        String32::from_str("rad").expect("statically valid");
-}
 
 /// Contextual data for running commands. Created from command line options.
 pub struct CommandContext {
@@ -53,13 +47,17 @@ impl CommandT for ShowProject {
         };
         let opt_project = command_context
             .client
-            .get_project((project_name.clone(), RAD_DOMAIN.clone()))
+            .get_project((project_name.clone(), ProjectDomain::rad_domain()))
             .await
             .unwrap();
 
         let project = match opt_project {
             None => {
-                println!("Project {}.{} not found", project_name, RAD_DOMAIN.clone());
+                println!(
+                    "Project {}.{} not found",
+                    project_name,
+                    ProjectDomain::rad_domain()
+                );
                 return Err(());
             }
             Some(project) => project,
@@ -125,8 +123,7 @@ impl CommandT for RegisterProject {
         let checkpoint_id = checkpoint_created.result.unwrap();
         println!("checkpoint created in block {}", checkpoint_created.block,);
 
-        let domain = String32::from_str("rad").expect("statically valid");
-        let project_id = (self.name.clone(), domain.clone());
+        let project_id: ProjectId = (self.name.clone(), ProjectDomain::rad_domain());
         let register_project_fut = client
             .sign_and_submit_message(
                 &command_context.author_key_pair,
@@ -144,7 +141,9 @@ impl CommandT for RegisterProject {
             Ok(()) => {
                 println!(
                     "project {}.{} registered in block {}",
-                    self.name, domain, project_registered.block,
+                    self.name,
+                    ProjectDomain::rad_domain(),
+                    project_registered.block,
                 );
                 Ok(())
             }
@@ -236,7 +235,7 @@ impl CommandT for TransferProjectFunds {
             .sign_and_submit_message(
                 &command_context.author_key_pair,
                 message::TransferFromProject {
-                    project: (self.project_name.clone(), RAD_DOMAIN.clone()),
+                    project: (self.project_name.clone(), ProjectDomain::rad_domain()),
                     recipient: self.recipient,
                     value: self.funds,
                 },
@@ -251,7 +250,7 @@ impl CommandT for TransferProjectFunds {
                     "transferred {} RAD from {}.{} to {} in block {}",
                     self.funds,
                     self.project_name,
-                    RAD_DOMAIN.clone(),
+                    ProjectDomain::rad_domain(),
                     self.recipient,
                     project_registered.block,
                 );
--- a/core/src/project_domain.rs
+++ b/core/src/project_domain.rs
@@ -1,0 +1,151 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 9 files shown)


NunoAlexandre:
@geigerzaehler regarding my last commit adding the `ProjectDomainError` type,  note this commit does not compile for wasm due to `std` being used by String32, which is transitive here since that's the type wrapped by ProjectDomain. More specifically, `String32::from_str` needs to be called and that's not available for wasm. Need to find a way to fix that. 

Also, is there another place but `std` where the `Error` trait can be found? It's not in `core`, unlikely other crates as you showed me earlier.

core/src/string32.rs
```diff
@@ -27,7 +27,7 @@ use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
 /// assert!(String32::from_string(long_string).is_err());
 /// ```
 #[derive(Encode, Clone, Debug, Eq, PartialEq)]
-pub struct String32(String);
+pub struct String32(pub String);
```
cloudhead: I don't have much context for why this change is being made, but I'd recommend instead to implement `Deref` on `String32`. Otherwise it circumvents the guarantees the type provides, because I can do `s.0 = "super long string"`.

FintanH: Or even a method: `fn value(&self) -> &str`?

NunoAlexandre: @cloudhead you are absolutely right, this is **not** a change I want to keep. Note this PR is a draft. I changed it to be able to proceed on the `ProjectDomain` side. I thought of either adding an unwrapping method as @FintanH suggests or implement the `ToString` trait. Could you expand on why you suggest to implement `Deref` and not, for instance, `ToString`?

cloudhead: :+1: 

`ToString` is normally not implemented by hand, but given for free when `Display` is implemented. However you could also implement `Into<String>`. `Deref` just offers the convenience of being able to call `String::*` methods directly on the `String32` type, without calling `into()`. If it's not something you need, then probably `Into<String>` is fine.

cloudhead: See: https://doc.rust-lang.org/std/string/trait.ToString.html

NunoAlexandre: Thanks for the explanation and for the link. I see. So Deref in a sense makes the more specific type a sort of subclass. Indeed, I don't need it in this case but good to know about it. Will implement the `Into<String>` then. Thanks!

NunoAlexandre: Implemented `Into` 👍

core/src/project_domain.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// The provided domain exceeds the 32 bytes length limit.
+    Inordinate,
+    /// The provided domain is not yet supported by Radicle.
+    NotYetSupported,
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ProjectDomainError {}
+
+impl core::fmt::Display for ProjectDomainError {
```
geigerzaehler: Probably a good idea to guard this to with the `std` feature because we don’t need it on the chain.

cli/src/commands.rs
```diff
@@ -59,7 +57,11 @@ impl CommandT for ShowProject {
 
         let project = match opt_project {
             None => {
-                println!("Project {}.{} not found", project_name, RAD_DOMAIN.clone());
+                println!(
+                    "Project {}.{:?} not found",
```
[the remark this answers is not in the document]
NunoAlexandre: I have a `Display` implementation, but on `core` instead of `std`, which is not picked up here. Is there a way of making that happen, or should I just change the impl to be against `std` instead?

core/src/project_domain.rs
NunoAlexandre: Is the rule of thumb to know what will be used on the chain? I get why this won't but would profit from knowing the bigger rule.

cli/src/commands.rs
```diff
@@ -15,12 +15,10 @@
 
 //! Defines [CommandT] trait, structs for all commands and their [CommandT] implementations.
 use radicle_registry_client::*;
-use std::str::FromStr;
 use structopt::StructOpt;
 
 lazy_static::lazy_static! {
-    static ref RAD_DOMAIN: String32 =
-        String32::from_str("rad").expect("statically valid");
+    static ref RAD_DOMAIN: ProjectDomain = ProjectDomain::rad_domain();
```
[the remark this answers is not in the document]
NunoAlexandre: I'm fine with that. I kept it so that we don't create a new ProjectDomain each time.

core/src/project_domain.rs
```diff
@@ -0,0 +1,143 @@
...
+    pub fn from_string32(domain: String32) -> Result<Self, ProjectDomainError> {
+        Self::from_string(domain.into())
+    }
+
+    pub fn rad_domain() -> ProjectDomain {
+        ProjectDomain(String32::from_str("rad").expect("statically valid"))
+    }
+}
+
+/// Possible Errors
```
NunoAlexandre:
```suggestion
/// Possible errors when attempting to build a ProjectDomain.
```

core/src/project_domain.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Encode, Clone, Debug, Eq, PartialEq)]
+pub enum ProjectDomainError {
+    /// The provided domain exceeds the 32 bytes length limit.
+    Inordinate,
```
NunoAlexandre: Shall anyone reviewing find "Inordinate" unhelpful, I suggest changing to "TooLong".

cli/src/commands.rs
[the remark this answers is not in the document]
geigerzaehler: I don’t see a `Display` implementation for `ProjectDomain`. Only for `ProjectDomainError` :thinking:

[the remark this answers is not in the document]
NunoAlexandre: AH, you're right! Good catch, thanks.

core/src/project_domain.rs
geigerzaehler: In general, if the runtime code compiles without a piece of code, then it is not needed and can be guarded. I think this is always true for `Display` implementations.

core/src/project_domain.rs
geigerzaehler: TIL “inordinate”

NunoAlexandre: Growing a richer language is a good thing, but making the code and docs obscure by using uncommon language might be unhelpful. Let me know if you prefer the suggested change.

core/src/project_domain.rs
NunoAlexandre: Guarded.

cli/src/commands.rs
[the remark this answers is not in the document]
NunoAlexandre: Fixed!

cli/src/commands.rs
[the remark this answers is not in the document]
NunoAlexandre: Done.

core/src/project_domain.rs
geigerzaehler: I think I have to get used to it but I kinda like it

PR #188, opened 2020-01-23
cli: Add CommandError and eliminate .unwrap()

### LINKED ISSUE #176: Eliminate `.unwrap()` in CLI
Calls to `Error::unwrap` are prevalent in the `run` implementation of the CLI commands. We should instead return these errors from `run`. This would require changing the error type returned by `run`.

geigerzaehler:
We add `CommandError` as the error for `CommandT::run` and provide `From` instances for client and dispatch errors.

We keep the `CommandError` very basic for now until we have revisited the client and ledger errors.

Closes #176

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -23,42 +23,85 @@ pub struct CommandContext {
     pub client: Client,
 }
 
+/// Error returned by [CommandT::run].
+///
+/// Implements [From] for client errors.
+#[derive(Debug, derive_more::From)]
+pub enum CommandError {
+    ClientError(Error),
+    FailedTransaction {
+        tx_hash: TxHash,
+        block_hash: BlockHash,
+    },
+    ProjectNotFound {
+        name: ProjectName,
+        domain: ProjectDomain,
+    },
+}
+
+impl core::fmt::Display for CommandError {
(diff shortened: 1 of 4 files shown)


cli/src/commands.rs
```diff
@@ -23,10 +23,30 @@ pub struct CommandContext {
     pub client: Client,
 }
 
+/// Error returned by [CommandT::run].
+///
+/// Wraps the message that is displayed to the user. We provide [From] instances for client [Error]
+/// and [DispatchError].
+#[derive(Clone, Debug)]
+pub struct CommandError(pub String);
```
NunoAlexandre: I am aware this is the CI and that these errors are not propagated down the road. However, we would benefit from having `CommandError` as an enum, where the variants would be: 

```enum CommandError {
    ClientError(Error).
    FailedTransaction(Block),
    InvalidProjectName(Error),
    ProjectNotFound
```

By then implementing `Display`, we would have a single place defining how these errors are displayed back to the user. Now we do it every time we build `CommandError`, a duplication that easily leads to a bad ux shall we happen to change one message and not keep the others consistent. Besides that point, it's also beneficial to have this type informing of all error types a user can encounter when interaction with our cli.

geigerzaehler: Yes, good piont. I’m using this approach now.

PR #189, opened 2020-01-24
Add ledger message documentation to core

geigerzaehler:
Following up on the discussion about the `registry-spec` package this is our first towards moving the information to the `core` package.

* Add documentation to all messages in `core/src/message.rs`
* Remove the corresponding message documentation from `registry-spec/src/lib.rs`
* Explain message documentation in `core/README.md`


core/src/message.rs
```diff
@@ -13,40 +13,111 @@
...
 use parity_scale_codec::{Decode, Encode};
 use sp_core::H256;
 
+/// Registers a project on the Radicle Registry with the given ID.
+///
+/// # State changes
+///
+/// If successful, a new [crate::Project] with the given properties is added to the state.
+///
+/// [crate::Project::members] is initialized with the transaction author.
```
NunoAlexandre:
```suggestion
/// [crate::Project::members] is initialized with the transaction author as the only member.
```

core/src/message.rs
```diff
@@ -13,40 +13,111 @@
...
+///
+/// [crate::Project::account_id] is generated randomly.
```
NunoAlexandre: Does that mean that say, project P is unregistered, then Project P' is registered, P' _can_ have the same `account_id` that P' had had?

core/src/message.rs
```diff
@@ -13,40 +13,111 @@
...
 
+/// Add a new checkpoint to the state.
+///
+/// # State changes
+///
+/// If successful, adds a new [crate::Checkpoint] with the given parameters to the state.
+///
+/// # State-dependent validations
+///
+/// If `previous_checkpoint_id` is provided a checkpoint with the given ID must exist in the state.
```
NunoAlexandre: Personally, I'd find this layout for these docs easier to read:

``` rust
/// Add a new checkpoint to the state.
///
/// # State changes
/// If successful, adds a new [crate::Checkpoint] with the given parameters to the state.
///
/// # State-dependent validations
/// If `previous_checkpoint_id` is provided, a checkpoint with the given ID must exist in the state.
```

NunoAlexandre:
```suggestion
/// The `previous_checkpoint_id` must exist in the state.
```

core/src/message.rs
geigerzaehler: Yes, but it is extremely unlikely. Almost as unlikely as a hash collision

core/src/message.rs
geigerzaehler: I’d like to keep the “if” because `previous_checkpoint_id` is an `Option`. I’d also prefer to be explicit about distinguishing IDs and their existence in the state.

core/src/message.rs
NunoAlexandre: Sure, just wondered if we generate a random account id different than any ever generated. Thanks

core/src/message.rs
NunoAlexandre: Ok. In that case, a comma would help:

```suggestion
/// If `previous_checkpoint_id` is provided, a checkpoint with the given ID must exist in the state.
```

geigerzaehler: The Rust Book [suggests](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html?highlight=doc,comment#making-useful-documentation-comments) a blank line after the heading. This style is also what the [standard library](https://doc.rust-lang.org/src/std/lib.rs.html#1-571) and other packages use.

NunoAlexandre: Shame, but TIL.

core/src/message.rs
```diff
@@ -13,40 +13,111 @@
...
+/// # State changes
+///
+/// If successful, `balance` is deducated from the transaction author account and added to the the
+/// recipient account. If the recipient account did not exist before, it is created.
+///
+/// The recipient account may be a user account or a project account.
+///
+/// # State-dependent validations
+///
+/// The author account must have a balance of at least `balance`.
```
rockbmb: Here and above, maybe disambiguate to

```suggestion
/// The author account must have a free balance of at least `balance`.
```

?

PR #191, opened 2020-01-24
Return proper error types bytes128 string32

### LINKED ISSUE #185: Return proper error type for conversion to Bytes128 and String32
Currently the errors returned for possible from conversions of vecs to `Bytes128` return a `String`. This makes it hard to keep appropriate context when these errors are propagated. A proper error like the client exposes for other cases would be helpful.

NunoAlexandre:
Closes #185

### Changes

- `Bytes128::from_vec` now returns `Result<Bytes128, InordinateVector>` instead of `Result<Bytes128, String>`
- `String32::from_string` now returns `Result<String32, InordinateString>` instead of `Result<String32, String>`

### Decisions

Initially,  I thought of having an `enum` for each of the types such as `String32Error` and `Bytes128Error` with all the error case possible. I decided to make things simpler and follow the _YAGNI_ principle. To date, the only reason for each of these types to fail to build is if their inputs are too long for their own requirements. Therefore I have both `InordinateString` and `InordinateVector` as a struct with a single field: the input, failing value. That way we can provide useful error messages in a single place too, by implementing `Display`.

### DIFF
--- a/core/src/bytes128.rs
+++ b/core/src/bytes128.rs
@@ -14,7 +14,6 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 /// `Bytes128` type, and its validation tests.
-use alloc::format;
 use alloc::prelude::v1::*;
 use core::convert::TryFrom;
 use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
@@ -29,13 +28,9 @@ impl Bytes128 {
     /// Smart constructor that attempts to build a Bytes128
     /// from a Vec<u8> with an arbitrary size. It fails if the
     /// input vector is larger than Bytes128::MAXIMUM_SUPPORTED_LENGTH.
-    pub fn from_vec(vector: Vec<u8>) -> Result<Self, String> {
+    pub fn from_vec(vector: Vec<u8>) -> Result<Self, InordinateVectorError> {
         if vector.len() > Self::MAXIMUM_SUPPORTED_LENGTH {
-            Err(format!(
-                "The provided vectors's length exceeded is {} bytes while Bytes128 is limited to {} bytes",
-                vector.len(),
-                Self::MAXIMUM_SUPPORTED_LENGTH
-            ))
+            Err(InordinateVectorError())
         } else {
             Ok(Bytes128(vector))
         }
@@ -43,7 +38,7 @@ impl Bytes128 {
 }
 
 impl TryFrom<Vec<u8>> for Bytes128 {
-    type Error = String;
+    type Error = InordinateVectorError;
 
     fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
         Bytes128::from_vec(value)
@@ -68,7 +63,7 @@ impl Bytes128 {
     }
 
     /// Generate a random Bytes128 vector with as many bytes as specified with 'size'.
-    pub fn random_with_size(size: usize) -> Result<Self, String> {
+    pub fn random_with_size(size: usize) -> Result<Self, InordinateVectorError> {
         Bytes128::from_vec(Self::random_vector(size))
     }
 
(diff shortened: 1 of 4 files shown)


core/src/bytes128.rs
```diff
@@ -86,6 +81,22 @@ impl Decode for Bytes128 {
     }
 }
 
+/// Error type for a failed attempt to build a Bytes128 value from an inordinate Vec<u8>.
+#[derive(Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InordinateVector(Vec<u8>);
```
geigerzaehler: The field is private and everything that we use from it is the length in the `Display` implementation. We could just replace this with the length so we don’t have to carry the whole vector around. Also note that the `Debug` implementation will print the whole vector. Probably not what we want.

core/src/string32.rs
```diff
@@ -77,24 +75,41 @@ impl Decode for String32 {
     }
 }
 
+/// Error type for a failed attempt to build a String32 value from an inordinate String.
+#[derive(Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InordinateString(String);
```
geigerzaehler: A practice that for example the standard library follows it to suffix all errors with `Error` and I’d like us to follow that too. I also think that the `String` is redundant in the name, so `InordinateError` would be ok. But I don’t have a strong opinion on the second point.

core/src/bytes128.rs
NunoAlexandre: That's right, the vector is only used to Display its length in the error msg. I thought of keeping the vector instead of just its length shall we need the vector down the road. In any case, by having the vector we can calculate its length but not the way around. That was my reasoning, but I also get yours so I will adjust.

core/src/string32.rs
NunoAlexandre: I once read a nice article stating that in we don't suffix our types with "Class", "Struct", "Enum", etc, there is no reason to do that with errors; That the need for such suffix derived by the name of the error type not being clear enough. "InordinateString" is obviously reporting a string that is inordinate, so "Error" is not needing to understand it. I think that's a good point. However, I believe code should be written within its environment and culture, and indeed suffixing error types with `Error` is common practise, so we better do that.

`InordinateError` sounds like it's the error being inordinate. I will make it `InordinateStringError` for the sake of clarity.

NunoAlexandre:
@geigerzaehler @xla considering Thoma's comment about just passing the exceeding input value's length with the error type, I wonder if even that is of any use? Wouldn't just having the error type suffice?

core/src/string32.rs
geigerzaehler: Good point about `InordinateError`

geigerzaehler:
> @geigerzaehler @xla considering Thoma's comment about just passing the exceeding input value's length with the error type, I wonder if even that is of any use? Wouldn't just having the error type suffice?

I’d prefer that, in fact. I don’t think the additional information is particularly useful.

core/src/string32.rs
xla: The `Error` suffix makes sense if it's a single struct, it becomes superfluous when naming varaints of an `Error` enum. [A lint for it](https://rust-lang.github.io/rust-clippy/master/index.html#pub_enum_variant_names).

xla:
@NunoAlexandre The error itself is telling enough we don't need it to carry any extra information at this point. Generally for these kinds of validaiton errors it's best to reserve an Error/variant per "rule" for the consumer to decide how they propagate it.

NunoAlexandre:
@geigerzaehler all addressed and already squashed. Thank you both for your feedback!

core/src/bytes128.rs
```diff
@@ -106,7 +116,10 @@ mod test {
         for size in Bytes128::MAXIMUM_SUPPORTED_LENGTH + 1..Bytes128::MAXIMUM_SUPPORTED_LENGTH + 10
         {
             let random_vector = random_vector(size);
-            assert!(Bytes128::from_vec(random_vector).is_err())
+            assert_eq!(
+                Bytes128::from_vec(random_vector.clone()),
```
geigerzaehler: Not sure why clippy doesn’t catch this but the `.clone()` is redundant.

core/src/string32.rs
NunoAlexandre: I like that.

core/src/bytes128.rs
NunoAlexandre: Fixed 👌

PR #205, opened 2020-02-06
Properly isolate and document core state entities

geigerzaehler:
As part of the restructuring of the registry spec and the `core` package we add the `radicle_registry_core::state` module and move all definitions of state entities there.

This PR is split into two commits: The first removes the `registry-spec` package as it is not needed anymore. The state objects and documentation are now part of `core::state`. The errors that were part of the `registry-spec` have not been migrated to a different place yet but they are outdated anyways.

The second commit adds the `state` module and the necessary documentation. We also document how to document state entities in the `core` README.


core/src/state.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Checkpoints are stored as a map using [Checkpoint::id] to derive the key.
+///
+/// # Invariants
+///
+/// * If `parent` is [Some] then the referenced checkpoint exists in the state.
+///
+/// # Relevant messages
+///
+/// * [crate::message::CreateCheckpoint]
+/// * [crate::message::SetCheckpiont]
```
NunoAlexandre: I wouldn't mention the messages here, to not mention higher-level matters in a lower-level layer where a type is defined. But I leave it up to you.

core/src/state.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// Next index (nonce) for a transaction of an account.
+///
+/// The index for an [crate::AccountId] increases whenever a transaction by the account owner is
+/// applied.
+///
+/// # Storage
+///
+/// Indicies are stored as a map with a key derived from [crate::AccountId].
+pub type Index = u32;
```
NunoAlexandre: File-level comment: I understand the idea behind this state module, but by categorizing these types that way, as the state will grow, this module will need to grow along, which will make this a dump module. We can agree to deal with that later, but it will surely be something we will have to go through at some point.

runtime/tests/checkpoint_creation.rs
```diff
@@ -37,7 +37,7 @@ async fn create_checkpoint() {
     .result
     .unwrap();
 
-    let checkpoint1_ = Checkpoint {
+    let checkpoint1_ = state::Checkpoint {
```
NunoAlexandre: Regarding all the `state::`, I am not sure, tbh. If we'd have multiple `Checkpoint`, `Project`, etc, that had different meanings across the different executables this would be a must. But now it just feels like it's adding unnecessary verbosity to the code, more diffs, more to write.

core/src/state.rs
geigerzaehler: My intention here is to make it easy to map out the ledger behavior for readers. You often want to know what messages can change the state.

Could you elaborate on the high-level/low-level comment? In my understanding messages and state are of equal importance and very closely related. The state defines the shape of the ledger state machine and the messages define the possible changes.

core/src/state.rs
geigerzaehler: Do you have any ideas about splitting this up? I’m ok with how it is now, especially considering that we only include entities that are actually stored in the state and no consituent types. The same comment also applies to the messages.

runtime/tests/checkpoint_creation.rs
geigerzaehler: I was pondering the same thing and I’d be ok with doing `pub use state::*` at the `core` top-level. What prevented me from this is that it is more consistent with how the `message` module is used and that it adds some additional qualification so the code is easier to parse.

Another thing to consider is that we might have different views of a project. For example there is no reason to store the project ID in the state since it can be extracted from the state key. But for the client API it might make more sense to return a `Project` that includes the ID. Here we would have different projects.

Does this change your perspective? Otherwise, I’ll drop the `state` qualifier.

NunoAlexandre: > I was pondering the same thing and I’d be ok with doing pub use state::* at the core top-level. What prevented me from this is that it is more consistent with how the message module is used and that it adds some additional qualification so the code is easier to parse.

I also thought about this, but the difference there is that we have a `message` module both in `core` and in `client`.

> Another thing to consider is that we might have different views of a project. For example there is no reason to store the project ID in the state since it can be extracted from the state key. But for the client API it might make more sense to return a Project that includes the ID. Here we would have different projects.

I get your point. Before moving forward, we have to agree on terminology given the new spec. As of now, `ProjectId` is an alias for `(ProjectName, ProjectDomain)`. With the new spec, `ProjectDomain` is gone and the `id` of a project is `Org` dependent. So in my current work moving `Project` to the new spec, I'm calling `ProjectId` what was `ProjectName`, just like the spec does. We can discuss this further, but here and now I just wanted to clarify this part.

Having this in mind, now a `Project` should be stored in the storage having `(ProjectId, OrgId)` as its key. I'm not sure I get your comment  "there is no reason to store the project ID in the state since it can be extracted from the state key.". It is true that those values can be obtained from the key, but the reason to have the redundancy is to keep things simple. Not doing so _requires_ having a specialized `Project` type only used to be (de)serialized. That would introduce unnecessary complexity.  If you mean that as a mere example of a scenario, then yes, that can indeed happen or be beneficial to have different representations of a type and in such a case keep the `state::` qualifier is helpful.

> Does this change your perspective? Otherwise, I’ll drop the state qualifier.

Yes and no. I am fine with keeping the `state` qualifier, not quite so fond of the example. I do see the point of keeping it though, so let's do that.

core/src/state.rs
NunoAlexandre: I'd like to have a `state` namespace, but not necessarily have everything that's stored in this single `state` module. Let's not worry about this now though, let's see how it envolves and adjust accordingly in the future.

runtime/tests/checkpoint_creation.rs
geigerzaehler: > > Does this change your perspective? Otherwise, I’ll drop the state qualifier.
> 
> Yes and no. I am fine with keeping the `state` qualifier, not quite so fond of the example. I do see the point of keeping it though, so let's do that.

Cool. I get your point about the example. I haven’t thought this through fully and will expand on this when we actually want to change this.

core/src/state.rs
NunoAlexandre: > Could you elaborate on the high-level/low-level comment?

Thinking of Lego, a `Checkpoint` would be a block (low-level) and the state would be a context where it is used but unrelated to the type itself. By other words, the type `Checkpoint` has no idea where and it is used for.

In that sense, having these comments here is breaking that abstraction layer or so. 

> My intention here is to make it easy to map out the ledger behaviour for readers.

Yes, you are right on this. Leave it as it, dismiss my comment and sorry for the noise.

NunoAlexandre:
@geigerzaehler I'd prefer to keep the two commits instead of squashing them into one.

PR #206, opened 2020-02-06
Improve on transaction state documentation with example

geigerzaehler:
Improve on transaction state documentation with example

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -63,8 +63,33 @@ pub type Response<T, Error> = BoxFuture<'static, Result<T, Error>>;
 pub trait ClientT {
     /// Submit a signed transaction.
     ///
-    /// Succeeds if the transaction has been accepted by the node. The wrapped future that is
-    /// returned can be used to wait for the transaction to be applied and included in a block.
+    /// ```no_run
+    /// # use radicle_registry_client::*;
+    /// # async fn example<M: Message>(client: Client, tx: Transaction<M>) -> Result<(), Error> {
+    ///
+    /// // Submit the transaction to the node.
+    /// //
+    /// // If this is successful the transaction has been accepted by the node. The node will then
+    /// // dissemniate the transaction to the network.
+    /// //
+    /// // This call fails if the transaction is invalid or if the RPC communication with the node
+    /// // failed.
+    /// let applied_fut = client.submit_transaction(tx).await?;
+    ///
+    /// // We can now wait for the transaction to be included in a block.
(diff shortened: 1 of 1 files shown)


client/src/interface.rs
```diff
@@ -65,6 +65,29 @@ pub trait ClientT {
     ///
     /// Succeeds if the transaction has been accepted by the node. The wrapped future that is
     /// returned can be used to wait for the transaction to be applied and included in a block.
+    ///
+    /// ```no_run
+    /// # use radicle_registry_client::*;
+    /// # async fn example<M: Message>(client: Client, tx: Transaction<M>) -> Result<(), Error> {
+    ///
+    /// // Submit the transaction. If this is successful the transaction is valid, has been
+    /// // accepted by the node and will be disseminated to the network.
```
NunoAlexandre: This is kind of hard to read and it repeats what's already said above:

**Above it says:**
> Succeeds if the transaction has been accepted by the node. The wrapped future that is
> returned can be used to wait for the transaction to be applied and included in a block.

**Here you now add:**
> Submit the transaction. If this is successful the transaction is valid, has been
> accepted by the node and will be disseminated to the network.
> ...
> We can now wait for the transaction to be included in a block.

Could you improve on this?

geigerzaehler: By “hard to read” do you mean that it is a comment within a code example? The alternative would be to have the comment as part of the method doc and have the code separately. The down-side is that it will be harder to get the example to compile.

Do you have any other suggestion how to improve this? I can only think of dropping the first explanation which is indeed very confusing.The documentation in the example reads fine for me, but I’m also very familiar with what it does.

NunoAlexandre: > By “hard to read” do you mean that it is a comment within a code example?

no, sorry. I meant that the sentence itself is difficult to read. I'd suggest improving it to: 

> Submit the transaction. When successful, the transaction:
> 1. is valid
> 2. has been accepted by the node
> 3. will be disseminated to the network 

Or

> A successful transaction means the transaction is valid, that it has been accepted by the node, and that it will be disseminated to the network.

NunoAlexandre: Regarding

>I can only think of dropping the first explanation which is indeed very confusing

Let's also do that then.

geigerzaehler: Thanks! I’ve implemented your suggestions and hopefully made the comments more readable. Could you check again?

NunoAlexandre: Thanks!

PR #208, opened 2020-02-06
Adjust projects to orgs

NunoAlexandre:
### Note

This draft PR serves mostly for discussion on some key aspects we need to polish regarding moving the registry projects to the latest spec.

### Changes

- Create `Org` type to serve as support to move Project to be under an org
- Adjust `Project` to be under an org
- Drop `ProjectDomain`
- Change `type ProjectId = (ProjectName, ProjectDomain)` to `type ProjectId = String32` (#A)
- Store a project using the tuple `(ProjectId, OrgId)` as the key (#B)
- Adjusted the whole chain of calls to work with the two arguments `project_id: ProjectId, org_id: OrgId` instead of just `project_id: ProjectId` (where, on the latter, `ProjectId = (ProjectName, ProjectDomain)`.

#### #A

I'm following the spec in terms of naming here. What we called `ProjectName` was actually an `id`, but there is also the notion of "id" as being the primary key of a Project in the storage. In that sense, the project id was indeed the tuple (ProjectName, ProjectDomain). Now that ProjectDomain is gone, having ProjectId is in a way sensible - it's the project unique identifier, within the context of an org; and in a way ambiguous, because it can't be used as a primary key. We need to use the tuple (ProjectId, OrgId) for that (as done in this PR already). 

An alternative could be to rename this PR's `ProjectId` to `ProjectName` and make `type ProjectId = (ProjectName, OrgId)`.

#### #B

One requirement regarding Orgs is to limit the number of projects an org can register. If we store projects as map `(ProjectId, OrgId) -> Project`, in order to know how many projects are registered under an org we would need to go through every key in the storage, filter the ones where the second element is a given org id, and, finally, count. That means O(n) complexity, where `n` is the absolute number of registered projects in the registry. Shall we have millions, that's a serious problem that we better think of earlier than later. Especially given the non-trivial problem of updating schemes and data on a chain.

I will think further about this. Let me know if you happen to have a solution already.

Let me have your opinion and thoughts on this.

  
### Mergeability

This PR isn't really mergeable because we can't validate `SetCheckpoint`, since we'd need to have registered Orgs and verify whether the sender of the message is a member of the organization it affects. We can agree to merge it anyway and fix it after implementing Orgs, though. I should have seen that coming, my bad! 

Also, the fund transactions relative to projects need to be deleted but I was leaving them until now to have them as an example when implementing those for Orgs, fyi.


geigerzaehler:
On A: My take on this is that we should strive for alignment with the paper spec. So let’s keep it like it is for know. However, I agree that project ID is confusing because it only identifies the project within an organization. I’d prefer to use `ProjectName`. This would also eliminate the need to introduce the strangley named `GlobalProjectId` type. Let’s discuss this in an issue on the [registry spec](https://github.com/radicle-dev/registry-spec).

On B: If you look at section 3 of the spec you can see that the org state has a project component. The spec also tells us that registering a project adds the project to the orgs project set. This means we can just use the size of this set to determine the number of projects.

On mergeability: A way to address this would be to implement orgs as a concept first without tying it to projects. Then it would be possible to tackle all outstanding issues in this PR. We could also have a separate PR for switching the funds from orgs to projects.

NunoAlexandre:
> On A: My take on this is that we should strive for alignment with the paper spec. So let’s keep it like it is for know. However, I agree that project ID is confusing because it only identifies the project within an organization. I’d prefer to use ProjectName. This would also eliminate the need to introduce the strangley named GlobalProjectId type. Let’s discuss this in an issue on the registry spec.

Perfect.

> On B: If you look at section 3 of the spec you can see that the org state has a project component. The spec also tells us that registering a project adds the project to the orgs project set. This means we can just use the size of this set to determine the number of projects.

Somewhat offtopic: I am not sure whether the spec means the project set to be `Set<ProjectId>` or `Set<Project>`. On my draft `Org` type in this PR, I added it as `pub projects: Vec<ProjectId>`. I thought the meaning was for that field to be a set of project IDs and not a set of projects themselves, to avoid having Org being this (likely) huge value being passed around. Are we on the same page up to this point, @geigerzaehler and @cloudhead, or do you think this field should be a Set<Project> instead?

As for your comment, you're right! Once we start using this Org.projects field then we can indeed calculate the number of projects from an org more optimally.

> On mergeability: A way to address this would be to implement orgs as a concept first without tying it to projects. Then it would be possible to tackle all outstanding issues in this PR. We could also have a separate PR for switching the funds from orgs to projects.

Agreed!

cloudhead:
Regarding the projectID/name, I agree that it should really be `ProjectID = (OrgID, ProjectName)`, so that IDs are always globally unique identifiers. I'm happy to update the spec.

Regarding the set of projects, the spec isn't intended to force an exact implementation, but rather to hint that formally, an org is a set of projects. In terms of the implementation, it's perfectly reasonable to only store the IDs in there however.

I do think that `HashSet<ProjectID>` (or `Hashset<ProjectName>`) would be more correct than using a `Vec`, since it guarantees uniqueness.

NunoAlexandre:
> Regarding the projectID/name, I agree that it should really be ProjectID = (OrgID, ProjectName), so that IDs are always globally unique identifiers. I'm happy to update the spec.

👌 

> Regarding the set of projects, the spec isn't intended to force an exact implementation, but rather to hint that formally, an org is a set of projects. In terms of the implementation, it's perfectly reasonable to only store the IDs in there however.

Perfect. I assumed it to be that way but it's good to have it confirmed.

> I do think that HashSet<ProjectID> (or HashSet<ProjectName>) would be more correct than using a Vec, since it guarantees uniqueness.

Absolutely. It guarantees uniqueness and O(1) to check if an element is part of the set.

Thanks!

NunoAlexandre:
Closing this as it won't move forwards as a PR.

Issue #198, opened 2020-02-04
Transaction: Unregister an Org


NunoAlexandre:
![Screenshot 2020-02-04 at 13 56 10](https://user-images.githubusercontent.com/7075260/73746715-1c8d4000-4756-11ea-830a-dd117f37bba1.png)

NunoAlexandre:
Done ✅

Issue #195, opened 2020-02-04
Transaction: Register an Org


NunoAlexandre:
![Screenshot 2020-02-04 at 13 23 37](https://user-images.githubusercontent.com/7075260/73744669-8fe08300-4751-11ea-9f70-dd8aaecfa985.png)

NunoAlexandre:
Done ✅

PR #211, opened 2020-02-07
Feature/orgs [Part 1]

### LINKED ISSUE #194: Implement Org type
Screenshot 2020-02-04 at 13 21 22

### Scope

This issue is meant to only tackle the implementation of the `Org` type, not it's integration with the other types neither transactions involving it.

NunoAlexandre:
Closes #194, #195, #198, #212

### Changes

- Implement `Org` type (#194)
- Add `RegisterOrg` message ( #195)
- Add `UnregisterOrg` message (#198)
- Unit test both new messages
- Support `RegisterOrg` and `UnregisterOrg` from the cli (#212)

### Decisions

Regarding the `Org` type, I have opted for `Vec` instead of `HashSet` for the Org members and projects fields for the reason I explain at 72c63a3896eb25c96ace77082abc3c5c921f38e6:

> Note that we opt for Vec for members and projects since substrate
doesn't offer Encode/Decode implementations for HashSet. To be able to
use HashSet we would need to create wrapper types around
HashSet<AccountId> and HashSet<ProjectName> since it's not possible to
implement a trait for a type foreigner to the current crate.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -136,6 +136,64 @@ impl CommandT for ListProjects {
 }
 
 #[derive(StructOpt, Debug, Clone)]
+/// Register an org.
+pub struct RegisterOrg {
+    /// Id of the org to register.
+    id: OrgId,
+}
+
+#[async_trait::async_trait]
+impl CommandT for RegisterOrg {
+    async fn run(&self, command_context: &CommandContext) -> Result<(), CommandError> {
+        let client = &command_context.client;
+
+        let register_org_fut = client
+            .sign_and_submit_message(
+                &command_context.author_key_pair,
+                message::RegisterOrg {
+                    id: self.id.clone(),
+                },
+            )
+            .await?;
+        println!("Registering org...");
+
+        let org_registered = register_org_fut.await?;
+        transaction_applied_ok(&org_registered)?;
+        println!("Org {} is now registered.", self.id);
+        Ok(())
+    }
+}
+
+#[derive(StructOpt, Debug, Clone)]
+/// Unregister an org.
+pub struct UnregisterOrg {
+    /// Id of the org to unregister.
+    id: OrgId,
+}
+
+#[async_trait::async_trait]
+impl CommandT for UnregisterOrg {
+    async fn run(&self, command_context: &CommandContext) -> Result<(), CommandError> {
+        let client = &command_context.client;
+
+        let register_org_fut = client
+            .sign_and_submit_message(
+                &command_context.author_key_pair,
+                message::UnregisterOrg {
+                    id: self.id.clone(),
+                },
+            )
+            .await?;
+        println!("Unregistering org...");
+
+        let org_unregistered = register_org_fut.await?;
+        transaction_applied_ok(&org_unregistered)?;
+        println!("Org {} is now unregistered.", self.id);
+        Ok(())
+    }
+}
+
+#[derive(StructOpt, Debug, Clone)]
 /// Register a project under the default "rad" domain.
 pub struct RegisterProject {
     /// Name of the project to register.
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -20,6 +20,8 @@
 
 extern crate alloc;
 
+use alloc::vec::Vec;
+
 use sp_core::{ed25519, H256};
 use sp_runtime::traits::BlakeTwo256;
 
@@ -55,5 +57,38 @@ pub type Balance = u128;
 pub type ProjectName = String32;
 
 pub type ProjectId = (ProjectName, ProjectDomain);
+
+pub type OrgId = String32;
+
+/// Org
+///
+/// Different from [state::Org] in which this type gathers
+/// both the id and the other Org fields, respectively stored
+/// as an Org's storage key and data, into one complete type.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Org {
+    // Id of an Org.
+    pub id: OrgId,
+
+    /// See [state::Org::account_id]
+    pub account_id: AccountId,
+
+    /// See [state::Org::members]
+    pub members: Vec<AccountId>,
+
+    /// See [state::Org::projects]
+    pub projects: Vec<ProjectName>,
+}
+
+impl Org {
+    pub fn from(id: OrgId, org: state::Org) -> Org {
+        Org {
+            id,
+            account_id: org.account_id,
+            members: org.members,
+            projects: org.projects,
+        }
+    }
+}
 
 pub type CheckpointId = H256;
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -18,8 +18,43 @@
 //! See the README.md for more information on how to document messages.
 extern crate alloc;
 
-use crate::{AccountId, Balance, Bytes128, CheckpointId, ProjectId, H256};
+use crate::{AccountId, Balance, Bytes128, CheckpointId, OrgId, ProjectId, H256};
 use parity_scale_codec::{Decode, Encode};
+
+/// Registers an org on the Radicle Registry with the given ID.
+///
+/// # State changes
+///
+/// If successful, a new [crate::state::Org] with the given properties is added to the state.
+///
+/// [crate::state::Org::members] is initialized with the transaction author as the only member.
+///
+/// [crate::state::Org::account_id] is generated randomly.
+///
+/// # State-dependent validations
+///
+/// An Org with the same ID must not yet exist.
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct RegisterOrg {
+    pub id: OrgId,
+}
+
+/// Unregisters an org on the Radicle Registry with the given ID.
+///
+/// # State changes
+///
+/// If successful, the targeted Org is removed from the state.
+///
+/// # State-dependent validations
+///
+/// The targeted org must exist and have no projects and the
+/// the transaction origin must be its only member.
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct UnregisterOrg {
+    pub id: OrgId,
+}
 
 /// Registers a project on the Radicle Registry with the given ID.
 ///
--- a/core/src/state.rs
+++ b/core/src/state.rs
@@ -14,11 +14,12 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 //! Type definitions for all entities stored in the ledger state.
+
 use alloc::vec::Vec;
 use parity_scale_codec::{Decode, Encode};
 use sp_runtime::traits::Hash;
 
-use crate::{AccountId, Balance, Bytes128, CheckpointId, Hashing, ProjectId, H256};
+use crate::{AccountId, Balance, Bytes128, CheckpointId, Hashing, ProjectId, ProjectName, H256};
 
 /// A checkpoint defines an immutable state of a project’s off-chain data via a hash.
 ///
@@ -72,7 +73,7 @@ pub struct Project {
     /// ID of the project. The storage key is derived from this value.
     pub id: ProjectId,
 
-    /// Account ID that holds the projecs founds.
+    /// Account ID that holds the projecs funds.
     ///
     /// This is randomly generated and unlike for other accounts there is no private key that
     /// controls this account.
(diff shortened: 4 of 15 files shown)


cli/src/commands.rs
```diff
@@ -135,6 +135,60 @@ impl CommandT for ListProjects {
     }
 }
 
+#[derive(StructOpt, Debug, Clone)]
+pub struct RegisterOrg {
+    id: OrgId,
```
geigerzaehler: Please add some docs for the struct and field. Since we’re using `StructOpt` these docs will be used to generate the CLI help.

core/src/state.rs
```diff
@@ -117,3 +120,15 @@ pub type AccountBalance = Balance;
...
 /// Indicies are stored as a map with a key derived from [crate::AccountId].
 pub type Index = u32;
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Org {
+    pub id: OrgId,
+    pub account_id: AccountId,
+    pub members: Vec<AccountId>,
+    pub projects: Vec<ProjectName>,
+    pub contract: Contract,
```
geigerzaehler: I would omit this field for now. Contracts are very much underspecified and we don’t know how they will look yet.

runtime/src/registry.rs
```diff
@@ -84,6 +89,33 @@ pub mod store {
...
+        pub fn id_from_key(key: &[u8]) -> Result<OrgId, parity_scale_codec::Error> {
+            use parity_scale_codec::Decode;
+
+            let prefix = Self::final_prefix();
+            // Length of BlakeTwo128 output
+            let key_hash_prefix_length = 16;
+            let key_prefix_length = prefix.len() + key_hash_prefix_length;
+            let mut id_bytes = &key[key_prefix_length..];
+            OrgId::decode(&mut id_bytes)
+        }
```
geigerzaehler: This code is duplicated in `Projects`. Will the project implementation vanish? If not we should extract the duplication.

runtime/src/registry.rs
```diff
@@ -162,6 +194,56 @@ decl_module! {
...
+            let sender = ensure_signed(origin)?;
+
+            match store::Orgs::get(message.id.clone()) {
+                None => {},
+                Some(_) => return Err(RegistryError::DuplicateOrgId.into()),
+            }
+
+            let random_account_id = AccountId::unchecked_from(
+                pallet_randomness_collective_flip::Module::<T>::random(
+                    b"project-account-id",
```
geigerzaehler:
```suggestion
                    b"org-account-id",
```

core/src/state.rs
```diff
@@ -117,3 +120,15 @@ pub type AccountBalance = Balance;
 ///
 /// Indicies are stored as a map with a key derived from [crate::AccountId].
 pub type Index = u32;
+
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Org {
```
geigerzaehler: Don’t forget the documentation.

core/src/state.rs
```diff
@@ -117,3 +120,15 @@ pub type AccountBalance = Balance;
...
+    pub id: OrgId,
```
geigerzaehler: Is there a reason we store the ID alongside the org? I don’t think there is a need that arises from the spec. This is similar to how projects are implemented but I believe this is not necessary there either.

core/src/message.rs
```diff
@@ -18,9 +18,44 @@
...
+/// Unregisters an org on the Radicle Registry with the given ID.
+///
+/// # State changes
+///
+/// If successful, the targeted Org is removed from the state.
+///
+/// # State-dependent validations
+///
+/// The targeted org must have no projects and the transaction origin
+/// must be its only member.
```
geigerzaehler: And the targeted organization must exist

runtime/src/registry.rs
```diff
@@ -162,6 +194,56 @@ decl_module! {
...
+            store::Orgs::insert(new_org.id.clone(), new_org.clone());
+            Self::deposit_event(Event::OrgRegistered(new_org.id));
+            Ok(())
+        }
+
+        #[weight = SimpleDispatchInfo::FreeNormal]
+        pub fn unregister_org(origin, message: message::UnregisterOrg) -> DispatchResult {
+            fn can_be_unregistered(org: state::Org, sender: AccountId) -> bool {
+                org.members == vec![sender] && org.projects.is_empty()
+            }
```
geigerzaehler: Is there a reason to have this as a closure? Can’t we just inline the code? It is only used once.

test-utils/src/lib.rs
```diff
@@ -75,6 +75,12 @@ pub fn random_register_project_message(checkpoint_id: CheckpointId) -> message::
     }
 }
 
+pub fn random_register_org_message() -> message::RegisterOrg {
+    let org_id = String32::from_string(random_string(32)).unwrap();
+
+    message::RegisterOrg { id: org_id }
+}
```
geigerzaehler: How about replacing this with a `random_string32` function and constructing the message from that?

runtime/tests/org_registration.rs
```diff
@@ -0,0 +1,131 @@
...
+use radicle_registry_client::*;
+use radicle_registry_test_utils::*;
+
+#[async_std::test]
+async fn register_org() {
+    let client = Client::new_emulator();
+    let alice = key_pair_from_string("Alice");
+    let register_org_message = random_register_org_message();
+
+    let tx_applied = submit_ok(&client, &alice, register_org_message.clone()).await;
```
geigerzaehler: We’re neglecting to check that `tx_applied.result` is `Ok(())`. Same for `unregister_org`.

core/src/state.rs
NunoAlexandre: The `id` is a property of an Org, so I don't get your point. Are you concerned with storage optimization? If that's the case, that's not something we _need_ to worry about now. In the worse case, we could have dedicated types for storage that don't have the id to cut the redundancy. 

In any case, from a practical point of view, the org id will be required for a couple of operations such as unregister an org and any operation on projects, so I see useful keeping this field in Org to facilitate those use cases, without having to keep an org_id somewhere aside.

We can discuss this further on Monday!

runtime/src/registry.rs
NunoAlexandre: Readability. 

`if org.members == vec![sender] && org.projects.is_empty()` means what it means, `if can_be_unregistered(org, sender)` is more descriptive. We can move it out and unit test it, that'd better.

test-utils/src/lib.rs
NunoAlexandre: Good idea!

runtime/tests/org_registration.rs
NunoAlexandre: Hm, no, I'm checking that for all applied transactions.

cli/src/commands.rs
NunoAlexandre: Will do!

core/src/state.rs
NunoAlexandre: Agree, will remove.

core/src/state.rs
NunoAlexandre: Done!

core/src/message.rs
NunoAlexandre: done.

test-utils/src/lib.rs
NunoAlexandre: Done.

runtime/src/registry.rs
NunoAlexandre: ✅

runtime/src/registry.rs
NunoAlexandre: The project implementation will not vanish, this is indeed duplication we don't want to have. We can easily dry this, but I'd prefer us to think of a way of making this more reliable. It is rather fragile to manipulate such a critical part of the registry using hardcoded values and methods to extract ids from storage keys. For instance, `key_hash_prefix_length = 16;`. Shall BlakeTwo128 change its output size, this code won't complain but keep compiling, silently.

My suggestion is to create a `BlakeTwo128Decoder` where this is dry and placed within context. It would be even better if the keys specified in the `decl_storage` could automatically pickup (or have to) on this implementation when they have `hasher(blake2_128_concat)` enabled.

Will see what's possible. Let if you have any further or different insight.

core/src/state.rs
NunoAlexandre: ✅

runtime/src/registry.rs
NunoAlexandre: One point more: functions are not just useful to extract duplications but also to create layers of abstract. `if org.members == vec![sender] && org.projects.is_empty()` is what I call "naked code", `if can_be_unregistered(org, sender)` is abstracted and provides a richer idiom to think about the problem.

runtime/tests/org_registration.rs
geigerzaehler: Ah yes. I missed it.

runtime/src/registry.rs
geigerzaehler: Good point. I’d also be ok with leaving this for now and creating an issue.

runtime/tests/org_registration.rs
NunoAlexandre: No worries, I prefer false-negative than let letting bad stuff go in.

runtime/src/registry.rs
NunoAlexandre: Sounds good! Maybe we can work something out with the Substrate project to implement this out of the box.

core/src/state.rs
NunoAlexandre: Discussed offline: Storing data on the blockchain is expensive, for two main reasons:

1. data schema amendments are difficult since the data effectively lives forever
2. each full node needs to keep _all_ the data


**Solution**:

Create an internal `Org` type that is only used for storage, that does not contain the redundant `id` field. When fetching an org or multiple orgs, build the external `Org` type by extracting the id from the storage key. It changes nothing for the client and reduces costs in the chain itself.

Thanks, @geigerzaehler!

NunoAlexandre: @geigerzaehler this is done. I added a new `Org` type to `core` and adjusted everything accordingly.

core/src/lib.rs
```diff
@@ -56,4 +58,46 @@ pub type ProjectName = String32;
 
 pub type ProjectId = (ProjectName, ProjectDomain);
 
+pub type OrgId = String32;
+
+/// Org
+///
+/// A complete representation of an Org.
+/// Different from state::Org in which this type gathers
```
geigerzaehler: You can use [doc links](https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html)
```suggestion
/// Different from [state::Org] in which this type gathers
```

core/src/lib.rs
```diff
@@ -56,4 +58,46 @@ pub type ProjectName = String32;
 
 pub type ProjectId = (ProjectName, ProjectDomain);
 
+pub type OrgId = String32;
+
+/// Org
```
geigerzaehler: Not sure this should be in `core` because it’s only a client concern. But also ok with having it here.

core/src/lib.rs
```diff
@@ -56,4 +58,46 @@ pub type ProjectName = String32;
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Org {
+    // Id of an Org.
+    pub id: OrgId,
+
+    /// Account ID that holds the org funds.
+    ///
+    /// It is randomly generated and, unlike for other accounts,
+    /// there is no private key that controls this account.
+    pub account_id: AccountId,
```
geigerzaehler: Probably best to just link to the respective field on `state::Org` to prevent documentation drift.

core/src/lib.rs
NunoAlexandre: I thought about that as well. I decided to leave it in core since all other types are there as well, and because _this_ `Org` type is _the_ `Org` type that matters most in the sense that's the type to be used all around and that captures the whole of an org. The `state::Org` is just a useful type for storage optimization. Will leave it here for now.

core/src/lib.rs
NunoAlexandre: I agree. I left it like this expecting you to comment and suggest something better. Will leave a link in the format "See [state::Org::field]"

PR #215, opened 2020-02-10
Run end_to_end tests sequentially

### LINKED ISSUE #214: Can’t run end-to-end tests concurrently
With our current setup for end-to-end tests in is not possible to run tests concurrently. This does not show at the moment but prevents us from adding more tests.

### Symptoms

Trying to run tests concurrently may result in spurious `watch_extrinsic stream terminated` errors. These errors mask the fact that the transaction is invalid.

### Cause

The cause for this is the account nonce. We use the `ClientT::sign_and_submit_message` method to submit messages in tests. This method fetches the account nonce from the chain. Since there is only one account with funds on the test chain all tests use this account. If two tests run concurrently it might happen that they issue the request for the account nonce at the same time and get the same nonce back. This results in the two tests submitting a transaction with the same nonce. Only one of the transaction will be accepted and the other one will fail.

### Stop-gap solution

To work around this we can run the end-to-end tests with the `--threads=1` flag on CI and document this properly

### Long-term solution

In the long term, the best solution for this would probably be to create a mutex for accounts so that only one test submits a transaction at a time. We should also add more accounts to the test ledger so that the tests can use one of a variety of accounts.

NunoAlexandre:
Closes #214 

Use the [serial_test](https://crates.io/crates/serial_test) crate to mark specific tests to run sequentially instead of in parallel.

On the reasons for this PR, see https://github.com/radicle-dev/radicle-registry/issues/214.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3527,6 +3527,7 @@ dependencies = [
  "radicle-registry-test-utils",
  "rand 0.7.3",
  "sc-rpc-api",
+ "serial_test",
  "sp-core",
  "sp-io",
  "sp-rpc",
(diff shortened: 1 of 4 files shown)


NunoAlexandre:
@geigerzaehler can do! See https://github.com/radicle-dev/radicle-registry/issues/214#issuecomment-584149577.

NunoAlexandre:
@geigerzaehler ready for your review.

Issue #203, opened 2020-02-05
Adjust Project type


NunoAlexandre:
Currently, the `Project` type is:

``` rust
pub struct Project {
    pub id: ProjectId,
    pub account_id: AccountId,
    pub members: Vec<AccountId>,
    pub current_cp: CheckpointId,
    pub metadata: Bytes128,
}
```

Adjust it to reflect the definition from the latest spec:

<img width="501" alt="Screenshot 2020-02-05 at 11 06 08" src="https://user-images.githubusercontent.com/7075260/73831988-929daf80-4807-11ea-9e34-a1d531f20626.png">

NunoAlexandre:
Done ✅

Issue #202, opened 2020-02-04
Transaction: Add TransferFromOrganization


NunoAlexandre:
Supersedes `TransferFromProject`.

### Spec

![Screenshot 2020-02-04 at 14 18 24](https://user-images.githubusercontent.com/7075260/73748233-43994100-4759-11ea-83bd-a533c9c1f185.png)

NunoAlexandre:
Done ✅

Issue #201, opened 2020-02-04
Transaction: Drop TransferFromProject 


NunoAlexandre:
Superseded by #202. Once projects are within Orgs, transfers will happen at the org level instead of the project level.

NunoAlexandre:
Done ✅

Issue #199, opened 2020-02-04
Drop ProjectDomain


NunoAlexandre:
- Replace all usages of `ProjectDomain` the new corresponding Org ID 
- Drop `ProjectDomain`

NunoAlexandre:
Done ✅

PR #218, opened 2020-02-11
Feature // Project under Orgs

NunoAlexandre:
Closes 
- [Adjust Project type](https://github.com/radicle-dev/radicle-registry/issues/203)
- [Transaction: Register project within an Org](https://github.com/radicle-dev/radicle-registry/issues/196)
- [Drop ProjectDomain](https://github.com/radicle-dev/radicle-registry/issues/199)
- [Transaction: Drop TransferFromProject](https://github.com/radicle-dev/radicle-registry/issues/201)
- [Transaction: Add TransferFromOrganization](https://github.com/radicle-dev/radicle-registry/issues/202)

### Notes

- A `core::Project` type in addition to the adjusted `state::Project` type to separate the complete representation of a project and it's optimized storage representation. The same was previously done with `Org`.

### Discussion

As of now, `ProjectId` is the alias `type ProjectId = (OrgId, ProjectName)` as suggested in a previous discussion. I find this to be rather unfriendly in terms of type-safety and api. It's easy to mess up the tuple and have no guidance from the compiler and it's obscure to play with this type when operating on projects. 

Furthermore, we need to agree with the order in which OrgId and ProjectName are specified. When displaying this id, we go `{project_name}.{org_id}`, but when following the hierarchy, we have org_id before project_name. I am ok with having the "display" different from the rest, as long as the rest is always consistent. 

#### Alternative A

Make `ProjectId` a struct:

``` rust
struct ProjectId {
  org_id: OrgId,
  name: ProjectName
}
```

#### Alternative B

Rename `ProjectId` to `ProjectKey` and only use it in the registry storage. Everywhere else, pass `OrgId` and `ProjectName` as separate arguments. It flattens the access to these values and, like alternative A, improves readability (e.g, `project_id.0` vs `org_id`).

I prefer option B. The notion of "project id" is only useful or directly applicable when it comes to the storage. Elsewhere, it reads better "Register project with name N under org O" than "Register project with the id that's the tuple (org O, name N)".

None of these options solve the type-safety issue. I'd be fine considering fixing that part in a separate issue. At the moment, I am more concerned with how unidiomatic working with `ProjectId` reveals to be.


NunoAlexandre:
## Not ready yet

This PR is still a draft and not yet ready for review.

- [x] Ensure all spec details are covered
- [x] Ensure all relevant docs are updated and coherent with the changes
- [x] Update changelog

NunoAlexandre:
The PR is now ready for review.

core/src/lib.rs
```diff
@@ -91,4 +88,35 @@ impl Org {
...
+    pub metadata: Bytes128,
+}
+
+impl Project {
+    pub fn id(self) -> ProjectId {
+        (self.org_id, self.name)
+    }
+
+    /// Build a [crate::Project] given all its properties obtained from storage.
+    pub fn from(org_id: OrgId, name: ProjectName, project: state::Project) -> Self {
```
geigerzaehler: In Rust it is more idiomatic to call this `new`. `from` is an associated function of the `From` trait with a different signature.

core/src/message.rs
```diff
@@ -56,31 +56,31 @@ pub struct UnregisterOrg {
...
 ///
 /// # State-dependent validations
 ///
-/// A project with the same ID must not yet exist.
-///
 /// A checkpoint with the given ID must exist.
+///
+/// The involved org must exit.
+///
+/// A project with the same name must not yet exist in the same org.
```
geigerzaehler:
```suggestion
/// A project with the same name must not yet exist in the org.
```

client/tests/end_to_end.rs
```diff
@@ -29,24 +29,31 @@ async fn register_project() {
...
+    assert_eq!(
+        project.current_cp.clone(),
+        register_project_message.checkpoint_id
+    );
+    assert_eq!(project.metadata.clone(), register_project_message.metadata);
 
     assert_eq!(
         tx_applied.events[0],
-        RegistryEvent::ProjectRegistered(project_id.clone(), project.account_id).into()
+        RegistryEvent::ProjectRegistered(project_id.clone()).into()
```
geigerzaehler: I think it would also be great to check here that the project is a member of the org project list

core/src/state.rs
```diff
@@ -155,3 +142,19 @@ pub struct Org {
...
+impl Org {
+    /// Add the given project to the list of [Org::projects].
+    /// Return a new Org with the new project included or the
+    /// same org if the org already contains that project.
+    pub fn add_project(self, project_name: ProjectName) -> Org {
+        if self.projects.contains(&project_name) {
+            self
+        } else {
+            Self {
+                projects: [&self.projects[..], &[project_name]].concat(),
```
geigerzaehler: We can leverage interior mutability here since we are consuming `self`
```rust
self.projects.push(project_name)
```

runtime/src/registry.rs
```diff
@@ -165,33 +165,36 @@ decl_module! {
...
                 current_cp: message.checkpoint_id,
                 metadata: message.metadata
             };
 
-            store::Projects::insert(project_id.clone(), project);
+            store::Projects::insert(project_id.clone(), new_project);
+            store::Orgs::insert(
+                org_id.clone(),
+                store::Orgs::get(org_id).unwrap().add_project(project_name)
+            );
```
geigerzaehler: You can also use [`StorageMap::mutate`](https://substrate.dev/rustdocs/master/frame_support/storage/trait.StorageMap.html#tymethod.mutate)

runtime/src/registry.rs
```diff
@@ -165,33 +165,36 @@ decl_module! {
 
         #[weight = SimpleDispatchInfo::InsecureFreeNormal]
         pub fn register_project(origin, message: message::RegisterProject) -> DispatchResult {
-            let sender = ensure_signed(origin)?;
+            let _sender = ensure_signed(origin)?;
```
geigerzaehler: We’re not verifying that the sender is a member of the organization. I’m not sure whether this was clear from the spec though. We should also add tests for this.

core/src/message.rs
```diff
@@ -56,31 +56,31 @@ pub struct UnregisterOrg {
...
+///
```
geigerzaehler: Please also document that the author of the transaction must be a member of the organization.

test-utils/src/lib.rs
```diff
@@ -52,24 +49,38 @@ pub async fn create_project_with_checkpoint(
     .result
     .unwrap();
 
-    let message = random_register_project_message(checkpoint_id);
+    let register_project_message = random_register_project_message(checkpoint_id);
+    let register_org_message = message::RegisterOrg {
+        id: register_project_message.id.0.clone(),
+    };
```
geigerzaehler: See my comment about the end-to-end tests.

cli/src/commands.rs
```diff
@@ -263,8 +257,9 @@ impl CommandT for ShowGenesisHash {
 /// Transfer funds to recipient
 pub struct Transfer {
     #[structopt(parse(try_from_str = parse_account_id))]
-    /// Recipient Account in SS58 address format
```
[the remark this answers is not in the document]
NunoAlexandre: Yes. I always feel itchy having such detailed info in the docs that go so easily out of sync and that can be looked up, but since this is the cli I understand the trade-off.

cli/src/commands.rs
```diff
@@ -298,27 +293,30 @@ impl CommandT for Transfer {
 }
 
 #[derive(StructOpt, Debug, Clone)]
-/// Transfer funds from a project to a recipient. The author needs to be the owner of the project
-pub struct TransferProjectFunds {
-    /// Name of the project in the .rad domain
-    #[structopt(value_name = "project")]
-    project_name: ProjectName,
+/// Transfer funds from an org to a recipient.
+/// The author needs to be part of [crate::state::Org::members] of the org.
```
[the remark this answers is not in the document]
NunoAlexandre: Got it, will change.

client/tests/end_to_end.rs
```diff
@@ -29,24 +29,31 @@ async fn register_project() {
     .unwrap();
 
     let register_project_message = random_register_project_message(checkpoint_id);
-
+    let register_org_message = message::RegisterOrg {
+        id: register_project_message.id.0.clone(),
+    };
```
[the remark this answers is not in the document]
NunoAlexandre: Yeah, I agree. The thing with your example is that we are not building the `RegisterProject` like that, we would need to adjust `create_project_with_checkpoint` and `random_register_project_message` to receive an `OrgId` param. Will do, but I sure miss default arguments in Rust.

core/src/lib.rs
NunoAlexandre: ✅

runtime/src/registry.rs
NunoAlexandre: I tried that first but it didn't seem to work. When I switched to this approach the tests covering this bit passed. I will give it another try.

runtime/src/registry.rs
NunoAlexandre: Good catch! I remember thinking about this but then left it, confusing it with `register_org` where the sender can not be a member of the org since it doesn't yet exist.

core/src/message.rs
NunoAlexandre: ✅

runtime/src/registry.rs
NunoAlexandre: Fixed ✅

runtime/src/registry.rs
NunoAlexandre: After addressing https://github.com/radicle-dev/radicle-registry/pull/218#discussion_r378184406, it's no longer beneficial to use `mutate` since we already hold the org.

core/src/state.rs
NunoAlexandre: TIL 👌

client/tests/end_to_end.rs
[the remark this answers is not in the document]
NunoAlexandre: ✅

NunoAlexandre:
@geigerzaehler thank you for your thorough and fruitful review 👌 I have addressed everything. Can you give it another round?

PR #221, opened 2020-02-13
Dry id extraction from BlakeTwo128 storage keys #213

### LINKED ISSUE #213: Dry id extraction from BlakeTwo128 storage keys
We can easily dry this, but I'd prefer us to think of a way of making this more reliable. It is rather fragile to manipulate such a critical part of the registry using hardcoded values and methods to extract ids from storage keys. For instance, `key_hash_prefix_length = 16;`. Shall BlakeTwo128 change its output size, this code won't complain but keep compiling, silently.

My suggestion is to create a `BlakeTwo128Decoder` where this is dry and placed within context. It would be even better if the keys specified in the `decl_storage` could automatically pick up (or have to) on this implementation when they have `hasher(blake2_128_concat)` enabled.

Will see what's possible. Let me know if you have any further or different insight.

See [StorageKey](https://substrate.dev/rustdocs/master/sp_storage/struct.StorageKey.html) for possible useful insights.

NunoAlexandre:
Closes #213 

<blockquote>
We can easily dry this, but I'd prefer us to think of a way of making this more reliable. It is rather fragile to manipulate such a critical part of the registry using hardcoded values and methods to extract ids from storage keys. For instance, `key_hash_prefix_length = 16;`. Shall BlakeTwo128 change its output size, this code won't complain but keep compiling, silently.

My suggestion is to create a `BlakeTwo128Decoder` where this is dry and placed within context. It would be even better if the keys specified in the `decl_storage` could automatically pick up (or have to) on this implementation when they have `hasher(blake2_128_concat)` enabled.

Will see what's possible. Let me know if you have any further or different insight.

See [StorageKey](https://substrate.dev/rustdocs/master/sp_storage/struct.StorageKey.html) for possible useful insights.

_Originally posted by @NunoAlexandre in https://github.com/_render_node/MDI0OlB1bGxSZXF1ZXN0UmV2aWV3Q29tbWVudDM3NjkyMjczMQ==/comments/review_comment_
</blockquote>

### How

Add a `BlakeTwo128KeyDecoder` that provides the following decode function:

``` rust
pub fn decode<V: parity_scale_codec::Decode>(
    // The raw key to be decoded.
    key: &[u8],
    // The substrate storage prefix.
    prefix: [u8; 32],
) -> Result<V, parity_scale_codec::Error> {
```

Decode the key:

``` rust
let org_id = BlakeTwo128KeyDecoder::decode(&key, orgs_prefix)
let project_id = BlakeTwo128KeyDecoder::decode(&key, project_prefix)
```

This is the result of being pragmatic after trying the following described below.

### What I also tried

Regarding the quote above from the linked issue, my aim was to be able to do something like this:

``` rust
impl BlakeTwo128KeyDecoder for store::Projects {}

store::Projects::decode_key(&key);
```

Having such a set up would mean that we could get the `prefix` and the key type (as in StorageMap)` "for free" and thus be able to do the right thing without boilerplate code.

I tried doing so through the following (rust pseudo-code):

``` rust
pub trait BlakeTwo128KeyDecode
where
    Self: StorageMap<K, V>,
{
    fn decode(key: &[u8]) -> Result<V, parity_scale_codec::Error> {
         // Length of BlakeTwo128 output
         let key_hash_prefix_length = 16;
         let key_prefix_length = Self::prefix().len() + key_hash_prefix_length;
         let mut id_bytes = &key[key_prefix_length..];
         V::decode(&mut id_bytes)
    }
}
```

The thing is that `K` and `V` can't be implicitly inferred. Passing those types are parameters to `BlakeTwo128KeyDecode` introduced boilerplate and space for error, and it didn't really help because, in the end, `StorageMap` doesn't provide the `final_prefix()` function needed to make this worth it. I tried doing that using another substrate storage trait that did provide it but then something else was missing (maybe the Key type, I guess).

In conclusion, I think this aim is still something worth achieving but there should be a better way of doing so. Ideally, Substrate would provide such a decode function for storage using a hasher that allows the key to be decoded. I will look into that in my spare time.

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -59,64 +59,6 @@ pub mod store {
             pub Checkpoints: map hasher(blake2_256) CheckpointId => Option<state::Checkpoint>;
         }
     }
-
-    #[cfg(feature = "std")]
-    impl Projects {
-        /// Get the project ID from the projects storage key.
-        ///
-        /// The following property holds.
-        /// ```
-        /// # use radicle_registry_core::*;
-        /// # use radicle_registry_runtime::registry::store;
-        /// # use frame_support::storage::generator::StorageMap;
-        /// # use std::str::FromStr;
-        /// let project_id = (
-        ///     OrgId::from_str("Monadic").unwrap(),
-        ///     ProjectName::from_str("radicle").unwrap()
-        /// );
-        ///
-        /// let key = store::Projects::storage_map_final_key(project_id.clone());
-        /// let extracted_project_id = store::Projects::id_from_key(&key).unwrap();
-        /// assert_eq!(project_id, extracted_project_id)
-        /// ```
-        pub fn id_from_key(key: &[u8]) -> Result<ProjectId, parity_scale_codec::Error> {
-            use parity_scale_codec::Decode;
-
-            let project_prefix = Self::final_prefix();
-            // Length of BlakeTwo128 output
-            let key_hash_prefix_length = 16;
-            let key_prefix_length = project_prefix.len() + key_hash_prefix_length;
-            let mut project_id_bytes = &key[key_prefix_length..];
-            ProjectId::decode(&mut project_id_bytes)
-        }
-    }
-
-    #[cfg(feature = "std")]
-    impl Orgs {
-        /// Get the org ID from the orgs storage key.
-        ///
-        /// The following property holds.
-        /// ```
-        /// # use radicle_registry_core::*;
-        /// # use radicle_registry_runtime::registry::store;
-        /// # use frame_support::storage::generator::StorageMap;
-        /// # use std::str::FromStr;
-        /// let org_id = OrgId::from_str("org").unwrap();
-        /// let key = store::Orgs::storage_map_final_key(org_id.clone());
-        /// let extracted_org_id = store::Orgs::id_from_key(&key).unwrap();
-        /// assert_eq!(org_id, extracted_org_id)
-        /// ```
-        pub fn id_from_key(key: &[u8]) -> Result<OrgId, parity_scale_codec::Error> {
-            use parity_scale_codec::Decode;
-
-            let prefix = Self::final_prefix();
-            // Length of BlakeTwo128 output
-            let key_hash_prefix_length = 16;
-            let key_prefix_length = prefix.len() + key_hash_prefix_length;
-            let mut id_bytes = &key[key_prefix_length..];
-            OrgId::decode(&mut id_bytes)
-        }
-    }
 }
 
 pub use store::Store;
@@ -345,3 +287,78 @@ decl_event!(
         CheckpointSet(ProjectId, CheckpointId),
     }
 );
+
+/// DecodeKey trait
+///
+/// A DecodeKey type must implement the decode_key function and
+/// it is free to choose what decoding algorithm it prefers.
+///
+/// DecodedKey::Key is the expected decoded key type
+///
+pub trait DecodeKey {
+    type Key: parity_scale_codec::Decode;
+
+    /// Decode the given raw storage map `key`. It's the inverse of
+    /// [frame_support::generator::StorageMap::storage_map_final_key],
+    /// so applying `decode_key` after `storage_map_final_key` must
+    /// yield identity as to the original input key.
+    fn decode_key(key: &[u8]) -> Result<Self::Key, parity_scale_codec::Error>;
+}
+
+impl DecodeKey for store::Orgs {
+    type Key = OrgId;
+
+    fn decode_key(key: &[u8]) -> Result<OrgId, parity_scale_codec::Error> {
+        decode_blake_two128_concat_key(key)
+    }
+}
+
+impl DecodeKey for store::Projects {
+    type Key = ProjectId;
+
+    fn decode_key(key: &[u8]) -> Result<ProjectId, parity_scale_codec::Error> {
+        decode_blake_two128_concat_key(key)
+    }
+}
+
+/// Decode a blake_two128_concat hashed key to the inferred type K.
+///
+/// The key consists of the concatenation of the module prefix hash (16 bytes),
+/// the storage prefix hash (16 bytes), the key hash (16 bytes), and
+/// finally the raw key. See the actual implementation of this key concatenation at
+/// [frame_support::storage::generator::StorageMap::storage_map_final_key].
+pub fn decode_blake_two128_concat_key<K: parity_scale_codec::Decode>(
+    key: &[u8],
+) -> Result<K, parity_scale_codec::Error> {
+    let final_storage_key_prefix_length = 48;
+    let mut id_bytes = &key[final_storage_key_prefix_length..];
+    K::decode(&mut id_bytes)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use frame_support::storage::generator::StorageMap;
+
+    #[test]
+    /// Test that store::Orgs::decode_key after store::Orgs::storage_map_final_key
+    /// is identify as to the original input id.
+    fn orgs_decode_key_identity() {
+        let org_id = OrgId::from_string("Monadic".into()).unwrap();
+        let hashed_key = store::Orgs::storage_map_final_key(org_id.clone());
+        let decoded_key = store::Orgs::decode_key(&hashed_key).unwrap();
+        assert_eq!(decoded_key, org_id);
+    }
+
+    #[test]
+    /// Test that store::Projects::decode_key after store::Projects::storage_map_final_key
+    /// is identify as to the original input id.
+    fn projects_decode_key_identity() {
+        let org_id = OrgId::from_string("Monadic".into()).unwrap();
+        let project_name = ProjectName::from_string("Radicle".into()).unwrap();
+        let project_id: ProjectId = (org_id, project_name);
+        let hashed_key = store::Projects::storage_map_final_key(project_id.clone());
+        let decoded_key = store::Projects::decode_key(&hashed_key).unwrap();
+        assert_eq!(decoded_key, project_id);
+    }
+}
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -37,7 +37,7 @@ use parity_scale_codec::{Decode, FullCodec};
 
 use frame_support::storage::generator::{StorageMap, StorageValue};
 use frame_support::storage::StoragePrefixedMap;
-use radicle_registry_runtime::{balances, registry, Runtime};
+use radicle_registry_runtime::{balances, registry, registry::DecodeKey, Runtime};
 
 mod backend;
 mod error;
@@ -231,7 +231,7 @@ impl ClientT for Client {
         let keys = self.backend.fetch_keys(&orgs_prefix, None).await?;
         let mut org_ids: Vec<OrgId> = Vec::with_capacity(keys.len());
         for key in keys {
-            let org_id = registry::store::Orgs::id_from_key(&key)
+            let org_id = registry::store::Orgs::decode_key(&key)
                 .expect("Invalid runtime state key. Cannot extract org ID");
             org_ids.push(org_id)
         }
@@ -249,7 +249,7 @@ impl ClientT for Client {
         let keys = self.backend.fetch_keys(&project_prefix, None).await?;
         let mut project_ids = Vec::with_capacity(keys.len());
         for key in keys {
-            let project_id = registry::store::Projects::id_from_key(&key)
+            let project_id = registry::store::Projects::decode_key(&key)
                 .expect("Invalid runtime state key. Cannot extract project ID");
             project_ids.push(project_id);
         }


runtime/src/blake2_128_key_decoder.rs
```diff
@@ -0,0 +1,37 @@
...
+        // The substrate storage prefix.
+        prefix: [u8; 32],
+    ) -> Result<V, parity_scale_codec::Error> {
+        // Length of BlakeTwo128 output
+        let key_hash_prefix_length = 16;
+        let key_prefix_length = prefix.len() + key_hash_prefix_length;
+        let mut id_bytes = &key[key_prefix_length..];
+        V::decode(&mut id_bytes)
+    }
+}
```
NunoAlexandre: @geigerzaehler I decided to have this in a separate module to avoid increasing the amount of stuff in `registry` and `lib`, which in my experience is making it somewhat difficult to go through.

runtime/src/blake2_128_key_decoder.rs
```diff
@@ -0,0 +1,37 @@
...
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
CodeSandwich: Is this header necessary in every file we create? Why is it here?

runtime/src/registry.rs
```diff
@@ -38,6 +38,20 @@ pub trait Trait:
     type Event: From<Event> + Into<<Self as frame_system::Trait>::Event>;
 }
 
+pub trait BlakeTwo128KeyDecoder {}
+impl dyn BlakeTwo128KeyDecoder {
```
CodeSandwich: What's that? I've never seen that syntax before!

runtime/src/blake2_128_key_decoder.rs
NunoAlexandre: It's a fair question. The header is the GNU License disclaimer and I see it in every file so I have it here too. I have the idea of hearing this was necessary during team week but my memory isn't clear on that. @geigerzaehler?

runtime/src/registry.rs
NunoAlexandre: I just learned it. The normal way is to have a trait `T` and then have some concrete type `X` implementing `T`. That way, `T` can be seen as `X` and it provides the interface established by `T`, if any. In this case, we don't need to have multiple types implementing `BlakeTwo128KeyDecoder`, we just want to have this object `BlakeTwo128KeyDecoder` that provides this decode function. So having it declared as an empty trait and then having it implementing itself, creating what's called in Rust a "trait object". To make that explicit, Rust requires the `dyn` keyword.

Note that this could also be done using a struct instead:

``` rust
pub struct BlakeTwo128KeyDecoder {}

impl BlakeTwo128KeyDecoder {
    pub fn decode<V: parity_scale_codec::Decode>(
        // The raw key to be decoded.
        key: &[u8],
        // The substrate storage prefix.
        prefix: [u8; 32],
    ) -> Result<V, parity_scale_codec::Error> {
        // Length of BlakeTwo128 output
        let key_hash_prefix_length = 16;
        let key_prefix_length = prefix.len() + key_hash_prefix_length;
        let mut id_bytes = &key[key_prefix_length..];
        V::decode(&mut id_bytes)
    }
}
```

As far as I see, they are equivalent solutions. I ended up with the trait alternative given where I was coming from with this solution (see the PR description) and because it seems more appropriate to think of this as a trait object than as an empty struct.

CodeSandwich: That's a great find, now trait objects are first class citizens with their own `impl`s and stuff :O I was even able to write `impl Trait1 for dyn Trait2` and it works!

On the code review side I'm not sure if this trait is necessary at all. It only provides a single static method (without `self`) and in the end it's nothing more than a regular function.

NunoAlexandre: Ahah nice!

> On the code review side I'm not sure if this trait is necessary at all. It only provides a single static method (without self) and in the end, it's nothing more than a regular function.

That's true. I like to have it as is for namespacing but it's worth considering the trade-off. @geigerzaehler how do you feel about it?

NunoAlexandre: The namespacing can be achieved using the module name itself if imported in a qualified fashion, so let me know what you guys find better as readers.

runtime/src/blake2_128_key_decoder.rs
geigerzaehler: We decided on licensing the code in this repo as GPLv3. This requires us to add the header to every file. You can find the header in `LICENSE_HEADER` and we check that it is there on every CI build.

runtime/src/registry.rs
CodeSandwich: Now it's full path is `radicle_registry_runtime::blake2_128_key_decoder::BlakeTwo128KeyDecoder::decode(key, prefix)`, plain function's would be `radicle_registry_runtime::blake2_128_key_decoder::decode(key, prefix)`. I don't think that function is less discoverable. Unless one overimports, it's reasonable to expect in code  `blake2_128_key_decoder::decode(key, prefix)`.

NunoAlexandre:
@geigerzaehler 

> I’m not sure about this approach. The main issue I see here is that the decision what key extraction logic to use is made at the call site, i.e. in the client. 

That's exactly what this PR does by letting and having the call site choosing to use `BlakeTwo128KeyDecoder`.


<blockquote>
Ideally I would like to keep the API:

``` rust
let org_id = registry::store::Org::id_from_key(&key)
```

Me too, as explained in the PR description regarding the different approach I tried.

> This way we can swap out the implementation of how the keys are formed without the client knowing.

This is a valid and good point. It sounds different from your first point though, but I might be getting it wrong.

> The simplest way to solve this would be to extract the logic into a simple function and use that for the implementation of id_from_key.

There is a difference between _simple_ and _easy_. I don't see this alternative you present being simpler. Furthermore, it misses the storage prefix, which would also need to be passed to `extract_blake_two128_concat_key`.

> Obviously this is not super dry and structured but I believe that this is good enough and does not involve creating super complicated machinery.

I honestly miss seeing the advantage in this alternative over what this PR provides at the moment, apart from your point regarding being able to swap the extraction without modifying the call site. I will take on this point and rethink the solution.

Thanks!

geigerzaehler:
> > This way we can swap out the implementation of how the keys are formed without the client knowing.
> 
> This is a valid and good point. It sounds different from your first point though, but I might be getting it wrong.

My point here was meant to clarify why letting the call-site choose the extractor is probably not ideal. To elaborate, assume that we change the hasher in the runtime but forget to update it in the client then the code is broken. Moreover, if we, at some point, have multiple places that extract a project ID from a key then we need to update it in all the places. We should try to keep code that needs to change in lock-step close together.

I think another symptom of this is that the code is not tested at the moment.

> Furthermore, it misses the storage prefix, which would also need to be passed to `extract_blake_two128_concat_key`.

We only need the storage prefix for its lenght. Fortunately, the storage prefix always does not depend on the storage we use so we can hard code that to `32`, too.

I

runtime/src/registry.rs
geigerzaehler: I’d also prefer using functions and modules instead of structs or traits for namespacing. I think we should only use structs when we actually construct and consume them as values and use traits when we have structs implementing them. Otherwise people might stumble across the code and get confused what the purpose of the struct or trait is.

NunoAlexandre:
@geigerzaehler 

> My point here was meant to clarify why letting the call-site choose the extractor is probably not ideal.

Clear now 👍 

> To elaborate, assume that we change the hasher in the runtime but forget to update it in the client then the code is broken.

Ah, that's a major point! It should be covered by the tests as you say, but I missed that I was calling `BlakeTwo128KeyDecoder::decode` from the client. 

> I think another symptom of this is that the code is not tested at the moment.

**TODO** 👍 

> We only need the storage prefix for its lenght. Fortunately, the storage prefix always does not depend on the storage we use so we can hard code that to 32, too.

Ok! Do you have docs referenced for that btw?

geigerzaehler:
> Ok! Do you have docs referenced for that btw?

Unfortunately not. I just looked at the code that implements the key generation logic. In fact, that code doesn’t event use `final_prefix`. So even using this does not guarantee that we’re doing the correct thing. This is why we should have unit tests.

runtime/src/registry.rs
```diff
@@ -333,3 +289,36 @@ decl_event!(
...
+        decode_blake_two128_concat_key(key)
+    }
+}
+
+/// Decode a blake_two128_concat hashed key to the inferred type K.
+pub fn decode_blake_two128_concat_key<K: parity_scale_codec::Decode>(
+    key: &[u8],
+) -> Result<K, parity_scale_codec::Error> {
+    let storage_key_prefix_length = 16;
+    let storage_key_hash_length = 32;
```
geigerzaehler: Could you add some documentation where these values come from?

runtime/src/registry.rs
```diff
@@ -333,3 +289,36 @@ decl_event!(
...
     }
 );
+
+/// DecodeKey trait
+///
+/// A DecodeKey type must implement the decode_key function and
+/// it is free to choose what decoding algorithm it prefers.
+///
+/// K is the expected decoded key type
+pub trait DecodeKey<K: parity_scale_codec::Decode> {
```
geigerzaehler: I’d suggest an associated type parameter instead of a type parameter. This makes it impossible to have multiple implementations for different `K` for the same type.

runtime/src/registry.rs
```diff
@@ -38,6 +38,20 @@ pub trait Trait:
...
+        prefix: [u8; 32],
+    ) -> Result<V, parity_scale_codec::Error> {
+        // Length of BlakeTwo128 output
+        let key_hash_prefix_length = 16;
+        let key_prefix_length = prefix.len() + key_hash_prefix_length;
+        let mut id_bytes = &key[key_prefix_length..];
+        V::decode(&mut id_bytes)
+    }
+}
+
```
geigerzaehler: I guess we can remove this now

runtime/src/registry.rs
```diff
@@ -333,3 +289,36 @@ decl_event!(
         CheckpointSet(ProjectId, CheckpointId),
     }
 );
+
+/// DecodeKey trait
+///
+/// A DecodeKey type must implement the decode_key function and
+/// it is free to choose what decoding algorithm it prefers.
```
geigerzaehler: Can we expand the docs a bit? We should mention that this extracts the map keys from the raw storage keys and that it must be the inverse of `frame_support::generator::StorageMap::storage_map_final_key`. A usage example would be great, too.

NunoAlexandre:
@geigerzaehler @CodeSandwich could you give it a second round? I have redesigned the solution mostly considering Thomas' comment on the need for separation between client - registry regarding the decoding method to use.

NunoAlexandre:
@geigerzaehler, regarding

> Could you add some tests?

This is already tested when testing project and org registration as we fetch all the keys and compare the id against the original one. As far as I see, that suffices. Let me know if you see it otherwise.

runtime/src/registry.rs
NunoAlexandre: Weird that it didn't report dead code.

runtime/src/registry.rs
NunoAlexandre: The constraint is indeed true, but it's not a bad thing. Why would we want to implement `DecodeKey` for, say, `store::Projects` for different key types if `store::Projects` can only have one?

runtime/src/registry.rs
NunoAlexandre: Do you happen to have them with you already? If not, I will dig to find them.

geigerzaehler:
> This is already tested when testing project and org registration as we fetch all the keys and compare the id against the original one. As far as I see, that suffices. Let me know if you see it otherwise.

That’s true. For these kind of well defined pieces I like to have unit tests nevertheless. Doc tests are fine, too.

runtime/src/registry.rs
geigerzaehler: > Why would we want to implement `DecodeKey` for, say, `store::Projects` for different key types if `store::Projects` can only have one?

Indeed, we would not want this. The associated types would help us enforce this and also serve as a hint to a reader about the relation ship between the key type and the trait.

runtime/src/registry.rs
NunoAlexandre: Sounds good, agree 👍

runtime/src/registry.rs
NunoAlexandre: Discussed offline, I was misunderstanding the comment. I have already changed it to have a `Key` associated type 👌

runtime/src/registry.rs
NunoAlexandre: Done 👍

runtime/src/registry.rs
NunoAlexandre: Done 👍

NunoAlexandre:
@geigerzaehler ready or one more round :)

runtime/src/registry.rs
```diff
@@ -345,3 +287,82 @@ decl_event!(
...
+    let final_storage_key_length = storage_key_prefix + hashed_key_length;
+    let mut id_bytes = &key[final_storage_key_length..];
+    K::decode(&mut id_bytes)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    /// This specific import is required over [frame_support::storage::StorageMap] to build.
+    use crate::sp_api_hidden_includes_construct_runtime::hidden_include::storage::generator::StorageMap;
```
geigerzaehler: This is what the compiler suggests. But it is only an alias for the following
```suggestion
    use frame_support::storage::generator::StorageMap;
```

NunoAlexandre: It didn't _build_ with the `frame_support` but I will give it another spin!

NunoAlexandre: For the record, it needs the "generator" part (as suggested). I was using `frame_support::storage::StorageMap`. Solved.

PR #223, opened 2020-02-13
PoW consensus

CodeSandwich:
This adds a dummy PoW consensus, miners just sleep for random time and call it a day.

There is a local copy of pow-consensus crate from Substrate. It has fixes in 2 separate commits, which at some point can be cherry-picked and merged in upstream Substrate.


geigerzaehler:
After talking with parity we realized that it should be possible to make it work without forking `sc_consensus_pow`. So no need to review this just yet

PR #224, opened 2020-02-14
Flatten ProjectId usage

NunoAlexandre:
Doing so improves readability, makes the api more idiomatic and less
obscure, and separates concerns (storage/state and transactions).

### Changes

There are 3 changes here:

- Flatten `ProjectId` usages, so to have two params (`ProjectName` and `OrgId`) instead of one `ProjectId`
- ~~Move`ProjectId` to `state.rs` since it's not only used as the storage key for Projects and checkpoints.~~ (See https://github.com/radicle-dev/radicle-registry/pull/224#issuecomment-587020442)
- Swap `ProjectName` and `OrgId` in `ProjectId`, so to be `(ProjectName, OrgId)`, being thus consistent with all the usages of these two values everywhere in the registry.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -96,14 +96,14 @@ impl CommandT for ShowProject {
     async fn run(&self, command_context: &CommandContext) -> Result<(), CommandError> {
         let opt_project = command_context
             .client
-            .get_project((self.org_id.clone(), self.project_name.clone()))
+            .get_project(self.project_name.clone(), self.org_id.clone())
             .await?;
 
         let project = match opt_project {
             None => {
                 return Err(CommandError::ProjectNotFound {
-                    org_id: self.org_id.clone(),
                     project_name: self.project_name.clone(),
+                    org_id: self.org_id.clone(),
                 });
             }
             Some(project) => project,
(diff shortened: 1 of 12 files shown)


NunoAlexandre:
@geigerzaehler I hear you. `ProjectId` is, herein, only used within the `state` context, and it is indeed stored in the ledger, as a map key. So it makes sense to me to have it in `state`, to be clear about its scope. Does this change your view on it?

geigerzaehler:
> @geigerzaehler I hear you. `ProjectId` is, herein, only used within the `state` context, and it is indeed stored in the ledger, as a map key. So it makes sense to me to have it in `state`, to be clear about its scope.

I’m ok with that. But then we should be consistent and move all key types into `state` and document the guildeline in the readme.

NunoAlexandre:
@geigerzaehler I agree with the consistency part, but if we are to move the key types you mentioned, then other types should be moved as well, such as `ProjectName`, since it's also stored. I'd prefer to have this discussion in a separate thread. It doesn't add much value and we have bigger fish to fish right now. I thus reverted moving ProjectId to state to have the PR limited to the flattening of the ProjectId into separate params.

PR #225, opened 2020-02-14
Remove duplicate type definition in runtime

geigerzaehler:
Remove a type definition in the runtime that is already provided by the
`core` crate.

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -105,7 +105,10 @@ pub trait ClientT {
     ) -> Result<Response<TransactionApplied<Message_>, Error>, Error>;
 
     /// Fetch the nonce for the given account from the chain state
-    async fn account_nonce(&self, account_id: &AccountId) -> Result<state::Index, Error>;
+    async fn account_nonce(
+        &self,
+        account_id: &AccountId,
+    ) -> Result<state::AccountTransactionIndex, Error>;
 
     /// Return the gensis hash of the chain we are communicating with.
     fn genesis_hash(&self) -> Hash;
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -208,7 +208,10 @@ impl ClientT for Client {
         self.backend.get_genesis_hash()
     }
 
-    async fn account_nonce(&self, account_id: &AccountId) -> Result<state::Index, Error> {
+    async fn account_nonce(
+        &self,
+        account_id: &AccountId,
+    ) -> Result<state::AccountTransactionIndex, Error> {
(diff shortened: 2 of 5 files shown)


geigerzaehler:
> Where is this type redefined in `core`? I've been able to find only `Index`, which is not `AccountIndex`.

`core` indeed fines it as `Index`. `AccountIndex` from the `runtime` served the same purpose but was just not used. How do you feel about renaming `Index` to `AccountIndex` in `core`.

CodeSandwich:
> `core` indeed fines it as `Index`. `AccountIndex` from the `runtime` served the same purpose but was just not used. How do you feel about renaming `Index` to `AccountIndex` in `core`.

Are they equivalent? `runtime::AccountIndex` is described as
```
The type for looking up accounts.
```

while `core::Index` is described as
```
Next index (nonce) for a transaction of an account.
```

geigerzaehler:
`runtime::AccountIndex` is a remnant of the substrate template. In fact they used this type for something different from account transaction indices, namely for short account addresses. So using `AccountIndex` might be confusing. I’d suggest using the term `AccountTransactionIndex` to be clear. In any case `runtime::AccountIndex` is not used.

CodeSandwich:
>  I’d suggest using the term `AccountTransactionIndex` to be clear.

It would be way clearer this way, this name is obvious in terms of meaning. I would only probably shorten it, 23 characters is almost 1/4 of the line! How about `AccTxIdx`?

geigerzaehler:
> It would be way clearer this way, this name is obvious in terms of meaning. I would only probably shorten it, 23 characters is almost 1/4 of the line! How about `AccTxIdx`?

I think readability is a bigger concern than writability (especially with autocompletion and autoformatting) and I would like to keep the full name. It is also not widely used

geigerzaehler:
Maybe we could shorten it to `TransactionIndex` which is still descriptive enough in my oppinion.

CodeSandwich:
Everybody has code completion, so writing is never an issue. The 1/4 line argument is for readability, rustfmt may be forced to devastate a lot of concise one-line code and spread it across multiple lines with baroque indentation. All 3 parts of this name have widely used abbreviations, so they shouldn't be hard to understand even while skimming.

I like `AccountTransactionIndex` much more than `TransactionIndex`, it highlights that it's about transaction in context of one specific account.

geigerzaehler:
Could you have another look @CodeSandwich. I opted for writing out the name since it is more consistent with the existing style.

CodeSandwich:
Looks good :+1: The convention is an important thing.

PR #227, opened 2020-02-17
Switch to dummy PoW

CodeSandwich:
There are changes in runtime, node and in client. The node's `service` is made easier to be migrated to its newer upstream versions.


node/src/dummy_pow.rs
```diff
@@ -0,0 +1,56 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use radicle_registry_runtime::opaque::Block;
+use radicle_registry_runtime::Hash;
+use sc_consensus_pow::{Error, PowAlgorithm};
+use sp_consensus_pow::Seal;
+use sp_runtime::generic::BlockId;
+
+#[derive(Clone)]
+pub struct DummyPow;
```
geigerzaehler: Could you add some documentation for this struct and the module?

runtime/src/lib.rs
```diff
@@ -321,31 +301,15 @@ impl_runtime_apis! {
...
-        fn authorities() -> Vec<AuraId> {
-            Aura::authorities()
-        }
-    }
-
     impl sp_session::SessionKeys<Block> for Runtime {
-        fn generate_session_keys(seed: Option<Vec<u8>>) -> Vec<u8> {
-            opaque::SessionKeys::generate(seed)
+        fn generate_session_keys(_seed: Option<Vec<u8>>) -> Vec<u8> {
+            Default::default()
```
geigerzaehler: Pleaes document why this is a noop.

CodeSandwich: I myself am not sure, I don't understand yet what session keys are used for, I hope to get to that point soon. The old code was using Aura and Grandpa for it and Kulupu has these noops. Removing this code completely derails compilation.

Can you help me with that? Do you understand why session keys are required, but dummies work fine?

geigerzaehler: You can add this comment
```rust
// An implementation for the `SessionKeys` runtime API is required by the types
// of [sc_service::ServiceBuilder]. However, the implementation is otherwise unused
// and has no effect on the behavior of the runtime. Hence we implement a dummy
// version.
``` 
Maybe we could also move it to another `impl_runtime_apis!` invocation that is guarded by `#[cfg(feature = "std")]`.

CodeSandwich:
I've extracted macros to separate file (the diff IS cleaner!) and pushed `import_setup` out of macro (it allowed full and light queue initialization unification).

I was trying to wrap these parts in functions instead of macros, but I've given up. No matter if I was trying to replicate type provided by the `with_import_queue`or type required by `PowBlockImport::new`, I was ending up with dozens of entangled types pulled from all over Substrate. This solution seems very brittle. `proc_macro`'s duck typing is a huge relief in this context.

runtime/src/lib.rs
CodeSandwich: Thank you!

The `cfg` flag just won't work. The `impl_runtime_apis!` macro can't be called twice, because it generates duplicate code, that clashes with itself. The macro doesn't survive feature gating of its elements either, neither with `#[cfg...]` nor with `#![cfg...]`.

geigerzaehler: To bad.

PR #230, opened 2020-02-17
Fix CI failing due to bintray being full

CodeSandwich:
The `upload_to_bintray` fix doesn't work, due to `set -euo pipefail` it fails anyway

### DIFF
--- a/ci/run
+++ b/ci/run
@@ -115,7 +115,9 @@ function upload_to_bintray () {
     -H "X-Bintray-Override: 1" \
     -H "X-Checksum-Sha2: $file_checksum" \
     "$api_url/content/$subject/$repo/$remote_file" \
-    --data-binary @$local_file
+    --data-binary @$local_file \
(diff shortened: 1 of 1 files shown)


ci/run
```diff
@@ -149,5 +152,5 @@ if [[ -n "${BINTRAY_API_KEY:-}" ]]; then
   echo "--- Upload bintray artifacts"
   # We make this optional until the issue with the storage limit is
```
NunoAlexandre: Note that now this comment doesn't quite make much sense to the context of the line below.

geigerzaehler:
Judging from [a recent master build](https://buildkite.com/monadic/radicle-registry/builds/1360#8da65cbd-8123-4f0e-a512-130fd4e36743) my solution is working. Could you point me to a failed build?

CodeSandwich:
You're right, it's working fine. I must have made a mistake and looked at the wrong build, sorry about the noise.

PR #233, opened 2020-02-18
Document dummy pow average block time

geigerzaehler:
Extract the parameters that control the DummyPoW behavior and document how to calculate the average block time from these parameters.

### DIFF
--- a/node/src/dummy_pow.rs
+++ b/node/src/dummy_pow.rs
@@ -26,8 +26,18 @@ use sp_runtime::generic::BlockId;
 /// a random length sleep and successful returning of a 0-byte nonce.
 ///
 /// It accepts all seals. Verification is always successful.
+///
+/// The mining behavior is controlled by [MINE_DURATION] and [MINE_SUCCESS_PROBABILITY]. The
+/// average block time is
+///
+///     (MINE_DURATION / n) * (1 / MINE_SUCCESS_PROBABILITY)
+///
+/// where `n` is the number of miners.
 #[derive(Clone)]
 pub struct DummyPow;
+
+const MINE_DURATION: std::time::Duration = std::time::Duration::from_millis(10);
+const MINE_SUCCESS_PROBABILITY: f32 = 0.005;
 
 impl PowAlgorithm<Block> for DummyPow {
     type Difficulty = u128;
@@ -53,8 +63,8 @@ impl PowAlgorithm<Block> for DummyPow {
         _difficulty: Self::Difficulty,
         _round: u32,
     ) -> Result<Option<Seal>, Error<Block>> {
-        std::thread::sleep(std::time::Duration::from_millis(10));
-        if rand::random::<f32>() < 0.005 {
+        std::thread::sleep(MINE_DURATION);
+        if rand::random::<f32>() < MINE_SUCCESS_PROBABILITY {
             Ok(Some(vec![]))
         } else {
             Ok(None)


CodeSandwich:
The formula doesn't add up for me. Every `MINE_DURATION` everybody rolls a dice and there's `1 - ((1 - MINE_SUCCESS_PROBABILITY) ^ n)` probability that somebody will succeed.

geigerzaehler:
> The formula doesn't add up for me. Every `MINE_DURATION` everybody rolls a dice and there's `1 - ((1 - MINE_SUCCESS_PROBABILITY) ^ n)` probability that somebody will succeed.

Mining success follows a [geometric distribution](https://en.wikipedia.org/wiki/Geometric_distribution) since we are interested in the probability of the number of attempts until success. The mean of the geometric distribution is `1/MINE_SUCCESS_PROBABILITY` and there are `n / MINE_DURATION` attempts per second. This leads me to the formula.

The formula is also consistent with the observations on the devnet after factoring in block propagation

CodeSandwich:
According to `n / ( MINE_DURATION * MINE_SUCCESS_PROBABILITY )` the more nodes are mining the longer it takes to mine a block, linearly.

geigerzaehler:
> According to `n / ( MINE_DURATION * MINE_SUCCESS_PROBABILITY )` the more nodes are mining the longer it takes to mine a block, linearly.

That does not make sense indeed! It should be `MINE_DURATION / ( n * MINE_SUCCESS_PROBABILITY )`. Fixing it

node/src/dummy_pow.rs
```diff
@@ -26,9 +26,19 @@ use sp_runtime::generic::BlockId;
...
+/// The mining behavior is controlled by [MINE_DURATION] and [MINE_SUCCESS_PROBABILITY]. The
+/// average block time is
+///
+///     n / ( MINE_DURATION * MINE_SUCCESS_PROBABILITY )
+///
+/// where `n` is the number of miners.
 #[derive(Clone)]
 pub struct DummyPow;
 
+const MINE_DURATION: std::time::Duration = std::time::Duration::from_millis(10);
```
NunoAlexandre:
```suggestion
const MINE_DURATION_MILLIS: std::time::Duration = std::time::Duration::from_millis(10);
```

geigerzaehler: `MINE_DURATION` is a `Duration` that is not tied to a certain unit. For example `MINE_DURATION = std::time::Duration::from_seconds(1)` would also be ok. `MINE_DURATION_MILLIS` would make sense if it was a plain `u32`.

NunoAlexandre: I got that, but now this value represents milliseconds, so the name could show that instead of forcing a look up.

CodeSandwich: I think it's ok the way it is. `MINE_DURATION` is a `Duration`, not a `MillisecondContainer`. The fact that we used a millisecond constructor is a detail, which unravels when you want to find out the actual value.

The docs embed value in the name of a variable, we don't.

geigerzaehler: > Edit: That's how the Rust docs do so: https://doc.rust-lang.org/std/thread/fn.sleep.html

They are calling it `ten_millis` because the value of the variable is equal to 10ms. The name has no relation to how it is defined. It could be defined as `Duration::new(0, 10*1000*1000)`

> I got that, but now this value represents milliseconds, so the name could show that instead of forcing a look up.

For users of the variable the unit that it was defined with does not play a role. That is why we use `Duration` instead of a value. If they want to get the number of milliseconds they do `MINE_DURATION.as_millis()`. If they want to get the number of seconds they do `MINE_DURATION.as_secs_f32()`, etc.

If we were to append `_MILLIS` to the name and we decide that the value should be two seconds we sould either end up with `MINE_DURATION_MILLIS = Duration::new(2, 0)` or we would need to change the name. Neither is a good option. Without using the suffix we don’t have the problem.

NunoAlexandre: More often than not, the languages I used didn't offer type-safety regarding durations, and not having that info as part of the naming was a real pain. My take on this might just be a residue of that pain.  I understand that in this case, we do have that type-safety and adding the suffix is not of such value.

You are both happy with how it is now, and I am too since we have type-safety, so we are good.

node/src/dummy_pow.rs
```diff
@@ -26,9 +26,19 @@ use sp_runtime::generic::BlockId;
 /// a random length sleep and successful returning of a 0-byte nonce.
 ///
 /// It accepts all seals. Verification is always successful.
+///
+/// The mining behavior is controlled by [MINE_DURATION] and [MINE_SUCCESS_PROBABILITY]. The
+/// average block time is
+///
+///     n / ( MINE_DURATION * MINE_SUCCESS_PROBABILITY )
```
CodeSandwich: This formula doesn't make sense

geigerzaehler: Forgot to push :man_facepalming:

Issue #231, opened 2020-02-18
Add hashing PoW


CodeSandwich:
Right now we have a dummy PoW, we need a real PoW based on actual work with hashing.

We want to keep both versions interchangable (at least for now). Dummy should be used for integration tests and when `--dev` flag is used on the binary. Otherwise the real hashing should be used.

The hashing should be based on Ethereum with still warm Blake3 algorithm.

The real difficulty adjustment is not a goal yet, it can be a simple function of block height.

NunoAlexandre:
@CodeSandwich you can close this one 🙌

PR #234, opened 2020-02-19
Add PoW algorithm CLI switch

CodeSandwich:
This is the first part of https://github.com/radicle-dev/radicle-registry/issues/231: add the PoW algorithm switch mechanism triggerable by the `--dev` CLI flag. Now it provides only 1 option, but adding another one becames trivial.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -18,10 +18,12 @@
 //! Available chain specs
 //! * [dev] for runnning a single node locally and develop against it.
 //! * [local_devnet] for runnning a cluster of three nodes locally.
+use crate::pow_alg_config::PowAlgConfig;
 use radicle_registry_runtime::{
     AccountId, BalancesConfig, GenesisConfig, SudoConfig, SystemConfig, WASM_BINARY,
 };
 use sp_core::{Pair, Public};
+use std::convert::TryFrom;
 
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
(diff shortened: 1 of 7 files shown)


node/src/service_customization.rs
```diff
@@ -15,9 +15,41 @@
 
 //! This module contains chunks of code needed for customizations in [crate::service] module.
 
+use radicle_registry_runtime::GenesisConfig;
+use sc_service::config::Configuration;
+
+#[derive(serde::Deserialize, serde::Serialize)]
+pub enum PowAlgConfig {
```
geigerzaehler: Could you please add some documentation?

node/src/chain_spec.rs
```diff
@@ -149,3 +153,11 @@ pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Pu
...
         .public()
 }
+
+fn new_properties(pow_alg: PowAlgConfig) -> sc_service::Properties {
+    let mut map = sc_service::Properties::new();
+    let pow_alg_value =
+        serde_json::to_value(pow_alg).expect("failed to serialize PowAlgConfig for ChainSpec");
+    map.insert(PowAlgConfig::JSON_KEY.to_string(), pow_alg_value);
+    map
+}
```
geigerzaehler: We’re splitting the code that extracts `PowAlgConfig` from `Properties` and the code that creates `Properties` from `PowAlgConfig` . I’d prefer to have it in the same place. This would also allow us to keep `PowAlgConfig::JSON_KEY` private.

PR #237, opened 2020-02-25
Add hashing PoW

CodeSandwich:
This is another step of https://github.com/radicle-dev/radicle-registry/issues/231. It adds real PoW that hashes on non-`--dev` runs. The last element it's lacking is difficulty adjustment.


node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
+// Radicle Registry
```
NunoAlexandre: It would be great to bring some structure to our executables. I am not aware whether it's common to practise to use directories to structure modules in rust, but I'd see beneficial to create a `node > pow` dir:

``` python
- node
  - pow
    - blake3.rs
    - dummy.rs
    - config.rs
```

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+pub struct Blake3Pow;
+
+impl PowAlgorithm<Block> for Blake3Pow {
+    type Difficulty = Difficulty;
+
+    fn difficulty(&self, _parent: &BlockId<Block>) -> Result<Self::Difficulty, Error<Block>> {
+        Ok(Difficulty::from(10_000_000))
+    }
+
+    fn verify(
```
NunoAlexandre: Could you add docstrings?

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+/// This is a PoW algorithm implementation based on Blake3 hashing.
+///
+/// In order to check if data hash passes difficulty test, it must be interpreted as
+/// a big-endian 256-bit number. If it's smaller than or equal to value of threshold, it passes.
+/// The threshold is calculated from difficulty as `U256::max_value / difficulty`.
+///
+/// The difficulty is an average number of hashes that need to be checked in order mine a block.
+/// It must be capped at `2^192 - 1`. This limit guarantees that Substrate can calculate a sum
+/// of difficulties of blocks in a chain of up to `2^64` height and not overflow. This cap
+/// should be impossible to break with current technology due to constraints of the universe.
```
NunoAlexandre: I'd prefer to have this doc down at `fn difficulty` but leave that up to you and Thomas.

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+        difficulty: Self::Difficulty,
+    ) -> Result<bool, Error<Block>> {
+        let mut payload = pre_hash.as_bytes().to_vec();
+        payload.extend_from_slice(seal);
+        let hash = blake3::hash(&payload);
+        let threshold = difficulty_to_threshold(difficulty)?;
+        Ok(hash_passes_threshold(hash, threshold))
+    }
+
+    fn mine(
```
NunoAlexandre: docstrings

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+            if hash_passes_threshold(hash, threshold) {
+                return Ok(Some(nonce.to_vec()));
+            }
+            payload.truncate(Hash::len_bytes());
+        }
+        Ok(None)
+    }
+}
+
+fn difficulty_to_threshold(difficulty: Difficulty) -> Result<Threshold, Error<Block>> {
```
NunoAlexandre: Not sure what this means to convert a difficulty to a threshold. Could you add docs?

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+    if !(difficulty >> 192).is_zero() {
+        return Err(Error::BlockProposingError(
+            "difficulty too high".to_string(),
+        ));
+    }
+    Ok(Threshold::max_value() / difficulty)
+}
+
+fn hash_passes_threshold(hash: blake3::Hash, threshold: Threshold) -> bool {
```
NunoAlexandre: Same here. Not sure what it means for a hash to pass a threshold. The language can also be improved. Maybe `hash_surpases_threshold` or `hash_exceeds_threshold`?

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+    let hash_value = Threshold::from_big_endian(hash.as_bytes());
+    hash_value <= threshold
+}
+
+fn nonces_for_next_round() -> impl Iterator<Item = [u8; 8]> {
```
NunoAlexandre: It'd be also nice to have docs for this.

node/src/chain_spec.rs
```diff
@@ -102,7 +102,7 @@ pub fn devnet() -> ChainSpec {
         None, // telemetry endpoints
         // protocol_id
         Some("devnet"),
-        Some(sc_service::Properties::try_from(PowAlgConfig::Dummy).unwrap()),
+        Some(sc_service::Properties::try_from(PowAlgConfig::Blake3).unwrap()),
```
NunoAlexandre: I'd prefer this to be done the way around: `pow::Blake3::Config`. As I see it, it's the pow algorithms that have a config, not the config that holds an algorithm. I need to see the rest to get a better picture of this though.

After seeing `PowAlgConfig`, my understanding is that this type actual stands to list the algorithms we support, it's not really a _config_. Then, depending on the specific algorithm, we can obtain the associated properties, but that's something separate.

I have to think more about this.

node/src/service_customization.rs
```diff
@@ -79,6 +86,12 @@ macro_rules! node_import_queue {
                 $inherent_data_providers,
                 crate::dummy_pow::DummyPow
             ),
+            PowAlgConfig::Blake3 => node_import_queue_for_pow_alg!(
+                $client,
+                $select_chain,
+                $inherent_data_providers,
+                crate::blake3_pow::Blake3Pow
```
NunoAlexandre: It looks like that, for each algorithm, we will have the same block.

I suggest that, in `PowAlgConfig`, we provide a function that returns the corresponding implementation. In rust pseudo-code:

``` rust
pub fn implementation(self) -> PowAlgorithm {
   match self {
    Dummy => crate::pow::Dummy,
    Blake3 => crate::pow::Blake3
  }
}
```

NunoAlexandre: Then, here, we can just do:

``` rust
let algorithm = PowAlgConfig::try_from($config)?;
node_import_queue_for_pow_alg!(
    $client,
    $select_chain,
    $inherent_data_providers,
    algorithm.implementation()
),
```

node/src/service_customization.rs
```diff
@@ -42,6 +42,13 @@ macro_rules! new_full {
                 inherent_data_providers,
                 crate::dummy_pow::DummyPow
             ),
+            PowAlgConfig::Blake3 => start_mine!(
+                block_import,
+                service,
+                proposer,
+                inherent_data_providers,
+                crate::blake3_pow::Blake3Pow
```
NunoAlexandre: Same thing here, having the `implementation()` function we could also dry this.

node/src/service_customization.rs
```diff
@@ -79,6 +86,12 @@ macro_rules! node_import_queue {
                 $inherent_data_providers,
                 crate::dummy_pow::DummyPow
             ),
+            PowAlgConfig::Blake3 => node_import_queue_for_pow_alg!(
+                $client,
+                $select_chain,
+                $inherent_data_providers,
+                crate::blake3_pow::Blake3Pow
+            ),
```
geigerzaehler: Is it possible to simplify the code and reduce the duplication? I know that matching and generics don’t play well but maybe we can do something about it.

node/src/blake3_pow.rs
CodeSandwich: The `difficulty_to_threshold` function is private. It's not a part of an API unlike the silent contract with the user that the difficulty must be capped.

node/src/blake3_pow.rs
CodeSandwich: Same as above

node/src/blake3_pow.rs
CodeSandwich: This method comes from a trait and the trait provides docs for these methods. You can see it when you run `cargo doc --workspace --open`. I've never seen trait methods documented on the implementation side.

node/src/blake3_pow.rs
geigerzaehler: I think we should document functions no matter whether there public or private in terms of visibility. The documentation is for code contributors and as such we need to deal with any functions.

node/src/blake3_pow.rs
CodeSandwich: For me it seems straight forward if one knows what difficulty and threshold are. Reading struct doc should be enough as well. This is private function, so it's not necessarily important to be able to use it without any context.

node/src/blake3_pow.rs
CodeSandwich: Same as above

node/src/blake3_pow.rs
CodeSandwich: Same as above

node/src/service_customization.rs
CodeSandwich: The problem here is that if we define `pub fn implementation(self) -> impl PowAlgorithm`, the compiler must decide to pick one specific type, so it won't compile.

node/src/service_customization.rs
CodeSandwich: Same as above

node/src/blake3_pow.rs
NunoAlexandre: Right, which is not the case for me, for instance. I like the approach to have code being readable to as many people as possible, which is much better than the way around.

node/src/blake3_pow.rs
NunoAlexandre: Exactly inline to what Thomas said.

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use radicle_registry_runtime::opaque::Block;
+use radicle_registry_runtime::Hash;
+use sc_consensus_pow::{Error, PowAlgorithm};
+use sp_consensus_pow::Seal;
```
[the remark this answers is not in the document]
CodeSandwich: No, they're `sc_` and `sp_`

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+/// a big-endian 256-bit number. If it's smaller than or equal to value of threshold, it passes.
+/// The threshold is calculated from difficulty as `U256::max_value / difficulty`.
+///
+/// The difficulty is an average number of hashes that need to be checked in order mine a block.
+/// It must be capped at `2^192 - 1`. This limit guarantees that Substrate can calculate a sum
+/// of difficulties of blocks in a chain of up to `2^64` height and not overflow. This cap
+/// should be impossible to break with current technology due to constraints of the universe.
+///
+/// There's no difficulty adjustment algorithm yet.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
CodeSandwich: Why do we need this? `Copy` is very strict and hard to future-proof.

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+fn hash_passes_threshold(hash: blake3::Hash, threshold: Threshold) -> bool {
+    let hash_value = Threshold::from_big_endian(hash.as_bytes());
+    hash_value <= threshold
+}
+
+fn nonces_for_next_round() -> impl Iterator<Item = [u8; 8]> {
+    const COUNT: usize = 10_000_000;
+    static NEXT_FROM: AtomicU64 = AtomicU64::new(0);
```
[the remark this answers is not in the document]
CodeSandwich: We don't put it in `Blake3Pow` because it's going to be cloned unknown number of times by Substrate and we don't know how many and which copies are going to be run, maybe even in parallel. Now we are 100% certain that the whole node will never be mining the same block with the same nonce.

Randomization is a great point.

node/src/blake3_pow.rs
NunoAlexandre: That's fair!

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+use sc_consensus_pow::{Error, PowAlgorithm};
+use sp_consensus_pow::Seal;
+use sp_core::uint::U256;
+use sp_runtime::generic::BlockId;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+type Difficulty = U256;
+type Threshold = U256;
+
+/// This is a PoW algorithm implementation based on Blake3 hashing.
```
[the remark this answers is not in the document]
CodeSandwich: I'd rather keep it on the structure. This is the only visible part of the module API and it's having some characteristics important for the user. The module may contain dozens of different things. The corresponding functions probably shouldn't be mentioned, they're private in current API and by definition unstable. But the other way around seems fine, mention public struct docs from private function docs.

node/src/service_customization.rs
CodeSandwich: I doubt it. This is a generic function and it must be resolved to a specific type during compilation. The only thing we could do is wrap `DummyPow` and `Blake3Pow` in an enum implementing `PowAlgorithm`. That seems way overengineered, we won't have many more algorithms, will we?

node/src/service_customization.rs
NunoAlexandre: I gave this a try. To complete your comment, the issue is that the compiler needs the associated type `Difficulty` to be resolved, so it can't do it. Shame.

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+use radicle_registry_runtime::opaque::Block;
+use radicle_registry_runtime::Hash;
+use sc_consensus_pow::{Error, PowAlgorithm};
+use sp_consensus_pow::Seal;
+use sp_core::uint::U256;
+use sp_runtime::generic::BlockId;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+type Difficulty = U256;
+type Threshold = U256;
```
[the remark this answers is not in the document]
CodeSandwich: It's documented on the only public piece of the API: `Blake3Pow`. We can either copy and paste it or remove from the public API and spread it. I prefer having it all in one place, because it paints the whole image without stitching together contextless docs from helper functions.

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        Ok(None)
+    }
+}
+
+fn difficulty_to_threshold(difficulty: Difficulty) -> Result<Threshold, Error<Block>> {
+    if !(difficulty >> 192).is_zero() {
+        return Err(Error::BlockProposingError(
+            "difficulty too high".to_string(),
+        ));
```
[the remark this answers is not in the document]
CodeSandwich: I'm not sure how to make consensus ensure that. Do you have any tips?

node/src/blake3_pow.rs
CodeSandwich: The context is the key here. Any contributor must first read the `Blake3Pow` docs to understand what's going on. When one knows what's difficulty and what's threshold, it's obvious what `difficulty_to_threshold` helper function does.

node/src/service_customization.rs
NunoAlexandre: Another solution is to have Difficulty for Dummy being U256 (same as Blake3) but shall we introduce another algorithm that needs another type of difficulty and it won't work.

node/src/blake3_pow.rs
CodeSandwich: Did you read the `Blake3Pow` docs before reading its internal functions? If you did and you don't understand what's difficulty and threshold, then we must fix that. I think that we can expect understanding what structure does from anybody who wants to modify it.

NunoAlexandre: We can also thrive to _create_ understanding so that people will get engaged to want to contribute. If I read you well, you don't want what you see as noisy documentation because you expect a certain level of knowledge which makes that unnecessary. I get that. Reading the `Blake3Pow` did help, indeed. You might be right that it ain't needed. We are both biased here though.

node/src/blake3_pow.rs
NunoAlexandre: the language point is still applicable, unless if that's really the common language. "To pass" is not nearly as clear as "to exceed".

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+use sp_core::uint::U256;
+use sp_runtime::generic::BlockId;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+type Difficulty = U256;
+type Threshold = U256;
+
+/// This is a PoW algorithm implementation based on Blake3 hashing.
+///
+/// In order to check if data hash passes difficulty test, it must be interpreted as
```
NunoAlexandre:
```suggestion
/// In order to check if the data hash passes the difficulty test, it must be interpreted as
```

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+/// a big-endian 256-bit number. If it's smaller than or equal to value of threshold, it passes.
```
NunoAlexandre:
```suggestion
/// a big-endian 256-bit number. If it's smaller than or equal to the value of the threshold, it passes.
```

or 

```suggestion
/// a big-endian 256-bit number. If it's smaller than or equal to the threshold, it passes.
```

node/src/blake3_pow.rs
```diff
@@ -0,0 +1,107 @@
...
+/// The threshold is calculated from difficulty as `U256::max_value / difficulty`.
```
NunoAlexandre:
```suggestion
/// The threshold is obtained from `U256::max_value / difficulty`.
```

node/src/blake3_pow.rs
NunoAlexandre: @CodeSandwich you wrote this, we are reading it. This will be read many more times than touched, so I recommend we listen to the readers here, especially in this case that there is someone who is a noob and someone who is familiar to this.

Having some redudant docs won't hurt, not having it will raise some level of confusion, as witnessed here.

node/src/blake3_pow.rs
CodeSandwich: I wanted to avoid suggestion of the math beneath. How about `hash_passes_threshold_test`?

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,122 @@
...
+    Ok(Threshold::max_value() / difficulty)
+}
+
+fn hash_passes_threshold(hash: blake3::Hash, threshold: Threshold) -> bool {
+    let hash_value = Threshold::from_big_endian(hash.as_bytes());
+    hash_value <= threshold
+}
+
+fn nonces_for_next_round() -> impl Iterator<Item = [u8; 8]> {
+    const COUNT: usize = 10_000_000;
```
NunoAlexandre: You use this same value above. It is an idea to create a shared const value used in both places?

CodeSandwich: It's a coincidence, they have no direct relation. These numbers worked nicely when running node while testing.

NunoAlexandre: 👍

node/src/blake3_pow.rs
CodeSandwich: The noise is not a problem, docs are easy to skip. My main concerns are maintainability of non-DRY documentation and discoverability. If we spread documentation, it will be hard to put the whole image together. The user will know that threshold is calculated from difficulty, but won't have a clue why and what for, it will require a lot more effort. That's off-putting and invites stupid bugs. If we copy and paste it, we will have to maintain all these copies.

If this module was huge and required knowledge wide, it could make sense to decentralize, but here it's tiny, the whole docs are less 9 sentences.

node/src/service_customization.rs
geigerzaehler: > The only thing we could do is wrap `DummyPow` and `Blake3Pow` in an enum implementing `PowAlgorithm`. That seems way overengineered, we won't have many more algorithms, will we?

I agree. This couples the two implementation and we move the matching to the instance definitions which is not a good idea. So let’s keep it like it is

node/src/blake3_pow.rs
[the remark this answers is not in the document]
geigerzaehler: For now I think we can just remove this. In the future this should be handled by the `difficulty` method of the PowAlgorithm trait. Maybe we can just add a comment that the value may not be too large so we don’t accidentally change it.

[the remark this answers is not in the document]
geigerzaehler: For now I think we can just remove this. In the future this should be handled by the `difficulty` method of the PowAlgorithm trait. Maybe we can just add a comment that the value may not be too large so we don’t accidentally change it.

node/src/blake3_pow.rs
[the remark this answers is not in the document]
geigerzaehler: > We don't put it in `Blake3Pow` because it's going to be cloned unknown number of times by Substrate and we don't know how many and which copies are going to be run, maybe even in parallel.

Say we write
```rust
#[derive(Clone)]
struct Blake3Pow {
  current_nonce: Arc<AtomicU64>
}
```
And in the `mine` implementation we pass the field to `nonces_for_next_round`.

Then substrate can clone the struct multiple times and they will use the same shared state even if we mine on the clones in parallel.

node/src/blake3_pow.rs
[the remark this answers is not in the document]
geigerzaehler: My inclination is to optimize for what we have now and not care to much about future-proofing since we don’t know what the future brings. But you make a good point and I don’t care too much. (I also suspect that the APIs actually require an `Clone` implementation)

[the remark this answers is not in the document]
CodeSandwich: That's right, we don't work with Blake3Pow directly, we only pass it to Substrate and it only clones.

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,137 @@
...
+        self.payload.extend_from_slice(nonce);
+        let hash = blake3::hash(&self.payload);
+        let is_nonce_valid = hash_passes_threshold_test(hash, self.threshold);
+        self.payload.truncate(Hash::len_bytes());
+        is_nonce_valid
+    }
+}
+
+fn difficulty_to_threshold(difficulty: Difficulty) -> Threshold {
+    // Assumption, to be enforced later: only 192 lower bits are used
```
NunoAlexandre: What part is assumed to be at most 192 lower bits? The `max_value()`, the `difficulty`, the returned `Threshold`? That needs to be clarified and the comment could then be clearer as:

> It assumes that <part> is only 192 lower bits

or 

> it assumes that <part> only uses 192 lower bits

node/src/pow/config.rs
```diff
@@ -18,15 +18,16 @@ use std::convert::{TryFrom, TryInto};
 
 /// Configuration of PoW algorithm, can be stored as chain spec property
 #[derive(serde::Deserialize, serde::Serialize)]
-pub enum PowAlgConfig {
+pub enum Config {
```
NunoAlexandre: 👌

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,137 @@
...
+use std::sync::Arc;
+
+type Difficulty = U256;
+type Threshold = U256;
+
+const NONCES_PER_MINING_ROUND: usize = 10_000_000;
+
+/// The implementor of the Blake3 PoW algorithm.
+///
+/// For more information about this PoW algorithm see [module](index.html) documentation.
```
NunoAlexandre: I tried running `cargo doc` to see these as docs but for some reason, I don't see it as part of the targe/doc output.
That said, two things:

1. 
```suggestion
/// For more information about this PoW algorithm see the [module](index.html) documentation.
```

2. The `[module](index.html)` seems to point to nowhere? 🤔

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,137 @@
...
+use sp_runtime::generic::BlockId;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+
+type Difficulty = U256;
+type Threshold = U256;
+
+const NONCES_PER_MINING_ROUND: usize = 10_000_000;
+
+/// The implementor of the Blake3 PoW algorithm.
```
NunoAlexandre: nit: Is this _the_ implementation of the Blake3 PoW algorithm or _an_ implementation? if it's _an_, using _the_ might be confusing, as if we picked this up from some open-source gist or so. just an observation.

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,137 @@
...
+        .take(NONCES_PER_MINING_ROUND)
+        .map(u64::to_ne_bytes)
+    }
+}
+
+impl PowAlgorithm<Block> for Blake3Pow {
+    type Difficulty = Difficulty;
+
+    fn difficulty(&self, _parent: &BlockId<Block>) -> Result<Self::Difficulty, Error<Block>> {
+        // To be enforced: only 192 lower bits can be used
```
NunoAlexandre: Could you explain or guide me to where I can learn about what `X lower bits can be used` means?

On another note, do you have any ideas on how to _enforce_ that? I'd suggest you leave a `TODO(igor)` and/or create an issue here on GH.

node/src/pow/blake3_pow.rs
```diff
@@ -0,0 +1,137 @@
...
+/// For more information about this PoW algorithm see [module](index.html) documentation.
+#[derive(Clone)]
+pub struct Blake3Pow {
+    next_nonce: Arc<AtomicU64>,
+}
+
+impl Blake3Pow {
+    pub fn new() -> Self {
+        Blake3Pow {
+            next_nonce: Arc::new(AtomicU64::new(rand::random())),
```
NunoAlexandre: Just an observation: Having it this way is fair and convenient. However, `Blake3Pow::new` is effectful. I usually prefer to have both the convenience and control:

``` rust
/// Grants control, dependency injection, unit-testable, flexibility 
pub fn new(x: AtomicU64) -> Self {
  Blake3Pow { 
    nextOnce: Arc::new(x),
  }
}

/// Grants convenience of use
pub fn new_random() -> Self {
  Self::new(AtomicU64::new(rand::random()))
}
```

I miss default arguments in Rust, where we could have both in one:

``` rust
pub fn new(x: AtomicU64 = AtomicU64::new(rand::random())) -> Self {
  Blake3Pow { 
    nextOnce: Arc::new(x),
  }
}
```

NunoAlexandre:
@CodeSandwich there's also the CI failure:

> License header missing for node/src/pow/mod.rs. Rerun with --fix to add license header.

node/src/pow/blake3_pow.rs
CodeSandwich: You can run `cargo doc --open` to open it in the browser, you can then use the regular search to find this item.

The structure has address `struct.Blake3Pow.html` and the module has just `index.html`. I don't like it (assumption of HTML output), but that seems to be the right way to create internal links in rustdoc :man_shrugging:

node/src/pow/blake3_pow.rs
CodeSandwich: I have no idea how to enforce that and I don't care yet beyond the fact that the returned constant conforms. This method is not fully implemented, addition of difficulty adjusting is a different task and it has its own issue.

PR #238, opened 2020-02-25
Update substrate to latest master

geigerzaehler:
This update requires the following changes

* In the client we match all invalid transaction statuses with one pattern.
* The account nonce is now retrieved via the `frame_system::Account` storage.
* The `frame_system::Trait::Event` associated type has now a type bound that is generic over `Self`. This requires us to explicitly state some type bounds.


geigerzaehler:
The builds fail because we seem to generate invalid transactions for the end-to-end tests. This looks like quite the issue. We need to find out why the transaction is invalid. Unfortunately the RPC apis don’t provide enough error information I believe. We might need to have the node log information to determine the error. It’s most likely that the transactions are invalid because calling `<UncheckedExtrinsic as Checkable>::check()` fails. The place too look would be `client/src/transaction.rs`.

NunoAlexandre:
any luck, @CodeSandwich?

CodeSandwich:
We were able to find out that it's the node that is rejecting the transactions, it's sending `Invalid` message to the client. We've pinned down the exact commit in Substrate that introduces this behavior: https://github.com/paritytech/substrate/commit/0c8e05928d9d50e20df4388b9294fadc21058dce. Now I'm trying to find out what exactly in this commit is offensive to create a fix or workaround.

CodeSandwich:
The upgrade is finally fixed :tada: 

**DO NOT MERGE**
It depends on a commit in a feature branch, please wait for it getting merged and upgrade to newest master to avoid depending on a squashed-away commit: https://github.com/paritytech/substrate/pull/5227

CodeSandwich:
Ready to be merged!

PR #240, opened 2020-03-02
Feature/transaction fees

NunoAlexandre:
This PR is a draft, showing how we could implement our tx fee pipeline, where a `bid` field is being added to the `message::SetCheckpoint` and consumed in the corresponding `set_checkpoint` function in the `runtime::registry`.

For the sake of clarity, the idea is:

1. A user submits a tx with a `bid`
2. The registry charges the fees and uses the remainder as a tip
3. The withdrew amounts will later be transferred to some other account with a small burn

This draft PR addresses 1. and 2. for now, but only touching the `message::SetCheckpoint` and the `runtime > tests > checkpoint_setting.rs > set_checkpoint`.

You can test that specific test by running:

``` bash
cargo test -p radicle-registry-runtime set_checkpoint -- --exact
```

### Feedback request

Since [our requirements](https://radicle.community/t/transaction-fees-on-the-registry-requirements-flows-plan-and-questions/139) are such that we need control over who pays the fees, per transaction, and given the overall tx fee design we want to achieve, opting for a custom implementation over using the default substrate way is preferred.

Now, since going custom means more ways of doing it, I appreciate your early feedback.

I have left already some comments below to guide it. 

Some open question I have and want to have addresses in this draft:

- Where to place the `bid` field (See https://github.com/radicle-dev/radicle-registry/pull/240#discussion_r386267919)
- Do we charge the `base_fee` and the `tip` separately? See https://github.com/radicle-dev/radicle-registry/pull/240#discussion_r386279245 and https://github.com/radicle-dev/radicle-registry/pull/240#discussion_r386280733

### DIFF
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -122,6 +122,8 @@ pub struct SetCheckpoint {
     pub project_name: ProjectName,
     pub org_id: OrgId,
     pub new_checkpoint_id: CheckpointId,
+
+    pub bid: Balance
 }
 
 /// Transfer funds from an org account to an account.
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -255,7 +255,7 @@ decl_module! {
                     if !org.members.contains(&sender) {
                         return Err(RegistryError::InsufficientSenderPermissions.into())
                     }
-                    charge_fee(TransactionFee::Tip(123), &sender);
+                    charge_fee(TransactionFee::Tip(message.bid - TransactionFee::BaseFee.balance()), &sender);
 
                     state::Project {
                         current_cp: message.new_checkpoint_id,
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -20,7 +20,7 @@ use frame_support::{
     decl_event, decl_module, decl_storage,
     dispatch::DispatchResult,
     storage::StorageMap as _,
-    traits::{Currency, ExistenceRequirement, WithdrawReason, Randomness as _},
+    traits::{Currency, ExistenceRequirement, Randomness as _},
     weights::SimpleDispatchInfo,
 };
 use frame_system as system; // required for `decl_module!` to work
@@ -30,7 +30,7 @@ use sp_runtime::traits::Hash as _;
 
 use radicle_registry_core::*;
 
-use crate::{AccountId, Hash, Hashing};
+use crate::{AccountId, Hash, Hashing, transaction_fee::{charge_fee, TransactionFee}};
 
 pub trait Trait:
     frame_system::Trait<AccountId = AccountId, Origin = crate::Origin, Hash = Hash>
@@ -242,6 +242,7 @@ decl_module! {
             message: message::SetCheckpoint,
         ) -> DispatchResult {
             let sender = ensure_signed(origin)?;
+            charge_fee(TransactionFee::BaseFee, &sender);
 
             if store::Checkpoints::get(message.new_checkpoint_id).is_none() {
                 return Err(RegistryError::InexistentCheckpointId.into())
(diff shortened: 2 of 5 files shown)


core/src/message.rs
```diff
@@ -122,6 +122,8 @@ pub struct SetCheckpoint {
     pub project_name: ProjectName,
     pub org_id: OrgId,
     pub new_checkpoint_id: CheckpointId,
+
+    pub bid: Balance
```
NunoAlexandre: Added this field to `message::SetCheckpoint` was an initial example. 

This is a decision we need to make: where do we place the `bid` field? In every specific message or, for example, in the `TransactionExtra`? I decided to with the former alternative because 1) it grants transparency (It's obvious that a bid needs to be passed), 2) it grants flexibility (shall we want some tx to not be charged, we don't run into a leaked abstraction a la OOP inheritance, and 3) because we can access it easily at the respective tx function in the `registry.rs`.

runtime/src/transaction_fee.rs
```diff
@@ -0,0 +1,85 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{AccountId, Balance};
+
+use frame_support::traits::{Currency, ExistenceRequirement, WithdrawReason};
+
+
+/// This module models our Transaction Fee setup.
```
NunoAlexandre: Right now, this module is very drafty, no need to nitpick here. The idea is to isolate the handling of tx fees in this module, as well as provide useful error types and decisions on where the withdrew balances go.

runtime/src/registry.rs
```diff
@@ -242,6 +242,7 @@ decl_module! {
             message: message::SetCheckpoint,
         ) -> DispatchResult {
             let sender = ensure_signed(origin)?;
+            charge_fee(TransactionFee::BaseFee, &sender);
```
NunoAlexandre: I am not sure whether we want to charge the `BaseFee` aside from the `Tip`. This is a way of addressing my question at https://radicle.community/t/transaction-fees-on-the-registry-requirements-flows-plan-and-questions/139:

<blockquote>
Since org-related transactions should be withdrawn from the involved org’s account, we must authorize the tx author as a member of that org. Otherwise, it’d be an open door to destroy an org’s funds.

Doing so also means that there’s still an open door to issue transactions that will cause stress on the runtime (checking if the tx author belongs to the involved org). Note that a bad actor could - following the design we have at this point - issue as many transactions involving an org they do not belong to and pay nothing for it.

A solution could be to charge the base_fee as early as possible and only later the tip, when these authorizations steps can take place.

Let me have your thoughts!
</blockquote>

runtime/src/registry.rs
```diff
@@ -254,6 +255,8 @@ decl_module! {
                     if !org.members.contains(&sender) {
                         return Err(RegistryError::InsufficientSenderPermissions.into())
                     }
+                    charge_fee(TransactionFee::Tip(message.bid - TransactionFee::BaseFee.balance()), &sender);
```
NunoAlexandre: Following the [same comments above], then here - once we can authorize the tx author - we charge the tip. Here we are not charging the org (should we? We need to make a specific listing of who pays for each transaction, anyway).

runtime/src/registry.rs
NunoAlexandre: Another thing to do here is to ensure that `bid` is >= BaseFee.balance()`. I am on it, so no need to dedicate focus on that.

NunoAlexandre: an (implicit) TO DO here is to stop the execution of `charge_fee` fails, fyi.

runtime/src/registry.rs
NunoAlexandre: Answer: yes, the org should be the one charged at this point.

PR #241, opened 2020-03-02
Feature/transaction fees [Part 1]

NunoAlexandre:
## What

In #240, I show how we could implement our own custom fee payment setup. This PR is a first and clean iteration of the emerging ideas after playing around and suggesting the draft.

It introduces a sub crate `runtime::fees` which contains the following modules:

- `runtime::fees` - The basic types to support the notion of `Fee` 
- `runtime::fees::bid`- A module dedicated to modelling and constructing a `Bid`
- `runtime::fees::payment` - A module dedicated to the processing the payment of fees.

### Usage

Looking at the bare example at #240, these new items here introduced could be used as follows:

``` rust
// Create a `Bid` from a raw bid (Balance) value defined by the user in a core::message
let bid = fees::bid::Bid::new(message.bid)?;

// Pay the base_fee
fees::payment::pay_fee(bid.base_fee, &sender);

// Pay the tip
fees::payment::pay_fee(bid.tip, &sender);
```

### Decisions

1. `Bid`
I decided to create the `Bid` type to fully capture what it means. A bid, from the user perspective, is a total amount they indicate that is meant to cover all mandatory fees and the rest to be used as a tip. Therefore, I liked to have that captured in a type. Having this type also grants us the calculation breakdown is done in a single place, instead of having us decrementing value to the entry `bid: Balance` every time we want to pay a fee. This approach also makes this breakdown unit-testable.

2. `Fee` trait
Each fee has a `value` and a `withdraw_reason` associated. `BaseFee` has a fixed value, `Tip` has a calculated one in the moment `Bid` is created. `BaseFee` has `WithdrawReason::TransactionPayment ` as withdraw reason, `Tip` has `WithdrawReason::Tip`. By having this abstraction, we can easily add more `Fee` types without changing the function withdrawing them (`fees::payment::pay_fee`). It also allows us, for instance, to make `Bid` a fee itself, where it'd calculate all of its inner fees and where the reason would be `WithdrawReasons` (that type would need to be changed but the idea still stands): `WithdrawReason::Tip | WithdrawReason::TransactionPayment `.

3. The breakdown into small modules
I am having some difficulty digesting the long modules we have at the moment, that gather _a lot_ of different matters. For peace of mind, I separated the three concepts in 3 different modules. I am open to rethinking it, but I'd appreciate if we could consider moving towards this approach which is not meant to go to the other extreme.

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -47,7 +47,9 @@ pub type Hashing = BlakeTwo256;
 /// Each account has an associated [message::AccountBalance] and [message::Index].
 pub type AccountId = ed25519::Public;
 
-/// Balance of an account.
+/// The non-negative balance of anything storing the amount of currency.
+/// It can be used to represent the value of anything describing an amount,
+/// e.g. an account balance, the value of a fee, etc.
 pub type Balance = u128;
 
 /// The id of a project. Used as storage key.
--- a/runtime/src/fees/bid.rs
+++ b/runtime/src/fees/bid.rs
@@ -1,0 +1,65 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{Balance};
+use crate::fees::{BaseFee, Fee, Tip};
+use frame_support::dispatch::DispatchError;
+
+/// Bid
+///
+/// A Bid is an offer defined by a transaction author for the
+/// registry to process that transaction. The bid should cover
+/// all mandatory fees. The remainder left after deducting the
+/// mandatory fees is used as a tip, which will grant priority
+/// to the transaction in question accordingly.
+pub struct Bid {
+    base_fee: BaseFee,
+    tip: Tip
+}
+
+impl Bid {
+    /// Create a Bid with the given `bid`.
+    /// Fail when `bid` is insufficient to cover all the
+    /// mandatory fees, now being the `base_fee` alone.
+    pub fn new(bid: Balance) -> Result<Self, InvalidBid> {
+        let base_fee = BaseFee{};
+        let base_fee_value = base_fee.value();
+        if bid < base_fee_value {
+            return Err(InvalidBid::Insufficient)
+        }
+        Ok(
+            Self {
+                base_fee,
+                tip: Tip(bid - base_fee_value),
+            }
+        )
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum InvalidBid {
+    /// The bid is insufficient to cover all mandatory costs.
+    Insufficient,
+}
+
+impl From<InvalidBid> for DispatchError {
+    fn from(error: InvalidBid) -> Self {
+        DispatchError::Module {
+            index: 42,
+            error: error as u8,
+            message: None,
+        }
+    }
+}
--- a/runtime/src/fees/bid.rs
+++ b/runtime/src/fees/bid.rs
@@ -13,53 +13,63 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{Balance};
 use crate::fees::{BaseFee, Fee, Tip};
-use frame_support::dispatch::DispatchError;
+use crate::Balance;
 
 /// Bid
 ///
-/// A Bid is an offer defined by a transaction author for the
-/// registry to process that transaction. The bid should cover
+/// A Bid is an offer defined by transaction authors for the
+/// registry to process their transactions. The bid should cover
 /// all mandatory fees. The remainder left after deducting the
 /// mandatory fees is used as a tip, which will grant priority
-/// to the transaction in question accordingly.
+/// to the transaction in question accordingly to its value.
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Bid {
-    base_fee: BaseFee,
-    tip: Tip
+    pub base_fee: BaseFee,
+    pub tip: Tip,
 }
 
 impl Bid {
     /// Create a Bid with the given `bid`.
     /// Fail when `bid` is insufficient to cover all the
     /// mandatory fees, now being the `base_fee` alone.
-    pub fn new(bid: Balance) -> Result<Self, InvalidBid> {
-        let base_fee = BaseFee{};
-        let base_fee_value = base_fee.value();
-        if bid < base_fee_value {
-            return Err(InvalidBid::Insufficient)
-        }
-        Ok(
-            Self {
-                base_fee,
-                tip: Tip(bid - base_fee_value),
-            }
-        )
+    pub fn new(bid: Balance) -> Option<Self> {
+        let base_fee = BaseFee;
+        bid.checked_sub(base_fee.value()).map(|remainder| Self {
+            base_fee,
+            tip: Tip(remainder),
+        })
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-pub enum InvalidBid {
-    /// The bid is insufficient to cover all mandatory costs.
-    Insufficient,
-}
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rand::Rng;
 
-impl From<InvalidBid> for DispatchError {
-    fn from(error: InvalidBid) -> Self {
-        DispatchError::Module {
-            index: 42,
-            error: error as u8,
-            message: None,
+    #[test]
+    fn invalid_bid_insufficient() {
+        // The bid is insufficient to cover all mandatory fees.
+        assert_eq!(Bid::new(0), None);
+    }
+
+    #[test]
+    fn valid_bid_just_enough() {
+        // The bid is just enough to cover the mandatory base fee,
+        // leaving a tip of 0 left.
+        let bid = Bid::new(BaseFee.value()).unwrap();
+        assert_eq!(bid.base_fee.value(), 1);
+        assert_eq!(bid.tip.value(), 0);
+    }
+
+    #[test]
+    fn valid_bid_random() {
+        for _ in 0..50 {
+            // Generate a random bid between 1 and 9999.
+            let random_bid: Balance = rand::thread_rng().gen_range(1, 10000);
+            let bid = Bid::new(random_bid).unwrap();
+            assert_eq!(bid.base_fee.value(), 1);
+            assert_eq!(bid.tip.value(), random_bid - BaseFee.value())
         }
     }
 }
--- a/runtime/src/fees/mod.rs
+++ b/runtime/src/fees/mod.rs
@@ -47,3 +47,29 @@ impl Fee for Tip {
         WithdrawReason::Tip
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use rand::Rng;
+
+
+    #[test]
+    fn withdraw_reason() {
+        assert_eq!(BaseFee{}.withdraw_reason(), WithdrawReason::TransactionPayment);
+        assert_eq!(Tip(123).withdraw_reason(), WithdrawReason::Tip);
+    }
+
+    #[test]
+    fn base_fee_value() {
+        assert_eq!(BaseFee{}.value(), 1);
+    }
+
+    #[test]
+    fn tip_value() {
+        for _ in 0 .. 50 {
+            let random_tip: Balance = rand::thread_rng().gen();
+            assert_eq!(Tip(random_tip).value(), random_tip);
+        }
+    }
+}
--- a/runtime/src/fees/mod.rs
+++ b/runtime/src/fees/mod.rs
@@ -17,10 +17,11 @@ use crate::Balance;
 use frame_support::traits::WithdrawReason;
 
 mod bid;
+mod payment;
 
 pub trait Fee {
     fn value(&self) -> Balance;
-    fn withdraw_reason() -> WithdrawReason;
+    fn withdraw_reason(&self) -> WithdrawReason;
 }
 
 #[derive(Clone, Debug)]
@@ -30,7 +31,7 @@ impl Fee for BaseFee {
         1
     }
 
-    fn withdraw_reason() -> WithdrawReason {
+    fn withdraw_reason(&self) -> WithdrawReason {
         WithdrawReason::TransactionPayment
     }
 }
@@ -42,7 +43,7 @@ impl Fee for Tip {
         self.0
     }
 
-    fn withdraw_reason() -> WithdrawReason {
+    fn withdraw_reason(&self) -> WithdrawReason {
         WithdrawReason::Tip
     }
 }
--- a/runtime/src/fees/mod.rs
+++ b/runtime/src/fees/mod.rs
@@ -1,0 +1,48 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::Balance;
+use frame_support::traits::WithdrawReason;
+
+mod bid;
+
+pub trait Fee {
+    fn value(&self) -> Balance;
+    fn withdraw_reason() -> WithdrawReason;
+}
+
+#[derive(Clone, Debug)]
+pub struct BaseFee;
+impl Fee for BaseFee {
+    fn value(&self) -> Balance {
+        1
+    }
+
+    fn withdraw_reason() -> WithdrawReason {
+        WithdrawReason::TransactionPayment
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Tip(Balance);
+impl Fee for Tip {
+    fn value(&self) -> Balance {
+        self.0
+    }
+
+    fn withdraw_reason() -> WithdrawReason {
+        WithdrawReason::Tip
+    }
+}
--- a/runtime/src/fees/mod.rs
+++ b/runtime/src/fees/mod.rs
@@ -13,18 +13,28 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+//! Fee module
+//!
+//! This crate defines all things fees:
+//! * the types of fees supported by the registry
+//! * the [crate::fees::bid] module that abstracts the concept of bid.
+//! * the [crate::fees::payment] module where the withdrawing of fees takes place.
+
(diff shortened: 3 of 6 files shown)


runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Fee crate
+///
+/// This crate defines all things fees:
+/// * the types of fees supported by the registry
+/// * the [crate::fees::Bid] type that abstracts the concept of bid.
+/// * the [crate::fees::payment], where the withdrawing of fees takes place.
+pub mod bid;
+pub mod payment;
+
+/// Fee
```
CodeSandwich: I'm not sure if this trait requires docs, but in any case this doc doesn't seem very useful.

runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+pub trait Fee {
+    /// The associated [crate::Balance].
+    fn value(&self) -> Balance;
```
CodeSandwich: This signature smells like a confusion with naming. Does a fee have a value or does it have a balance?

runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    use super::*;
+    use rand::Rng;
+
+    #[test]
+    fn withdraw_reason() {
+        assert_eq!(
+            BaseFee {}.withdraw_reason(),
+            WithdrawReason::TransactionPayment
+        );
+        assert_eq!(Tip(123).withdraw_reason(), WithdrawReason::Tip);
```
CodeSandwich: This should be 2 tests. This test is designed to validate 2 different behaviors of 2 different structs.

runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[test]
+    fn base_fee_value() {
+        assert_eq!(BaseFee {}.value(), 1);
+    }
+
+    #[test]
+    fn tip_value() {
+        for _ in 0..50 {
+            let random_tip: Balance = rand::thread_rng().gen();
+            assert_eq!(Tip(random_tip).value(), random_tip);
```
CodeSandwich: This is a soak test of a simple getter, isn't it a bit paranoid?

runtime/src/fees/payment.rs
```diff
@@ -0,0 +1,37 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::fees::Fee;
+use crate::AccountId;
+use frame_support::traits::{Currency, ExistenceRequirement};
+
+/// Pay a given fee by withdrawing it from the `payee` account
+/// and transfering it, with a small burn, to <another account> (TODO(nuno)).
```
CodeSandwich: What does the `<xxx>` do? I'm really curious, the `another account` part is not visible in docs.

Should we put names in `TODO`s? Why? Maybe it would be better to describe *what should be done*, then it could be finished by anybody.

runtime/src/fees/payment.rs
```diff
@@ -0,0 +1,37 @@
...
+        fee.withdraw_reason().into(),
+        ExistenceRequirement::KeepAlive,
+    );
+    // 2. Transfer to ??? TODO(nuno)
+}
+
+/// The burn associated with the payment of a fee.
+/// When a fee is withdrew, it is then transfered to another account.
+/// We apply a small burn on that transfer to increase the value of
+/// our currency.
```
CodeSandwich: Should it be a `f64`? How do we apply it on a `u128` of coins?

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        Ok(Self {
+            base_fee,
+            tip: Tip(bid - base_fee_value),
+        })
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum InvalidBid {
```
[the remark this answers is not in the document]
NunoAlexandre: Now reading your comment I realise that I might have checkout a change explaining this type. The `InvalidBid` is mostly temporary to give shape to the problem at hand. I didn't want to make `new` return a `DispatchError` that would be breaking the abstraction layer where this sits. I find the simpler would be to make `new` return `Option<Self>` since there isn't much to the error case beside it being too low to cover the base fee. `new` will be called from the `registry`, where `Result` will be necessary. There I think we can convert the `option` to result where the error will be a new `RegistryError`, `InsufficientBid`.

What do you think?

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+    pub base_fee: BaseFee,
+    pub tip: Tip,
+}
+
+impl Bid {
+    /// Create a Bid with the given `bid`.
+    /// Fail when `bid` is insufficient to cover all the
+    /// mandatory fees, now being the `base_fee` alone.
+    pub fn new(bid: Balance) -> Result<Self, InvalidBid> {
+        let base_fee = BaseFee {};
```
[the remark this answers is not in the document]
NunoAlexandre: Ah, nice, thank you! I was firs trying `BaseFee()` but the compiler suggested to do `BaseFee {}`.

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum InvalidBid {
+    /// The bid is insufficient to cover all mandatory costs.
+    Insufficient,
+}
+
+impl From<InvalidBid> for DispatchError {
+    fn from(error: InvalidBid) -> Self {
+        DispatchError::Module {
+            index: 42,
```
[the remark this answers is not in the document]
NunoAlexandre: Same as explained above, this is a temporary code, I forgot to commit the comment explaining that. Thanks for being on the lookout anyway!

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl Bid {
+    /// Create a Bid with the given `bid`.
+    /// Fail when `bid` is insufficient to cover all the
+    /// mandatory fees, now being the `base_fee` alone.
+    pub fn new(bid: Balance) -> Result<Self, InvalidBid> {
+        let base_fee = BaseFee {};
+        let base_fee_value = base_fee.value();
+        if bid < base_fee_value {
```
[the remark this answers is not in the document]
NunoAlexandre: Sweet, thanks :)

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+        assert_eq!(bid.base_fee.value(), 1);
+        assert_eq!(bid.tip.value(), 0);
+    }
+
+    #[test]
+    fn valid_bid_random() {
+        for _ in 0..50 {
+            // Generate a random bid that between 1 and 101
+            let random_bid: Balance = rand::thread_rng().gen::<Balance>() * 100 + 1;
+            let bid = Bid::new(random_bid).unwrap();
```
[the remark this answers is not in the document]
NunoAlexandre: That is not a case we cover in this specific test, it is already tested aside. I'm fine adding but it makes the code more verbose unnecessarily, plus there's only one reason `Bid::new` would fail. See https://github.com/radicle-dev/radicle-registry/pull/241#discussion_r386997765.

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+        let bid = Bid::new(BaseFee {}.value()).unwrap();
+        assert_eq!(bid.base_fee.value(), 1);
+        assert_eq!(bid.tip.value(), 0);
+    }
+
+    #[test]
+    fn valid_bid_random() {
+        for _ in 0..50 {
+            // Generate a random bid that between 1 and 101
+            let random_bid: Balance = rand::thread_rng().gen::<Balance>() * 100 + 1;
```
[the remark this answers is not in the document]
NunoAlexandre: How's that?

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct Bid {
+    pub base_fee: BaseFee,
+    pub tip: Tip,
+}
+
+impl Bid {
+    /// Create a Bid with the given `bid`.
+    /// Fail when `bid` is insufficient to cover all the
+    /// mandatory fees, now being the `base_fee` alone.
+    pub fn new(bid: Balance) -> Result<Self, InvalidBid> {
```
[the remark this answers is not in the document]
NunoAlexandre: >  It will have to be passed when it stops being a dummy

Why?

> and we won't have to break this API then

What do you mean by breaking the API?

> It will also introduce better separation of the logic.

Could you expand?

runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::Balance;
+use frame_support::traits::WithdrawReason;
+
+/// Fee crate
```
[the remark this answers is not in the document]
NunoAlexandre: You are right, thanks!

runtime/src/fees/mod.rs
NunoAlexandre: The doc now is pointless, indeed. I just thought it'd look nicer on the generated docs.
Anyway, do you find the trait self-documenting enough as a reader?

runtime/src/fees/mod.rs
NunoAlexandre: I don't follow your confusion but I'd like us to address it. A "fee" has a concrete value associated with it. That value's unit is `Balance`. If this was `trait Car` and the `value` (or `price`) of the car would be expressed in Euros, but we'd still call the getter `value`/`price`, not `euros`. That's my interpretation and view on it. For me, "the value of a fee" is clear. I first called it `balance` (see my other PR), but that seemed a bit off since we now get the balance off accounts, so it could give the idea that we were getting the balance from a fee.

runtime/src/fees/mod.rs
NunoAlexandre: From 0 to 10, how strong do you feel about this?

runtime/src/fees/mod.rs
NunoAlexandre: It is indeed ~~a bit paranoid~~ unnecessary, deleted it. it had been copied and adjusted from other work here it made sense to have this.

runtime/src/fees/payment.rs
NunoAlexandre: > What does the <xxx> do? I'm really curious, the another account part is not visible in docs.

The `<xyz>` is just the `man` syntax for arguments, it doesn't stand for anything special here in the rust docs. It's part of the TODO.

> Should we put names in TODOs? Why? Maybe it would be better to describe what should be done, then it could be finished by anybody.

I always put my name but I very very very rarely leave a TODO when merging a PR, so it only serves as a reminder of things left to do before merging.

runtime/src/fees/payment.rs
NunoAlexandre: Good point. No, it shouldn't be an f64. It became clear after reading https://www.notion.so/monadic/Currency-bb0b3610cd534cde859b22f78f42b8b4.

runtime/src/fees/bid.rs
```diff
@@ -0,0 +1,96 @@
...
+pub enum InvalidBid {
+    /// The bid is insufficient to cover all mandatory costs.
+    Insufficient,
+}
+
+impl From<InvalidBid> for DispatchError {
+    fn from(error: InvalidBid) -> Self {
+        DispatchError::Module {
+            index: 42,
+            error: error as u8,
```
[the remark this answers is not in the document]
NunoAlexandre: Addressed at https://github.com/radicle-dev/radicle-registry/pull/241#discussion_r386997765.

runtime/src/fees/bid.rs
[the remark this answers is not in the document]
NunoAlexandre: Ah, I now get the overflow. Thanks!

runtime/src/fees/bid.rs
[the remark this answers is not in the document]
CodeSandwich: I think that the `Option` makes sense!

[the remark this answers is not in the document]
NunoAlexandre: Just pushed it 👍

runtime/src/fees/mod.rs
CodeSandwich: I agree that `foo()` doesn't necessarily have to return `Foo`. But the docs say that `Balance` is `Balance of an account`. It doesn't seem right to use it to hold value of the fee. Maybe we need a new type, e.g. a general `Value` or `Amount`, that could store a general amount of currency?

NunoAlexandre: Ah! Now I understand your problem 👌 

The docs should be corrected. That's the problem with mixing the definition of a type with where it is used in a specific case. The notion of `Balance` is transversal all over Substrate and it's really the general currency unit for "Value" / "Amount", thus we needing a `Balance` value when running `withdraw`. I see that that doc comment was copied from sample code from Substrate, I will adjust it. 👍

runtime/src/fees/mod.rs
CodeSandwich: I'd say around 7. Tests have some good practices and testing a single thing in a single test is one of them. Tests are often treated as secondary documentation and examples of usage, but it gets difficult when they don't do a single thing.

TBH I'm not sure if we need this test at all, it's not much more than the confirmation of a value of a constant.

runtime/src/fees/payment.rs
CodeSandwich: Aah, so it's not supposed to be merged yet? I've noticed that it was promoted from a draft to a regular PR and I thought that it's the potentially final version!

runtime/src/fees/mod.rs
CodeSandwich: I think that it's OK.

runtime/src/fees/payment.rs
CodeSandwich: The `<xyz>` won't work with 95% of Markdown parsers, it'll be interpreted as an HTML tag. That's why it disappeared.

NunoAlexandre: > I've noticed that it was promoted from a draft to a regular PR and I thought that it's the potentially final version!

You are right about this part. Before there was the draft feature, there was just a PR. For me, I prefer to mark it as ready even if there are a few things here and there to polish since probably the reviewers will comment on it anyway.

I will be more careful about this in coming PRs. In any case, the PR was ready for its core, but it wasn't ready for your review, that's why I had not requested it yet. But it doesn't hurt, no worries :)

runtime/src/fees/bid.rs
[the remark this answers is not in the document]
CodeSandwich: Right now in order to create a `Bid` all we need is its amount. `Bid` knows about the `BaseFee`, it creates a copy and applies it to itself. What I don't feel safe about here is that now it's not capable of working with different fees and it's not controllable how high it is. I think that it would be better to allow the users of a `Bid` construct all the fees and pass them into `new`. As soon as the `BaseFee` stops being just `1` (it's a developers' dummy, right?) it'll have to be somehow configurable. In current API desig it will require adding more parameters to `new` thus breaking it. I'm also not sure if `Bid` should really know how to build a `BaseFee` or any other fees, it's only job should be to apply them, the construction should be separated and put somewhere else.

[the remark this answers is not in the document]
NunoAlexandre: I think we are seeing this `Bid` type differently. From a user perspective, a bid is an amount they indicate to cover all transaction fees and have the rest being used as a tip. This `Bid` type basically gives body as to what it means in practice. having a total bid provided by the user, b, `Bid` will basically "spread that money" across the fees and use the rest as a tip; image having coins and having a slot of `base_fee` and another slot for `tip`. We'd put the contains to fill `base_fee` (mandatory) and the remainder we'd put in `tip`. That's essentially what it is doing. 

We don't need to receive BaseFee as an argument because it is a static and global value. The user doesn't get to define it, neither do we gain anything from having it as an argument since it that would be literally the same as having an argument for a global, static singleton. 

Now, the reason why `BaseFee` is a `Fee` and not just some constant value of type `Balance` is that the abstaction allows us to have a single `pay_fee` function that always does the right thing and doesn't care about what actual fee it is paying.

> 1 (it's a developers' dummy, right?

yes.

> it'll have to be somehow configurable

That's a question I raised before and it didn't seem to be something we have to worry about. The base_fee is static, global, and doesn't need to be configurable at runtime (if that's what you mean), as far as I understood. Ele will be posting the final values here, hopefully, today.

> I'm also not sure if Bid should really know how to build a BaseFee or any other fees, it's the only job should be to apply them, the construction should be separated and put somewhere else.

That's one of the main points for why this type exists in the first place, to dry and isolate the knowledge on how we can go to a raw bid on the user side (Balance) to something that breaks it down into the fees we support.

runtime/src/fees/mod.rs
NunoAlexandre: One of the great things about tests is that they tell you if just messed up. This test might seem redundant, but it defines the specification for each fee's withdraw reason. Shall we change that by mistake or for whatever reason, tests should fail to alert for that change in the specification.

As for your initial comment, I understand where you are coming from. I like to have one single test that asserts that the withdraw reasons for all the Fees we support are well defined, instead of having to look at multiple different tests to see what each should be. This is a trade-off, which I find more than fair over a mere principle.

>  This test is designed to validate 2 different behaviors of 2 different structs.

This is testing the property `withdraw_reason` of all the items that fall into the concept of `Fee`.

runtime/src/fees/mod.rs
```diff
@@ -0,0 +1,87 @@
...
+///
+/// This crate defines all things fees:
+/// * the types of fees supported by the registry
+/// * the [crate::fees::Bid] type that abstracts the concept of bid.
```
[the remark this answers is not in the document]
NunoAlexandre: Right, it should be `crate::fees::bid::Bid`.

runtime/src/fees/payment.rs
```diff
@@ -0,0 +1,36 @@
...
+const _FEE_PAYMENT_BURN: Balance = 1;
```
CodeSandwich: So now it's much safer. If fee is 0 or 1, the miner doesn't get anything, if higher, he gets `fee - 1` coins, right? Or do we reject transaction when fee is smaller than burn?

NunoAlexandre: After talking with Ele, we will have the burn being 1% of the negative_imbalance obtained from `withdraw`, which we will then `floor` to get a `u128` again. This isn't used at the moment and I have a TODO in place to finish this place so I'd just leave it as is and address it in the coming PR.

core/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub type Hashing = BlakeTwo256;
 /// Each account has an associated [message::AccountBalance] and [message::Index].
 pub type AccountId = ed25519::Public;
 
-/// Balance of an account.
+/// Balance currency type.
```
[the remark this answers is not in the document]
NunoAlexandre: 👌

runtime/src/fees/payment.rs
CodeSandwich: There's still an invalid HTML tag :(

runtime/src/fees/payment.rs
CodeSandwich: Good strategy. Could you put that in a comment so anybody can expect it to change in the near future?

core/src/lib.rs
```diff
@@ -47,7 +47,9 @@ pub type Hashing = BlakeTwo256;
 /// Each account has an associated [message::AccountBalance] and [message::Index].
 pub type AccountId = ed25519::Public;
 
-/// Balance of an account.
+/// The non-negative balance of anything storing the amount of currency.
+// It can be used to represent the value of anything describing an amount,
```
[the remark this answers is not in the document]
NunoAlexandre: Good catch, thanks!

runtime/src/fees/payment.rs
NunoAlexandre: Done.

runtime/src/fees/payment.rs
NunoAlexandre: Yeah, so the initial idea was to have this implemented before merging this PR, but - as discussed during team sync - I find it better to scope this PR as it is now and implements this function in an upcoming PR. In this case, I think it's fine to leave the TODO since I _really_ will follow it up.

PR #242, opened 2020-03-05
Add CONTRIBUTING.md

CodeSandwich:
Puts in motion https://www.notion.so/monadic/Contributing-md-draft-f45267e3c3424054b76456d648b69cf4 and https://www.notion.so/monadic/Project-abbreviations-glossary-draft-236640263822425aa5f5192c5fd884c6. They can be deleted when this PR is merged.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,54 @@
+# Abbreviations
+You may use common abbreviations in the codebase as long as they:
+- don't obscure their meaning for the sole sake of saving a few characters
+- are common programming abbreviations (e.g. `num`, `idx` or `iter`)
+
+For consistency we always use the following domain-specific abbreviation. You
+may not use any other domain-specific abbreviations.
+- Tx - Transaction
+- PoW - Proof of Work
+- Org - Organization
+
+# Git Flow
+
+TL;DR: This repository follows the Git rebase flow for the most part. Each feature branch must
+be reviewed and approved in a pull request. Once ready to merge, consider squashing the branch's
+commits when the separate commits don't add value, rebase it, force push with lease, and merge it
+via the GitHub UI.
+
+## Branches
+
+1. Create a separate branch for each issue your are working on
+2. Do your magic
+3. Keep your branch up to date by rebasing it from its base branch
+4. Delete the branch after its been both approved and merged. GitHub does this automatically.
+
+## Commits
+
+1. Make sure you author your commits with the right username and email
+2. Follow the git commit convention:
+  - Use the imperative mood in the subject line
+  - Limit the subject line to 50 chars
+  - Capitalise the subject line
+  - Wrap the description at 72 characters
+  - Have the description preferably explaining what and why instead of how
+  - Separate the subject from the body with an empty line
+
+# Documentation
+
+- Document all the public items of a module especially well given that they constitute its API.
+- Document private items when the reader's understanding to fully grasp its usage or implementation
+  is insufficient, despite the limited context of such items.
+- The code must be documented using Rustdoc comments.
+- Strive for self-documenting code.
+- The reader of the documentation is expected to have general knowledge in the fields of blockchains
+  and cryptocurrencies.
+- Leave additional comments explaining 'why' rather than 'how', strive to have the code clean
+  and elegant to make this aim viable.
+
+# Changelog
+
+We use `./CHANGELOG.md` to record all changes visible to users of the client.
+Changes are added to the “Upcoming” section of the change log as part of commit
+that makes the change. That is they are included in every pull request. For
+breaking changes a migration path must be provided.
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -13,12 +13,10 @@ The code is bootstrapped with the [`substrate-node-template`][node-template].
 - [Running development node](#running-development-node)
 - [Packages](#packages)
 - [Running checks and tests](#running-checks-and-tests)
-- [Changelog](#changelog)
 - [Continuous Deployment](#continuous-deployment)
 - [Local devnet](#local-devnet)
 - [Updating substrate](#updating-substrate)
 - [Updating Continuous Integration's base Docker image](#updating-continuous-integrations-base-docker-image)
-- [Git Flow](#git-flow)
 
 <!-- tocstop -->
 
@@ -96,16 +94,6 @@ To run specific tests sequentially as opposed to the parallel default,
 we use the [serial-test](https://crates.io/crates/serial_test) crate, simply
 having to mark the targeted tests with `#[serial]`.
 
-
-Changelog
----------
-
-We use `./CHANGELOG.md` to record all changes visible to users of the client.
-Changes are added to the “Upcoming” section of the change log as part of commit
-that makes the change. That is they are included in every pull request. For
-breaking changes a migration path must be provided.
-
-
 Continuous Deployment
 ---------------------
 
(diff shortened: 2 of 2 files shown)


ABBREVIATIONS.md
```diff
@@ -0,0 +1,6 @@
+These are the domain-specific abbreviations that are accepted and highly recommended across
+the whole codebase as long as they don't obscure the meaning.
+
+- Tx - transaction
+- PoW - proof of work
+- Org - organization
```
NunoAlexandre: What do you think about moving this small file into a `# Glossary` section in `CONTRIBUTING.md`?

CodeSandwich: Our current `CONTRIBUTING.md` is still small, but it probably will grow. When it does, finding the glossary inside will be a chore. `ABBREVIATIONS.md` should be quickly accessible and searchable, it's supposed to be an every day tool for a developer.

NunoAlexandre: I get your point, but I'd have the Glossary right on the top to give context to everything else, so it'd be as easy to find as it could be. In any case, Ctrl + F and voilá. 

> but it probably will grow.

Maybe, maybe not. Not it's less than 50 lines and we don't have planned work to make it bigger to justify this separation at this time. Shall the time come when this file is too big, then we think of a good day to split it, but now it's of no value, in my opinion.

CodeSandwich: I prefer the separation of concerns. One of them directly describes the meaning of code and the other human behaviors and soft conventions. One of them is updated every 10 commits or so, the other one when there's a collaboration problem.

I'm sure that `CONTRIBUTING.md` will grow, it's inevitable. Right now we have just a few small rules after 2 weeks of work, but there are still hundreds of upcoming events that will force us to officially set new ones.

The separate file is also much more discoverable, it's impossible to open the project and miss the fact that we've established a glossary. Who reads `CONTRIBUTING.md` before reading the code? Why would anybody looks for explanation of a weird blob of characters in `CONTRIBUTING.md`?

NunoAlexandre: @xla as a potential soon-to-be contributor of the registry, what's your opinion?

NunoAlexandre:
@CodeSandwich one other thing, this PR introduces an overlap between `DEVELOPING.md` and `CONTRIBUTING.md` that needs to be addressed.

ABBREVIATIONS.md
xla: Prematurely separating the information doesn't seem advisable. Besides that something like an `ABBREVATIONS.md` is extremely uncommon, keeping the definition close to where they are used will reduce surprises and the burden put on the reader to find the location of said definitions.

Generally this extraction makes it more obscure than clarifies. Usual navigation, reading and access patterns don't work well for fragmented documentation. A `CONTRIBUTING.md` is very common and I believe even suggested by Github as part of the project completion checklist.

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+Some abbreviations are widely understood and used in the software developer community in general,
+e.g. `num`, `idx` or `iter`. They should be used whenever they don't obscure the meaning.
+
+Other abbreviations are more domain-specific for blockchains, Substrate or just this project. They
+should be used whenever they don't obscure the meaning as well, but only if they are listed here:
+
+- Tx - transaction
+- PoW - proof of work
+- Org - organization
```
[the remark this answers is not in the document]
CodeSandwich: This is a neat improvement! I'm not sure about the `are well-established and agreed on in the context of the Blockchain, Substrate, or this project itself` part. This sentence removes the need for the list below, all of these 3 abbreviations are well established and agreed on. This is fine with me, but you were the first one to notice and rise alarm that they may not be understood by random contributors.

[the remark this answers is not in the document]
NunoAlexandre: Yeah, that's a good point.  We can just have the list to be explicit about which of those common abbreviations we use and agree on?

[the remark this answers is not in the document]
CodeSandwich: So we revert to not allowing well-established blockchain-specific abbreviations and meintening the list?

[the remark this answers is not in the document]
NunoAlexandre: yeah ok, whitelisting sounds like a good idea. I don't think there will be a lot of friction but yeah, whitelisting seems better here. 👍

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
+# Use of abbreviations
```
geigerzaehler: How do you feel about moving this section to the bottom? I think it’s the least relevant for users looking at this document for the first time.

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+Usage of abbreviations in the codebase is highly recommended for brevity.
```
geigerzaehler: When I read this the use of abbreviations sounds optional to me. This leaves the door open for inconsistencies like using `Tx` in some instances and `Transaction` in others. @NunoAlexandre’s suggestion has the same issue:

> The usage of abbreviations in the codebase is appreciated as long as they

If we want to go with abbreviations we should make use of them mandatory.

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+- Org - organization
+
+# Git Flow
+
+TL;DR: This repository follows the Git rebase flow for the most part. Each feature branch must
+be reviewed and approved in a pull request. Once ready to merge, consider squashing the branch's
+commits when the separate commits don't add value, rebase it, force push with lease, and merge it
+via the GitHub UI.
+
+## **Branches**
```
geigerzaehler: We do we mark this as bold?

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+- The code must be documented using Rustdoc comments.
+- Strive for self-documenting code.
+- The reader of the documentation is expected to have general knowledge in the fields of blockchain
+  and cryptocurrency.
+- Leave additional comments explaining 'why' rather than 'how', striving to have the code clean
+  and elegant to make this aim viable.
+- Document all the public items of a module especially well given that they constitute its API.
+- Document private items when the reader's understanding to fully grasp its usage or implementation
+  is insufficient, despite the limited context of such items.
```
geigerzaehler: I’d prefer to move these to the top since they are the most important points. They tell you the “what”, the rest the “how”.

CONTRIBUTING.md
NunoAlexandre: @geigerzaehler the argument is not to _sometimes_ use `tx` and other times use `transaction`. It's more to say `p` is not enough, `proj` is, just an example.

geigerzaehler: I understand that this is the intention. But I could be misread to mean the former so I think we should clarify it.

NunoAlexandre: Ah, I see what you mean 👍

CONTRIBUTING.md
NunoAlexandre: No need

CONTRIBUTING.md
CodeSandwich: I've put it on the top to make it the most accessible as a quick lookup and most visible for newcomers, because it's not a common thing for a project to have a dictionary.

geigerzaehler: That reasoning makes sense, too. Let’s keep it as it is.

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
+# Abbreviations
+The usage of common abbreviations in the codebase is appreciated as long as they:
```
geigerzaehler: Picking up on my [previous comment](https://github.com/radicle-dev/radicle-registry/pull/242#discussion_r389567057). To bring some more clarity around this I’d suggest to use the standard [RFC terminology](https://github.com/radicle-dev/radicle-registry/pull/242#discussion_r389567057).

```suggestion
You may use common abbreviations in the codebase as long as they:
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+- don't obscure their meaning for the sole sake of saving a few characters
+- are common programming abbreviations (e.g. `num`, `idx` or `iter`)
+
+The usage of domain-specific abbreviations is generally strongly discouraged.
+The only exception, when using abbreviations is mandatory, is for the names on this list:
```
geigerzaehler: Continuing the suggestion from above and being more prescriptive
```suggestion
For consistency we always use the following domain-specific abbreviation. You
may not use any other domain-specific abbreviations.
```

CodeSandwich: Fixed

CONTRIBUTING.md
CodeSandwich: Fixed

PR #243, opened 2020-03-06
Feature/plug transaction fees

NunoAlexandre:
Addresses #240 

### What

After adding the [foundation to charge tx fees](https://github.com/radicle-dev/radicle-registry/pull/241), in this second phase we plug it in and charge fees for all transactions.

- [x] Allow tx author to indicate a bid  through the client and the CLI (see [0])
- [x] Document tx fees for each transaction
- [x] Charge the appropriate account for each fee and each transaction
- [x] Unit test all possible scenarios (see [1])

Note that here we don't yet transfer the withdrew amount to the block author. That will be implemented in the following PR.

**[0]** In this PR, I added the `bid` field to each `core::message` since the bid is only useful at the registry.rs level, where we charge the fee. At that point, we only work with `message`, so adding this field here meant that no further changes were needed.  Adding it as a new field to Transaction or TransactionExtra would mean less redundancy but, on the other hand, it'd mean less flexibility (in the sense that bid is not a mandatory field for all transactions, forever), and more importantly, it would mean placing a field in a place where it's not used whatsoever, where we would need to pass it to to_runtime_call to have it in the registry.rs. Between a slight redundancy and the latter, I preferred the first. Happy to discuss it and find a better solution.

**[1]** Unit tests to verify tx fees correctness:
- [x] CreateCheckpoint
    - [x] Tx author has no funds enough to pay fees
    - [x] Bad previous_checkpoint_id
    - [x] Ok 
- [x] TransferFromOrg
    - [x] OK
    - [x] No funds enough to transfer
    - [x] Tx author has no funds enough to pay fees
    - [x] Bad actor
- [x] Transfer
    - [x] No funds enough to pay fees
    - [x] No funds enough to transfer
    - [x] OK
- [x] RegisterProject
    - [x] Org has no funds enough to pay bid
    - [x] Tx author has no funds enough to pay base fee
    - [x] Bad Actor
    - [x] Bad checkpoint
    - [x] Duplicate
    - [x] OK
- [x] UnRegisterOrg
    - [x] Org has no funds enough to pay bid
    - [x] Tx author has no funds enough to pay base fee
    - [x] Bad Actor
    - [x] Inexistent
    - [x] It has projects
    - [x] OK
- [x] RegisterOrg
    - [x] Too low bid
    - [x] Duplicate
    - [x] OK
    - [x] No funds to pay
- [x] SetCheckpoint
     - [x] Tx author has no funds enough to pay the base fee
    - [x] Inexistent checkpoint
    - [x] Bad Actor
    - [x] OK
    - [x] No such project
    - [x] No such org
    - [x] Org has no funds enough to pay bid
 - [x] Test end_to_end

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -147,6 +147,8 @@ impl CommandT for RegisterOrg {
                 &command_context.author_key_pair,
                 message::RegisterOrg {
                     org_id: self.org_id.clone(),
+                    //TODO(nuno) pass real bid here
+                    bid: 10,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -287,6 +287,8 @@ impl CommandT for Transfer {
                 message::Transfer {
                     recipient: self.recipient,
                     balance: self.funds,
+                    //TODO(nuno): pass real bid here
+                    bid: 10,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -235,6 +235,8 @@ impl CommandT for RegisterProject {
                     org_id: self.org_id.clone(),
                     checkpoint_id,
                     metadata: Bytes128::random(),
+                    //TODO(nuno): pass real bid here
+                    bid: 10,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -259,7 +259,7 @@ pub struct ShowGenesisHash {}
 impl CommandT for ShowGenesisHash {
     async fn run(&self, command_context: &CommandContext) -> Result<(), CommandError> {
         let genesis_hash = command_context.client.genesis_hash();
-        println!("Gensis block hash: 0x{}", hex::encode(genesis_hash));
+        println!("Genesis block hash: 0x{}", hex::encode(genesis_hash));
         Ok(())
     }
 }
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -330,6 +330,8 @@ impl CommandT for TransferOrgFunds {
                     org_id: self.org_id.clone(),
                     recipient: self.recipient,
                     value: self.funds,
+                    //TODO(nuno) Pass real bid here
+                    bid: 10,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -216,6 +216,8 @@ impl CommandT for RegisterProject {
                 message::CreateCheckpoint {
                     project_hash: self.project_hash.unwrap_or_default(),
                     previous_checkpoint_id: None,
+                    //TODO(nuno): pass real bid here
+                    bid: 10,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -222,7 +222,7 @@ impl CommandT for RegisterProject {
                 message::CreateCheckpoint {
                     project_hash: self.project_hash.unwrap_or_default(),
                     previous_checkpoint_id: None,
-                    bid: self.bid.clone(),
+                    bid: self.bid,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -149,7 +149,6 @@ impl CommandT for RegisterOrg {
                 &command_context.author_key_pair,
                 message::RegisterOrg {
                     org_id: self.org_id.clone(),
-                    bid: self.bid,
                 },
             )
             .await?;
@@ -181,7 +180,6 @@ impl CommandT for UnregisterOrg {
                 &command_context.author_key_pair,
                 message::UnregisterOrg {
                     org_id: self.org_id.clone(),
-                    bid: self.bid,
                 },
             )
             .await?;
@@ -222,7 +220,6 @@ impl CommandT for RegisterProject {
                 message::CreateCheckpoint {
                     project_hash: self.project_hash.unwrap_or_default(),
                     previous_checkpoint_id: None,
-                    bid: self.bid,
                 },
             )
             .await?;
@@ -240,7 +237,6 @@ impl CommandT for RegisterProject {
                     org_id: self.org_id.clone(),
                     checkpoint_id,
                     metadata: Bytes128::random(),
-                    bid: self.bid,
                 },
             )
             .await?;
@@ -295,7 +291,6 @@ impl CommandT for Transfer {
                 message::Transfer {
                     recipient: self.recipient,
                     balance: self.funds,
-                    bid: self.bid,
                 },
             )
             .await?;
@@ -340,7 +335,6 @@ impl CommandT for TransferOrgFunds {
                     org_id: self.org_id.clone(),
                     recipient: self.recipient,
                     value: self.funds,
-                    bid: self.bid,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -176,6 +176,8 @@ impl CommandT for UnregisterOrg {
                 &command_context.author_key_pair,
                 message::UnregisterOrg {
                     org_id: self.org_id.clone(),
+                    //TODO(nuno): receive real fee
+                    bid: 100,
                 },
             )
             .await?;
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -135,6 +135,8 @@ impl CommandT for ListProjects {
 pub struct RegisterOrg {
     /// Id of the org to register.
     org_id: OrgId,
+    // The bid to cover the tx fees and gain priority.
+    bid: Balance,
 }
 
 #[async_trait::async_trait]
@@ -147,8 +149,7 @@ impl CommandT for RegisterOrg {
                 &command_context.author_key_pair,
                 message::RegisterOrg {
                     org_id: self.org_id.clone(),
-                    //TODO(nuno) pass real bid here
-                    bid: 10,
+                    bid: self.bid,
                 },
             )
             .await?;
@@ -166,6 +167,8 @@ impl CommandT for RegisterOrg {
 pub struct UnregisterOrg {
     /// Id of the org to unregister.
     org_id: OrgId,
+    // The bid to cover the tx fees and gain priority.
+    bid: Balance,
 }
 
 #[async_trait::async_trait]
@@ -178,8 +181,7 @@ impl CommandT for UnregisterOrg {
                 &command_context.author_key_pair,
                 message::UnregisterOrg {
                     org_id: self.org_id.clone(),
-                    //TODO(nuno): receive real fee
-                    bid: 100,
+                    bid: self.bid,
                 },
             )
             .await?;
@@ -203,6 +205,10 @@ pub struct RegisterProject {
 
     /// Project state hash. A hex-encoded 32 byte string. Defaults to all zeros.
     project_hash: Option<H256>,
+
+    // The bid to cover the tx fees and gain priority.
+    // Charged once to create the checkpoint and then a second time to register the project.
+    bid: Balance,
 }
 
 #[async_trait::async_trait]
@@ -216,8 +222,7 @@ impl CommandT for RegisterProject {
                 message::CreateCheckpoint {
                     project_hash: self.project_hash.unwrap_or_default(),
                     previous_checkpoint_id: None,
-                    //TODO(nuno): pass real bid here
-                    bid: 10,
+                    bid: self.bid.clone(),
                 },
             )
             .await?;
@@ -235,8 +240,7 @@ impl CommandT for RegisterProject {
                     org_id: self.org_id.clone(),
                     checkpoint_id,
                     metadata: Bytes128::random(),
-                    //TODO(nuno): pass real bid here
-                    bid: 10,
+                    bid: self.bid,
                 },
             )
             .await?;
@@ -272,6 +276,8 @@ pub struct Transfer {
     recipient: AccountId,
     // The amount to transfer.
     funds: Balance,
+    // The bid to cover the tx fees and gain priority.
+    bid: Balance,
 }
 
 fn parse_account_id(data: &str) -> Result<AccountId, String> {
@@ -289,8 +295,7 @@ impl CommandT for Transfer {
                 message::Transfer {
                     recipient: self.recipient,
                     balance: self.funds,
-                    //TODO(nuno): pass real bid here
-                    bid: 10,
+                    bid: self.bid,
                 },
             )
             .await?;
@@ -319,6 +324,9 @@ pub struct TransferOrgFunds {
 
     // The balance to transfer from the org to the recipient.
     funds: Balance,
+
+    // The bid to cover the tx fees and gain priority.
+    bid: Balance,
 }
 
 #[async_trait::async_trait]
@@ -332,8 +340,7 @@ impl CommandT for TransferOrgFunds {
                     org_id: self.org_id.clone(),
                     recipient: self.recipient,
                     value: self.funds,
-                    //TODO(nuno) Pass real bid here
-                    bid: 10,
+                    bid: self.bid,
                 },
             )
             .await?;
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -107,6 +107,7 @@ pub struct RegisterProject {
 pub struct CreateCheckpoint {
     pub project_hash: H256,
     pub previous_checkpoint_id: Option<CheckpointId>,
+    pub bid: Balance,
 }
 
 /// Updates [crate::state::Project::current_cp].
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -152,6 +152,7 @@ pub struct TransferFromOrg {
     pub org_id: OrgId,
     pub recipient: AccountId,
     pub value: Balance,
+    pub bid: Balance,
 }
 
 /// Transfer funds from one account to another.
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -172,4 +172,5 @@ pub struct TransferFromOrg {
 pub struct Transfer {
     pub recipient: AccountId,
     pub balance: Balance,
+    pub bid: Balance,
 }
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -38,6 +38,7 @@ use parity_scale_codec::{Decode, Encode};
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
 pub struct RegisterOrg {
     pub org_id: OrgId,
+    pub bid: Balance,
 }
 
 /// Unregisters an org on the Radicle Registry with the given ID.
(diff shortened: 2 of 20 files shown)


runtime/src/registry.rs
```diff
@@ -144,6 +150,10 @@ decl_module! {
         #[weight = SimpleDispatchInfo::InsecureFreeNormal]
         pub fn register_org(origin, message: message::RegisterOrg) -> DispatchResult {
             let sender = ensure_signed(origin)?;
+            let bid: Bid = Bid::new(message.bid).ok_or(RegistryError::InsufficientBid)?;
```
NunoAlexandre: I'd like to dry this without making the `Bid::new` return a DispathResult or RegistryError. Maybe have a function in this module that does this for us? 

``` rust
make_bid(raw_bid: Balance) -> Either<Bid, DispatchError> {
   Bid::new(message.bid).ok_or(RegistryError::InsufficientBid)
}

File: cli/src/commands.rs [hunk 2]
```diff
@@ -135,6 +135,8 @@ impl CommandT for ListProjects {
 pub struct RegisterOrg {
     /// Id of the org to register.
     org_id: OrgId,
+    // The bid to cover the tx fees and gain priority.
+    bid: Balance,
```
geigerzaehler: Wouldn’t it make more sense to have the `bid` as a global option like the transaction author? This would eliminate some repetition and group the options properly

core/src/message.rs
```diff
@@ -35,9 +35,13 @@ use parity_scale_codec::{Decode, Encode};
 ///
 /// An Org with the same ID must not yet exist.
 ///
+/// # Transaction Fees
+///
+/// All the tx fees are paid by the tx author.
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
 pub struct RegisterOrg {
     pub org_id: OrgId,
+    pub bid: Balance,
```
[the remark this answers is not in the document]
geigerzaehler: I would prefer not to put the bid on the message and not to duplicate the fee charging code in the message handlers. This approach makes it easy to introduce bugs (for example by calling `pay_fee` incorrectly or forgetting it) and it makes it harder to refactor the code. It also makes it harder for contributors to add new messages because they need to know about fees.

runtime/src/registry.rs
```diff
@@ -108,20 +111,23 @@ decl_module! {
...
-                Some(o) => o,
-            };
+            let org = store::Orgs::get(message.org_id.clone())
+                .ok_or(RegistryError::InexistentOrg)?;
 
             if !org.members.contains(&sender) {
                 return Err(RegistryError::InsufficientSenderPermissions.into());
             }
 
+            pay_fee(bid.tip, &org.account_id)?;
```
geigerzaehler: What is the rationale behind charging the tip only if the transaction permissions are valid? Why don’t we charge everything up front or after all the validations have passed? This applies to other transactions as well.

cli/src/commands.rs
NunoAlexandre: Yes, it would. From the [discourse post](https://radicle.community/t/transaction-fees-on-the-registry-requirements-flows-plan-and-questions/139/6), concrete requirements were defined and discussed with Alexis. R2 means we need to charge (at least the tip) at the `registry.rs`.

Having that said, by adding `bid: Balance` to `sign_and_submit_message`, we'd need to pass it down the line, all the way to `into_runtime_call` so that we can meet R2. Doing this means we would need to change/break the APIs we have down to achieve the same thing that adding the field here does. Since that, for development, adding it here was a 3-second change, I just went ahead this way and have this discussion once everything else was done.

Again, I'm more than happy to clean this up. Hopefully, by discussing the high levels of this PR we can find better ways of doing it overall and see a cleaner way of doing this specific bit.

runtime/src/registry.rs
NunoAlexandre: The rationale is that some transactions involve orgs thus the org account should cover (some of) the tx fee costs. We decided to charge the base_fee to the tx author right away to avoid no fees from being charged at all in case of mal intentioned transactions, but wait until authorization (can) takes place to charge the tip from the org account.

core/src/message.rs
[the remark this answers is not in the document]
NunoAlexandre: There's an overlap with this other discussion: https://github.com/radicle-dev/radicle-registry/pull/243#discussion_r389620547

> This approach makes it easy to introduce bugs 

Unit testing the correctness of the tx fees charging is something that should be in place independently of how they are implemented. Having those tests, it's not easier to introduce bugs in either approach since they would be caught by the tests already.

> and it makes it harder to refactor the code

Not defending what we have atm, but how is that?

> It also makes it harder for contributors to add new messages because they need to know about fees.

Contributors would need to know about fees anyway; If they are adding new messages they would need to unit test them too.

cli/src/commands.rs
geigerzaehler: My comment here relates only to the CLI. Having the bid as a global (or common) option does not mean that we also need to change how it is implemented in the client or the ledger.

core/src/message.rs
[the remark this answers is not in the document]
geigerzaehler: > > This approach makes it easy to introduce bugs
> 
> Unit testing the correctness of the tx fees charging is something that should be in place independently of how they are implemented. Having those tests, it's not easier to introduce bugs in either approach since they would be caught by the tests already.

Having tests might increase the possibilities for bugs because the tests themselves might be lacking or buggy.

> > It also makes it harder for contributors to add new messages because they need to know about fees.
> 
> Contributors would need to know about fees anyway; If they are adding new messages they would need to unit test them too.

If we design this correctly then we don’t need to test the fees for every transaction because this is tested in a separate module and consistency is enforced by the code. Having to write the tests at the moment is a symptom of having to also write the charging code by hand.

> > and it makes it harder to refactor the code
> 
> Not defending what we have atm, but how is that?

Any change to the fee APIs (field names, function signatures, etc.) would need to be replicated across all of the transaction handlers.

geigerzaehler:
What is currently missing from this PR is fee validation. A full node needs to be able to validate that the fees that a user bid can be charged before including it in the block. Otherwise the network can be dos’ed. This needs to be implemented as part of the runtime so that the network rejects any block with an invalid transaction. In substrate this can be accomplished by implementing `SignedExtension` for a transaction extra.

With this PR the protocol still accepts transactions for the balances pallet but we don’t charge fees for these transactions. We will need to remove the balances pallet to avoid this.

core/src/message.rs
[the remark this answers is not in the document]
NunoAlexandre: > Having tests might increase the possibilities for bugs because the tests themselves might be lacking or buggy.

Having no code at all is the only way to not have bugs. Reading this, one could conclude that we'd be better off without tests?

> If we design this correctly then we don’t need to test the fees for every transaction because this is tested in a separate module and consistency is enforced by the code. Having to write the tests at the moment is a symptom of having to also write the charging code by hand.

If we can unit test the mechanism, I agree with this. The tests should not "trust" the implementation.

> Having to write the tests at the moment is a symptom of having to also write the charging code by hand.

The need to cover all scenarios through unit tests at the moment is a consequence of the requirements we need to meet and their implications as we foresaw them. Even with a dry setup, there are a lot of scenarios that might happen and, personally, I'd be really itchy to find out that our users are being wrongfully charged because we didn't cover all the cases that can and will happen. Again, I find that tests should be implementation agnostic.

NunoAlexandre:
@geigerzaehler thank you for all your fine feedback.

As discussed offline, we want to rethink this current solution not only to validate the bid before dispatch but also to dry the fee-charging mechanism.

As you also pointed out, in cases where the tx author bids a high value involving an org (that should pay for the fee) they don't belong to, currently no tip is paid, which means the tx gets prioritized while neither running fully, in the end, neither paying for what it has committed to.

You proposed creating a new `SignedExtension` implementation passed as a transaction extra to _validate_ that the tx author has enough funds to pay for the bid they have submitted. If so, the tx gets dispatched and _then_ in the registry module we would charge the entire fee from the tx author account, run the transaction, and finally, in transactions involving orgs, the org would reimburse the tx author if a member. This would satisfy the newly identified requirement of validating the tx before dispatch, but it still is too complex and easy to get wrong; drying the mechanism would help with that, though.

After thinking about this, a much simpler solution comes to mind. The whole reason for us to not use Substrate's default transaction payment is that it doesn't allow us to control who pays the fees; it always picks the tx author. Once we are able to pass the bid alongside with the Message in the transaction, instead of just validating the tx author to have enough funds, we can both obtain the message and know who must pay for the fees and authorize the tx author. I.e, **we can just charge the fees right away to the right account, or else the transaction doesn't get dispatched at all**. This seamlessly separates paying for transaction fees and running the transactions and avoids the need for reimbursements and whatnot.

Let me have your thoughts 🌱

NunoAlexandre:
@geigerzaehler I have pushed some of my local changes to show you preemptively where I am going with this v2 solution where the tx is validated and the code and logic around charging the fees are dry and simplified.

First, we add a `PayTxFee` impl of `SignedExtension` which will validate the tx by charging the given bid to the correct account.

https://github.com/radicle-dev/radicle-registry/blob/19e55ac38b61417b677c5fb1e7add886c8f59490/runtime/src/lib.rs#L214-L247

Basically when we handle a `Call::Registry`, we call `new_pay_fee` (tmp name), which will handle all the payment logic for us. If that succeeds, then we consider the tx valid and will set the priority accordingly to the tip (now I use a hardcode value since the `ValidTransaction.priority` field is of type `u64`, different than our `Balance` (tip) type). **Q**: I guess we should also charge the bid for all other calls?

Now, in `fees::payment` we now have the logic around what the pipeline to pay a fee is and how to decide who pays in the end.

The pipeline is:

https://github.com/radicle-dev/radicle-registry/blob/19e55ac38b61417b677c5fb1e7add886c8f59490/runtime/src/fees/payment.rs#L36-L39

As to how we `decide_payee`, we first need to check `who_should_pay` and then authorize that step. It's some lines of code in the end but I find that it describes the problem exactly who we would talk about it and I like having all this logic in the `payment` module, exactly where I'd expect it to be, and that `core::Message` are left completely alienated from this stuff. Plus, it's super easy to unit test these inner and, therefore, simplify the higher-level unit tests . 

https://github.com/radicle-dev/radicle-registry/blob/19e55ac38b61417b677c5fb1e7add886c8f59490/runtime/src/fees/payment.rs#L63-L105


In sum, this solution grants us the following:

- Validation of transactions before dispatch. Transactions that can't be paid won't even be dispatched.
- Full control over who pays the fee
- Total separation between tx fee payment and tx dispatch handling
- A more declarative description of the problem


Again, I am rather happy with your feedback and with your insights into Substrate and cleaner solutions. Let me know how you like it! I am cherry-picking the final changes of this PR into a clean branch where we will have a dedicated (and way fewer) unit tests testing the mechanism this solution offers.

runtime/src/lib.rs
```diff
@@ -196,6 +198,56 @@ impl registry::Trait for Runtime {
...
+
+/// Pay the transaction fees with the given `bid`.
+/// The bid is meant to cover all mandatory fees and have the remainder
+/// used as a tip to increase the priority of the transaction in the network.
+#[derive(Debug, Encode, Decode, Clone, Eq, PartialEq)]
+pub struct PayTxFee {
+    pub bid: Balance,
+}
+
+impl SignedExtension for PayTxFee {
```
geigerzaehler: We should move this to the fee module

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+        RegistryCall::unregister_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::transfer_from_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::set_checkpoint(m) => TxFeePayee::Org(m.org_id),
+        _ => TxFeePayee::TxAuthor,
+    }
+}
+
+/// The Transaction Fee Payee.
+enum TxFeePayee {
+    /// Represents that it should be the given Org paying for the tx fees.
```
geigerzaehler:
```suggestion
    /// The given org pays for the fees
```

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+        RegistryCall::register_project(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::unregister_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::transfer_from_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::set_checkpoint(m) => TxFeePayee::Org(m.org_id),
+        _ => TxFeePayee::TxAuthor,
+    }
+}
+
+/// The Transaction Fee Payee.
+enum TxFeePayee {
```
geigerzaehler:
```suggestion
/// The payer of a transaction fee if the transaction is authorized
enum TxFeePayer {
```

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+/// charge the tx fees to the right account, which depends on the `registry_call`.
+pub fn new_pay_fee(
+    author: AccountId,
+    bid: Bid,
+    registry_call: RegistryCall,
+) -> Result<(), DispatchError> {
+    let payee = decide_payee(author, registry_call);
+    withdraw_fee(bid, &payee)
+        .map(burn)
+        .and_then(pay_block_author)?;
```
geigerzaehler: We’re mixing imperative and functional style here. I would prefer imperative style because it is easier to read and more idiomatic
```suggestion
    let withdrawn_fee = withdraw_fee(bid, &payee)?;
    let reward = burn(with_drawn_fee);
    pay_block_author(reward);
```

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+fn decide_payee(author: AccountId, registry_call: RegistryCall) -> AccountId {
+    match who_should_pay(registry_call) {
+        TxFeePayee::Org(org_id) => match store::Orgs::get(org_id) {
+            Some(org) => {
+                if org.members.contains(&author) {
+                    org.account_id
+                } else {
+                    author
+                }
+            }
```
geigerzaehler: This code should live in the `registry` module and should be reused by the handlers. We should also split the authorization logic (`org.members.contains(&author)` from the logic that retrieves the account ID.

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+/// given that mal intended `RegistryCall`s might be issued that
+/// need to be authorized or else bad actors would be bankrupting
+/// innocent accounts.
+fn who_should_pay(registry_call: RegistryCall) -> TxFeePayee {
+    match registry_call {
+        RegistryCall::register_project(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::unregister_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::transfer_from_org(m) => TxFeePayee::Org(m.org_id),
+        RegistryCall::set_checkpoint(m) => TxFeePayee::Org(m.org_id),
+        _ => TxFeePayee::TxAuthor,
```
geigerzaehler: It’s probably prudent to list all calls explicitly. This prevents adding a call without having to figure out who is charged for the fee.

runtime/src/fees/payment.rs
```diff
@@ -13,24 +13,119 @@
...
+    }
+}
+
+/// Check who should pay for a given `RegistryCall`, that being
+/// either the tx author or an involved Org. This function does
+/// not determine whether the resolved [TxFeePayee] _must_ pay,
+/// given that mal intended `RegistryCall`s might be issued that
+/// need to be authorized or else bad actors would be bankrupting
+/// innocent accounts.
+fn who_should_pay(registry_call: RegistryCall) -> TxFeePayee {
```
geigerzaehler: I have the feeling that we need this much documentation because we’re lacking the right terms to describe what is going on and are unable to provide a descriptive name for the function.

How about about we rename `TxFeePayee` to `ChargeType`? This would make it clearer that the value just determines the charging mechanism but does not describe _who_ is charged.

runtime/src/fees/bid.rs
```diff
@@ -42,6 +42,16 @@ impl Bid {
...
 
+impl Fee for Bid {
+    fn value(&self) -> Balance {
+        self.base_fee.value() + self.tip.value()
+    }
+
+    fn withdraw_reasons(&self) -> frame_support::traits::WithdrawReasons {
+        self.base_fee.withdraw_reasons() | self.tip.withdraw_reasons()
+    }
+}
```
geigerzaehler: This indicates to me that we don’t actually need separate `BaseFee` and `Tip` structs. In fact, I believe we woudn’t need a `Fee` struct since `TxPayFee` only has a `fee: Balance` field.

I think it would greatly simplify the code if we drop all this.

NunoAlexandre: Yeah, absolutely.

runtime/src/fees/payment.rs
NunoAlexandre: > I have the feeling that we need this much documentation because we’re lacking the right terms to describe what is going on and are unable to provide a descriptive name for the function.

Keep in mind this is wip. I wrote the docs before writing the code just to think it through beforehand. We don't need this much docs, in my opinion the problem is self-explanatory, and, at worst, it's understandable within its context.

> How about about we rename TxFeePayee to ChargeType? This would make it clearer that the value just determines the charging mechanism but does not describe who is charged.

Happy to reformulate but `ChargeType` doesn't cut it. It doesn't determine a mechanism per se, only who, purely based on the transaction at hand, would be supposed to pay. I will think more about this.

runtime/src/fees/payment.rs
NunoAlexandre: Yeah, you are right. I first did that but then figured that since we will unit test this that it wouldn't be necessary but I agree it's better to be safe than sorry.

runtime/src/fees/payment.rs
NunoAlexandre: > This code should live in the registry module and should be reused by the handlers

Which part exactly are you referring to? I'd say the `org.members.contains(&author)` part but it becomes unclear with the rest of the paragraph.

runtime/src/fees/payment.rs
NunoAlexandre: https://media.giphy.com/media/d7rvF20PqNuGKSQGhf/giphy.gif

No, but I like your suggestion better too 👍

runtime/src/fees/payment.rs
NunoAlexandre: 👌

runtime/src/lib.rs
NunoAlexandre: Yeah, I was thinking the same. We won't need the current `Fee` and we could have this in `lib.rs` instead.

runtime/src/fees/payment.rs
geigerzaehler: I initially meant the whole case. But I think you’re right that probably only `org.members.contains(&author)` should be extracted.

NunoAlexandre:
Superseded by https://github.com/radicle-dev/radicle-registry/pull/255.

Issue #239, opened 2020-02-28
Adjust OrgId and ProjectName to new required validations


NunoAlexandre:
See: https://github.com/radicle-dev/registry-spec/pull/10#pullrequestreview-366403867

PR #246, opened 2020-03-06
Adjust org id validation

xla:
First part of #239 - addressing org ids.

### DIFF
--- a/core/src/org_id.rs
+++ b/core/src/org_id.rs
@@ -1,0 +1,225 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// `OrgId` is the unique identifier for organisations.
+///
+/// https://github.com/radicle-dev/registry-spec/blob/master/body.tex#L110
+use alloc::prelude::v1::*;
+use core::convert::{From, Into, TryFrom};
+use parity_scale_codec as codec;
+
+use crate::string32;
+
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct OrgId(String);
+
+impl OrgId {
+    fn from_string(input: String) -> Result<Self, InvalidOrgIdError> {
+        // Must be at least 1 character.
+        if input.is_empty() {
+            return Err(InvalidOrgIdError("must be at least 1 character"));
+        }
+        // Must be no longer than 32.
+        if input.len() > 32 {
+            return Err(InvalidOrgIdError("must not exceed 32 characters"));
+        }
+        // Must only contain a-z, 0-9 and '-' characters.
+        if !input
+            .chars()
+            .all(|c| c.is_ascii_digit() || c.is_ascii_lowercase() || c == '-')
+        {
+            return Err(InvalidOrgIdError("must only include a-z, 0-9 and '-'"));
+        }
+
+        // Must not start with a '-'.
+        if input.starts_with('-') {
+            return Err(InvalidOrgIdError("must not start with a '-'"));
+        }
+        // Must not end with a '-'.
+        if input.ends_with('-') {
+            return Err(InvalidOrgIdError("must not end with a '-'"));
+        }
+        // Must not contain sequences of more than one '-'.
+        if input.contains("--") {
+            return Err(InvalidOrgIdError(
+                "must not have more than one consecutive '-'",
+            ));
+        }
+
+        let id = Self(input);
+
+        Ok(id)
+    }
+}
+
+impl codec::Decode for OrgId {
+    fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+        let decoded: String = String::decode(input)?;
+
+        match OrgId::try_from(decoded) {
+            Ok(id) => Ok(id),
+            Err(err) => Err(codec::Error::from(err.what())),
+        }
+    }
+}
+
+impl Into<String> for OrgId {
+    fn into(self) -> String {
+        self.0
+    }
+}
+
+impl TryFrom<String> for OrgId {
+    type Error = InvalidOrgIdError;
+
+    fn try_from(input: String) -> Result<Self, Self::Error> {
+        Self::from_string(input)
+    }
+}
+
+impl TryFrom<&str> for OrgId {
+    type Error = InvalidOrgIdError;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        Self::from_string(input.to_string())
+    }
+}
+
+impl TryFrom<string32::String32> for OrgId {
+    type Error = InvalidOrgIdError;
+
+    fn try_from(input: string32::String32) -> Result<Self, Self::Error> {
+        Self::from_string(input.into())
+    }
+}
+
+impl core::str::FromStr for OrgId {
+    type Err = InvalidOrgIdError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::from_string(s.to_string())
+    }
+}
+
+#[cfg(feature = "std")]
+impl core::fmt::Display for OrgId {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "OrgId({})", self.0)
+    }
+}
+
+/// Error type when conversion from an inordinate input failed.
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidOrgIdError(&'static str);
+
+impl InvalidOrgIdError {
+    /// Error description
+    ///
+    /// This function returns an actual error str.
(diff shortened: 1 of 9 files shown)


core/src/org.rs
```diff
@@ -0,0 +1,115 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use core::convert::{From, Into, TryFrom};
+use parity_scale_codec as codec;
+
+use crate::string32;
+
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct Id(string32::String32);
```
NunoAlexandre: I think it's better to name this fully, so `OrgId`. Otherwise, we'd always have to reference it as `crate::org::Id` or have no context.

xla: It's not advised to prefix types with the module name, as it stutters. There is no drawback to `org::Id` in comparison to `OrgId`. Also the type is currently reexported as `OrgId` from `core`, so we should be safe.

NunoAlexandre: > It's not advised to prefix types with the module name, as it stutters.

I have never really heard this before as a best practice principle, although in my experience in Java the excessive namespace redundancy was sickening. So I surely know what you mean, but do you have a reference as to where _it's not advised_?

The stuttering argument is not really an issue since it'd only stutter when importing it.

> Also the type is currently reexported as OrgId from core, so we should be safe.

I see a drawback in needing to have this. Having it is essentially a symptom that `org::Id` isn't optimal. Moreover, when going through this module and reading `Id`, we will need to know or lookup which module we are in; Once you do this for projects and users, it _will_ get in the way.

core/src/org.rs
```diff
@@ -0,0 +1,115 @@
+// Radicle Registry
```
NunoAlexandre: Module comment:

One other thing that comes to mind is that in `org.rs` one would expect to see the `Org` type, that is in `state.rs`. I am not fully fond of either, but understand both. But again, it's confusing and will also get in the way. Renaming it to `org_id`  would help, and better capture the goal of the module.

runtime/tests/project_registration.rs
```diff
@@ -91,7 +93,7 @@ async fn register_project_with_inexistent_org() {
     .result
     .unwrap();
 
-    let inexistent_org_id = random_string32();
+    let inexistent_org_id = OrgId::try_from(random_string32()).unwrap();
```
NunoAlexandre: It'd be much cleaner to have a dedicated random_org_id function:

``` rust
fn random_org_id() -> OrgId {
  OrgId::try_from(random_string32()).unwrap();
}
```

core/src/org.rs
xla: I followed your advise for now and will leave any further structural changes to follow-up PRs. IMHO `lib.rs` should be empty, apart from the module declarations and public exports. So I could see `org.rs` being the place to put all none state related types.

core/src/org.rs
xla: > I have never really heard this before as a best practice principle, although in my experience in Java the excessive namespace redundancy was sickening. So I surely know what you mean, but do you have a reference as to where _it's not advised_?

https://rust-lang.github.io/rust-clippy/master/#module_name_repetitions



> > 
> 
> I see a drawback in needing to have this. Having it is essentially a symptom that `org::Id` isn't optimal. Moreover, when going through this module and reading `Id`, we will need to know or lookup which module we are in; Once you do this for projects and users, it _will_ get in the way.

I don't think that's the right conclusion, the conclusion is that unqualified imports are suboptimal. Comparing `org::Id` and `OrgId` they carry the same information. Yet the former is more honest, as in the domain of the `org` it is the `Id` type. And the domain of the org is defined by the module boundary, that there is another understanding of an id in another domain, i.e. `user` or `project` (so that one needs more investigation as the `ProjectId` might be a compound only relevant outside of the project mod) is to be expected especially with generic concepts like identifiers. It's the art of consolidating these domains higher up, which in turn shouldn't force these constraints downstream.

core/src/string32.rs
```diff
@@ -30,7 +30,7 @@ use parity_scale_codec::{Decode, Encode, Error as CodecError, Input};
 pub struct String32(String);
 
 impl String32 {
-    const MAXIMUM_SUPPORTED_LENGTH: usize = 32;
+    pub const MAXIMUM_SUPPORTED_LENGTH: usize = 32;
```
cloudhead: No longer needed.

core/src/org_id.rs
```diff
@@ -0,0 +1,227 @@
...
+    use parity_scale_codec::{Decode, Encode};
+
+    #[test]
+    fn id_too_short() {
+        assert!(OrgId::from_string("".into()).is_err());
+    }
+
+    #[test]
+    fn id_too_long() {
+        let input = std::iter::repeat("X").take(33).collect::<String>();
```
NunoAlexandre: you could use `MAX_SUPPORTED_LENGTH + 1` here instead of `33`

core/src/org.rs
NunoAlexandre: Thanks for the link!

> Why is this bad
> It requires the user to type the module name twice.

🤷‍♂ ok. And drawbacks?

Anyway, I get your point. Thanks for expanding on it!

core/src/org_id.rs
xla: Did that in the beginning, but didn't like the apparent leak of `String32` internals. WIth the following project name changes I plan to consolidate the shared validations in a reworked version of String32.

core/src/org_id.rs
```diff
@@ -0,0 +1,227 @@
...
+        if input.is_empty() {
+            return Err(InordinateOrgIdError("must be at least 1 character"));
+        }
+        // Must be no longer than 32.
+        if input.len() > 32 {
+            return Err(InordinateOrgIdError("must not exceed 32 characters"));
+        }
+        // Must only consist of ASCII-encoded characters.
+        if !input.is_ascii() {
+            return Err(InordinateOrgIdError("must be valid ASCII"));
```
geigerzaehler: This is also validated by the check below. If the check below fails then the input is also not valid ASCII. We can omit this check.

core/src/org_id.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        Ok(id)
+    }
+}
+
+impl codec::Decode for OrgId {
+    fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+        let decoded: String = String::decode(input)?;
+
+        match OrgId::try_from(decoded) {
```
geigerzaehler: Can we have at least one test that ensures that the validations are also run on decoding?

core/src/org_id.rs
```diff
@@ -0,0 +1,227 @@
...
+#[cfg(feature = "std")]
+impl core::fmt::Display for OrgId {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "OrgId({})", self.0)
+    }
+}
+
+/// Error type when conversion from an inordinate input failed.
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InordinateOrgIdError(&'static str);
```
geigerzaehler: The term “inordinate” does not seem appropriate anymore. It makes sense if the input may be too long. I’d just go with `InvalidOrgIdError`.

test-utils/src/lib.rs
```diff
@@ -82,6 +84,10 @@ pub async fn create_random_org(client: &Client, author: &ed25519::Pair) -> Org {
         .unwrap()
 }
 
+pub fn random_org_id() -> OrgId {
+    OrgId::try_from(random_string(16).to_lowercase()).unwrap()
```
geigerzaehler: I’d prefer to also randomize the length.

core/src/org_id.rs
```diff
@@ -0,0 +1,227 @@
...
+
+impl InordinateOrgIdError {
+    /// Error description
+    ///
+    /// This function returns an actual error str.
+    pub fn what(&self) -> &'static str {
```
geigerzaehler: Is this useful? It is the same as `Error::description`.

core/src/org_id.rs
NunoAlexandre: Right, that's a good point that I missed!

core/src/org_id.rs
xla: There is a semantic difference, one is that we check for valid encoding and the other one is to ensure we use the allowed characterset. I personally see value in clearly communicating which validation failed. But happy to consolidate if we think it's superfluous.

core/src/org_id.rs
xla: Renamed in 970d57f.

test-utils/src/lib.rs
xla: Randomized in d45e0bb.

core/src/org_id.rs
geigerzaehler: I’d prefer to consolidate this and have less code to think about.

core/src/org_id.rs
xla: Test added in 7a2f2a4.

core/src/org_id.rs
xla: Yeah, except that this is avaliable in `no_std`. Do we have conventions, needs for what we need from errors in that context?

core/src/org_id.rs
xla: Removed in d73ccb8.

core/src/org_id.rs
```diff
@@ -224,4 +224,12 @@ mod test {
 
         assert_eq!(id, decoded)
     }
+
+    #[test]
+    fn encoded_then_decode_invalid() {
+        let invalid = Encode::encode("-Invalid-");
+        let decoded = <OrgId>::decode(&mut &invalid[..]);
```
geigerzaehler:
```suggestion
        let decoded = OrgId::decode(&mut &invalid[..]);
```

PR #247, opened 2020-03-09
Adjust project name validations

### LINKED ISSUE #239: Adjust OrgId and ProjectName to new required validations

xla:
Closes #239 - second part addressing project names.

### DIFF
--- a/client/examples/project_registration.rs
+++ b/client/examples/project_registration.rs
@@ -19,7 +19,7 @@ async fn go() -> Result<(), Error> {
     let node_host = url::Host::parse("127.0.0.1").unwrap();
     let client = Client::create(node_host).await?;
 
-    let project_name = ProjectName::from_string("radicle-registry".to_string()).unwrap();
+    let project_name = ProjectName::try_from("radicle-registry").unwrap();
     let org_id = OrgId::try_from("monadic").unwrap();
 
     // Choose some random project hash and create a checkpoint
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -39,6 +39,9 @@ pub use string32::String32;
 mod org_id;
 pub use org_id::OrgId;
 
+mod project_name;
+pub use project_name::ProjectName;
+
 mod error;
 pub use error::RegistryError;
 
@@ -57,9 +60,6 @@ pub type Balance = u128;
 
 /// The id of a project. Used as storage key.
 pub type ProjectId = (ProjectName, OrgId);
-
-/// The name a project is registered with.
-pub type ProjectName = String32;
 
 /// Org
 ///
--- a/core/src/project_name.rs
+++ b/core/src/project_name.rs
@@ -1,0 +1,208 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// The name associated to a [`Project`].
+///
+/// https://github.com/radicle-dev/registry-spec/blob/master/body.tex#L306
+use alloc::prelude::v1::*;
+use core::convert::{From, Into, TryFrom};
+use parity_scale_codec as codec;
+
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct ProjectName(String);
+
+impl ProjectName {
+    fn from_string(input: String) -> Result<Self, InvalidProjectNameError> {
+        // Must be at least 1 character.
+        if input.is_empty() {
+            return Err(InvalidProjectNameError("must be at least 1 character"));
+        }
+        // Must be no longer than 32.
+        if input.len() > 32 {
+            return Err(InvalidProjectNameError("must not exceed 32 characters"));
+        }
+
+        // Must only contain a-z, 0-9, '-', '_' and '.' characters.
+        {
+            let check_charset = |c: char| {
+                c.is_ascii_digit() || c.is_ascii_lowercase() || c == '-' || c == '_' || c == '.'
+            };
+
+            if !input.chars().all(check_charset) {
+                return Err(InvalidProjectNameError(
+                    "must only include a-z, 0-9, '-', '_' and '.'",
+                ));
+            }
+        }
+
+        // Must not equal '.' or '..'.
+        if input == "." || input == ".." {
+            return Err(InvalidProjectNameError("must not be equal to '.' or '..'"));
+        }
+
+        let id = Self(input);
+
+        Ok(id)
+    }
+}
+
+impl codec::Decode for ProjectName {
+    fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+        let decoded: String = String::decode(input)?;
+
+        match Self::try_from(decoded) {
+            Ok(id) => Ok(id),
+            Err(err) => Err(codec::Error::from(err.what())),
+        }
+    }
+}
+
+impl Into<String> for ProjectName {
+    fn into(self) -> String {
+        self.0
+    }
+}
+
+impl TryFrom<String> for ProjectName {
+    type Error = InvalidProjectNameError;
+
+    fn try_from(input: String) -> Result<Self, Self::Error> {
+        Self::from_string(input)
+    }
+}
+
+impl TryFrom<&str> for ProjectName {
+    type Error = InvalidProjectNameError;
+
+    fn try_from(input: &str) -> Result<Self, Self::Error> {
+        Self::from_string(input.to_string())
+    }
+}
+
+impl core::str::FromStr for ProjectName {
+    type Err = InvalidProjectNameError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::from_string(s.to_string())
+    }
+}
+
+#[cfg(feature = "std")]
+impl core::fmt::Display for ProjectName {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Error type when conversion from an inordinate input failed.
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidProjectNameError(&'static str);
+
+impl InvalidProjectNameError {
+    /// Error description
+    ///
+    /// This function returns an actual error str when running in `std`
+    /// environment, but `""` on `no_std`.
+    #[cfg(feature = "std")]
+    pub fn what(&self) -> &'static str {
+        self.0
+    }
+
+    /// Error description
+    ///
+    /// This function returns an actual error str when running in `std`
+    /// environment, but `""` on `no_std`.
+    #[cfg(not(feature = "std"))]
+    pub fn what(&self) -> &'static str {
+        ""
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for InvalidProjectNameError {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "InvalidProjectNameError({})", self.0)
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for InvalidProjectNameError {
+    fn description(&self) -> &str {
+        self.0
+    }
+}
+
+impl From<&'static str> for InvalidProjectNameError {
+    #[cfg(feature = "std")]
+    fn from(s: &'static str) -> Self {
+        Self(s)
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn from(s: &'static str) -> Self {
+        InvalidProjectNameError(s)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::ProjectName;
+    use parity_scale_codec::{Decode, Encode};
+
+    #[test]
+    fn name_too_short() {
+        assert!(ProjectName::from_string("".into()).is_err());
+    }
+
+    #[test]
+    fn name_too_long() {
+        let input = std::iter::repeat("X").take(33).collect::<String>();
+        let too_long = ProjectName::from_string(input);
+        assert!(too_long.is_err());
+    }
+
+    #[test]
+    fn name_invalid_characters() {
+        let invalid_characters = ProjectName::from_string("AZ+*".into());
+        assert!(invalid_characters.is_err());
+    }
+
+    #[test]
+    fn name_is_dot() {
+        let dot = ProjectName::from_string(".".into());
+        assert!(dot.is_err());
+    }
+
+    #[test]
+    fn name_is_double_dot() {
+        let dot = ProjectName::from_string("..".into());
+        assert!(dot.is_err());
+    }
+
+    #[test]
+    fn name_valid() {
+        let valid = ProjectName::from_string("--radicle_registry001".into());
+        assert!(valid.is_ok());
+    }
+
+    #[test]
+    fn encode_then_decode() {
+        let id = ProjectName::from_string("monadic".into()).unwrap();
+        let encoded = id.encode();
+        let decoded = <ProjectName>::decode(&mut &encoded[..]).unwrap();
+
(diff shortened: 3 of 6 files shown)


core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+        // Must be no longer than 32.
+        if input.len() > 32 {
+            return Err(InordinateProjectNameError("must not exceed 32 characters"));
+        }
+        // Must only consist of ASCII-encoded characters.
+        if !input.is_ascii() {
+            return Err(InordinateProjectNameError("must be valid ASCII"));
+        }
+        // Must only contain a-z, 0-9 and '-' characters.
```
cloudhead: Comment doesn't reflect code

core/src/ascii32.rs
```diff
@@ -13,79 +13,85 @@
...
+pub struct Ascii32(String);
 
-impl String32 {
+impl Ascii32 {
+    const MINIMUM_SUPPORTED_LENGTH: usize = 1;
     const MAXIMUM_SUPPORTED_LENGTH: usize = 32;
 
-    /// Returns an error if [String::len] of the provided is greater than 32.
-    pub fn from_string(s: String) -> Result<Self, InordinateStringError> {
+    /// Attempts the cnversion from a [String] input.
```
NunoAlexandre:
```suggestion
    /// Attempts the conversion from a [String] input.
```

core/src/ascii32.rs
```diff
@@ -13,79 +13,85 @@
...
 /// assert!(String32::from_string(long_string).is_err());
 /// ```
-#[derive(Encode, Clone, Debug, Eq, PartialEq)]
-pub struct String32(String);
+#[derive(Clone, codec::Encode, Debug, Eq, PartialEq)]
+pub struct Ascii32(String);
 
-impl String32 {
+impl Ascii32 {
+    const MINIMUM_SUPPORTED_LENGTH: usize = 1;
```
NunoAlexandre: `NonEmptyAscii32`? :P

core/src/ascii32.rs
```diff
@@ -13,79 +13,85 @@
...
     const MAXIMUM_SUPPORTED_LENGTH: usize = 32;
 
-    /// Returns an error if [String::len] of the provided is greater than 32.
-    pub fn from_string(s: String) -> Result<Self, InordinateStringError> {
+    /// Attempts the cnversion from a [String] input.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if [String::len] of the provided is lesser than 1, greater than 3 or
```
NunoAlexandre:
```suggestion
    /// Returns an error if the input is either:
    ///     * is empty
    ///     * is greater than 32 characters
    ///     * contains non-ASCII charactes
```

core/src/ascii32.rs
```diff
@@ -13,79 +13,85 @@
...
+    /// contains non-ASCII charactes.
+    fn from_string(s: String) -> Result<Self, InordinateInput> {
         if s.len() > 32 {
```
NunoAlexandre: You document already what makes for a valid `Ascii32` but are not yet implementing it here. Missing `is_empty` and the non-ascii characters check.

core/src/lib.rs
```diff
@@ -33,8 +33,17 @@ pub mod state;
 pub mod bytes128;
 pub use bytes128::Bytes128;
 
-pub mod string32;
-pub use string32::String32;
+pub mod ascii32;
+pub use ascii32::Ascii32;
```
NunoAlexandre: 🙊

core/src/ascii32.rs
NunoAlexandre: I don't think we'd want `Ascii32` to be non-empty though?

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::ProjectName;
+    use parity_scale_codec::{Decode, Encode};
+
+    #[test]
+    fn id_too_short() {
+        assert!(ProjectName::from_string("".into()).is_err());
```
NunoAlexandre: I'd prefer if we verify what the actual error is since changing/messing that up will have consequences for registry users, like the upstream.

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[cfg(feature = "std")]
+    pub fn what(&self) -> &'static str {
+        self.0
+    }
+
+    /// Error description
+    ///
+    /// This function returns an actual error str when running in `std`
+    /// environment, but `""` on `no_std`.
+    #[cfg(not(feature = "std"))]
```
NunoAlexandre: Smart!

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+/// https://github.com/radicle-dev/registry-spec/blob/master/body.tex#L306
+use alloc::prelude::v1::*;
+use core::convert::{From, Into, TryFrom};
+use parity_scale_codec as codec;
+
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct ProjectName(String);
+
+impl ProjectName {
+    fn from_string(input: String) -> Result<Self, InordinateProjectNameError> {
```
NunoAlexandre: Repeating Thomas' comment on your previous PR, this type should be renamed to `InvalidProjectNameError`.

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// The name a project is registered under.
```
NunoAlexandre: This might be a matter of language but a project is not registered _under_ its name. This can be confusing compared to a project being registered _under_ an org.

core/src/ascii32.rs
```diff
@@ -13,79 +13,85 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-/// `String32` type, and its validation tests.
+/// `Ascii32` type, and its validation tests.
```
NunoAlexandre: I understand why you do this, but I am not sure about it. But let's go through it together. You change it to `Ascii32` since it only represents a valid, no longer than 32, sequence of Ascii characters. That's good. What I don't quite like is that, shall we change that validation to be utf8, or something, else, we will need to touch the whole project again, whereas if we have a less specific name, `String32`, we say "This is a sequence of characters that follows such and such validations". By writing this, I recognize that the same argument applies to have the `32` in the naming.

Changed my mind, sounds good 👍

NunoAlexandre:
Btw, @xla, this is a good example of a PR where I'd rather keep some history instead of squashing everything. Please keep at least one commit for the changes around `Ascii32` and another for the changes around `ProjectName`.

core/src/project_name.rs
xla: I'm not quite sure this is what we want. We have `description` already as @geigerzaehler pointed out, but I'm not sure how it behaves in `no_std`. I took this from the `decode::Error` impl from the substrates. Maybe we can get some calrity and clear recommendation what impls we expect for error types to behave proper on all our targets.

core/src/project_name.rs
xla: This is already asserted in the method signature as we only return `Result<Self, InvalidProjectNameError>`. :>

core/src/project_name.rs
xla: Globally it is located under a `ProjectId`, yet it is still registered under the name, inside of an org/user. The statement is not incorrect, but I'd like to hear what you'd like to see here communicated?

cloudhead: I think what he's saying is that this comment implies hierarchy between the name and the project, when there actually is none. Rather, we have a "has a" relationship: the name is an attribute of the project.

cloudhead: So eg. `/// The name associated to the project` may be clearer

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+impl core::str::FromStr for ProjectName {
+    type Err = InvalidProjectNameError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::from_string(s.to_string())
+    }
+}
```
cloudhead: It's too bad we need both `FromStr` and `TryFrom<&str>`

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+
+#[cfg(feature = "std")]
+impl core::fmt::Display for ProjectName {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "OrgId({})", self.0)
```
cloudhead: copy/:spaghetti: error

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+    #[cfg(not(feature = "std"))]
+    pub fn what(&self) -> &'static str {
+        ""
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for InvalidProjectNameError {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "InvalidOrgIdError({})", self.0)
```
cloudhead: copy/pasta error

core/src/project_name.rs
NunoAlexandre: Exactly what @cloudhead is saying.

xla: Addressed in 4e69db5.

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+            return Err(InvalidProjectNameError("must be at least 1 character"));
+        }
+        // Must be no longer than 32.
+        if input.len() > 32 {
+            return Err(InvalidProjectNameError("must not exceed 32 characters"));
+        }
+        // Must only consist of ASCII-encoded characters.
+        if !input.is_ascii() {
+            return Err(InvalidProjectNameError("must be valid ASCII"));
+        }
```
CodeSandwich: This check is the same as the one below?

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+        // Must only contain a-z, 0-9, '-' and '_' characters.
+        if !input
+            .chars()
+            .all(|c| c.is_ascii_digit() || c.is_ascii_lowercase() || c == '-' || c == '_')
+        {
+            return Err(InvalidProjectNameError(
+                "must only include a-z, 0-9 and '-'",
+            ));
+        }
```
CodeSandwich: I feel that it would be cleaner to extract the validation logic to a separate function

core/src/project_name.rs
CodeSandwich: I'm honestly curious, why?

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Self::from_string(s.to_string())
+    }
+}
+
+#[cfg(feature = "std")]
+impl core::fmt::Display for ProjectName {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+        write!(f, "ProjectName({})", self.0)
```
CodeSandwich: This output looks debug-ish. Don't we need a pretty `Display` with just the project name? After all it's officially just a validated string.

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+/// Error type when conversion from an inordinate input failed.
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct InvalidProjectNameError(&'static str);
```
CodeSandwich: We're deep in a library territory, wouldn't it be good to have an enum here? I think that it would comply better with our new error handling guidelines too.

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Error description
+    ///
+    /// This function returns an actual error str when running in `std`
+    /// environment, but `""` on `no_std`.
+    #[cfg(not(feature = "std"))]
+    pub fn what(&self) -> &'static str {
+        ""
+    }
+}
```
CodeSandwich: Can we use `thiserror` here?

core/src/project_name.rs
```diff
@@ -0,0 +1,197 @@
...
+            return Err(InvalidProjectNameError("must not exceed 32 characters"));
+        }
+        // Must only consist of ASCII-encoded characters.
+        if !input.is_ascii() {
+            return Err(InvalidProjectNameError("must be valid ASCII"));
+        }
+        // Must only contain a-z, 0-9, '-' and '_' characters.
+        if !input
+            .chars()
+            .all(|c| c.is_ascii_digit() || c.is_ascii_lowercase() || c == '-' || c == '_')
```
CodeSandwich: I've just noticed: in specs the org IDs and the user IDs have other naming rules from project IDs. The former can't have an `_`, and treat `-`s specially. Is this intentional?

core/src/project_name.rs
CodeSandwich: I'm honestly curious, why?

xla: They provide different APIS (`from_str` and `try_from("")` respectively) I saw a `FromStr` implementation. It seems the main reason for `FromStr` being around and distinct from `TryFrom<T>` is for parse semantics: https://github.com/rust-lang/rfcs/issues/2143#issuecomment-327952645

I'm not certain about our  requirements, but happy to settle on of the methods.

core/src/project_name.rs
xla: Was a leftover and has been removed.

core/src/project_name.rs
xla: Contemplated it and had a discussin with @cloudhead. Any enum here wouldn't serve us any better, just introduce more error infrastructure. The context to why the input was invalid is carried in the description of the error and none of the reasons this can fail can be fixed by a machine. Ultimately this has to bubble up back to the user and input has to be corrected. Same goes for the decoding case, there is no recovery we can capture in code.

core/src/project_name.rs
xla: I still need input what we need want and what is possible in the `no_std` context.

core/src/project_name.rs
xla: Yes it's intentional. See radicle-dev/registry-spec#10 and radicle-dev/registry-spec#17.

core/src/project_name.rs
xla: Fixed in a1ec5ca.

core/src/project_name.rs
xla: This is already the extracted function for validation logic. :>

core/src/project_name.rs
CodeSandwich: The bubbling up to the user part is what may be exactly the core reason to have strict enums: they make nice errors in GUIs possible along with i11n. With enums it's not easy, but it's feasible, but with `&str`s it's a parsing nightmare.

xla: Certainly, currently we don't accoutn for i18n. We can revisit this in the context of an overhaul of error strategy and when we know that our validation rules have bene solidified.

core/src/project_name.rs
CodeSandwich: It's the validation logic inside the conversion/construction logic. It's not a big deal, so I won't be pushing more, but in my opinion it's a violation of the single responsibility rule.

client/examples/project_registration.rs
```diff
@@ -19,7 +19,7 @@ async fn go() -> Result<(), Error> {
     let node_host = url::Host::parse("127.0.0.1").unwrap();
     let client = Client::create(node_host).await?;
 
-    let project_name = ProjectName::from_string("radicle-registry".to_string()).unwrap();
+    let project_name = ProjectName::try_from("radicle-registry".to_string()).unwrap();
```
geigerzaehler: Can’t we get rid of `.to_string()`?

core/src/project_name.rs
geigerzaehler: I’d prefer to have one way to do it with `TryFrom` but I’m ok if we leave it. If we change it here we should align it with the org ID.

core/src/project_name.rs
geigerzaehler: Seems to be still here. (Maybe forgot to push?)

core/src/project_name.rs
xla: Did an audit of the codebase for occurrences of `from_str` and found none. Lo and behold, `structopt` derivation needs it anyway. I suspect for the parse semantics mentioned above. Will leave it for now and we can revise later.

PR #249, opened 2020-03-11
Implement users registry

### LINKED ISSUE #245: Implement users registry
With the updated understanding of users in radicle-dev/registry-spec#11 we ought to provide them as first-class concepts by implementing the necessary transactions.

### LINKED ISSUE #256: Typo in runtime
This looks like a leftover from explorations and should read `Registry`:
https://github.com/radicle-dev/radicle-registry/blob/8a1b4f5186139b699f8869ba42cba629c12c32dc/runtime/src/registry.rs#L45

xla:
Closes #245 
Closes #256


runtime/src/registry.rs
```diff
@@ -348,7 +379,7 @@ mod test {
 
     #[test]
     /// Test that store::Orgs::decode_key after store::Orgs::storage_map_final_key
-    /// is identify as to the original input id.
+    /// is identified as the original input id.
```
NunoAlexandre: this was my typo and its meant to be "identity":

```suggestion
    /// is identity as to the original input id.
```

runtime/src/registry.rs
```diff
@@ -358,7 +389,7 @@ mod test {
 
     #[test]
     /// Test that store::Projects::decode_key after store::Projects::storage_map_final_key
-    /// is identify as to the original input id.
+    /// is identii as the original input id.
```
NunoAlexandre:
```suggestion
    /// is identity as to the original input id.
```

runtime/tests/user_registration.rs
```diff
@@ -0,0 +1,43 @@
...
+    let user: client::User = client
+        .get_user(register_user_message.user_id.clone())
+        .await
+        .unwrap()
+        .unwrap();
+    assert_eq!(user.id, register_user_message.user_id);
+    assert!(user.projects.is_empty());
+}
+
+async fn user_exists(client: &client::Client, user_id: client::UserId) -> bool {
```
NunoAlexandre: You could move this to test-utils btw.

cli/src/commands.rs
```diff
@@ -243,6 +243,35 @@ impl CommandT for RegisterProject {
...
+                message::RegisterUser {
+                    user_id: self.user_id.clone(),
+                },
+            )
+            .await?;
+        println!("Registering user...");
+
+        let user_registered = register_user_fut.await?;
+        transaction_applied_ok(&user_registered)?;
+        println!("Org {} is now registered.", self.user_id);
```
cloudhead:
```suggestion
        println!("User {} is now registered.", self.user_id);
```

core/src/lib.rs
```diff
@@ -120,3 +123,30 @@ impl Project {
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct User {
+    // Unique id of a User.
+    pub id: UserId,
+
+    /// See [state::User::account_id]
+    pub account_id: AccountId,
+
+    /// See [state::User::projects]
+    pub projects: Vec<ProjectName>,
```
cloudhead: Shouldn't this be `ProjectId`?

core/src/state.rs
```diff
@@ -154,3 +154,38 @@ impl Org {
...
+
+    /// Set of all projects owned by the user.
+    pub projects: Vec<ProjectName>,
+}
+
+impl User {
+    pub fn add_project(mut self, project_name: ProjectName) -> User {
+        if !self.projects.contains(&project_name) {
+            self.projects.push(project_name);
+        }
```
cloudhead: Guess it would be better if we used a `HashSet` for projects, if we're checking before we `push`.

core/src/user_id.rs
```diff
@@ -0,0 +1,215 @@
...
+        if input.contains("--") {
+            return Err(InvalidUserIdError(
+                "must not have more than one consecutive '-'",
+            ));
+        }
+
+        let id = Self(input);
+
+        Ok(id)
+    }
```
cloudhead: Could be a good idea to put this code in one place and share it between org and user id. We then only need to test it once.

runtime/src/registry.rs
```diff
@@ -189,6 +193,24 @@ decl_module! {
...
+            }
+
+            let new_user = state::User {
+                account_id: sender,
+                projects: Vec::new(),
+            };
+            store::Users::insert(message.user_id.clone(), new_user);
+            Self::deposit_event(Event::UserRegistered(message.user_id));
+            Ok(())
+        }
```
cloudhead: We should be checking here that there are no other users registered with that `account_id`.

core/src/state.rs
NunoAlexandre: Substrate `store` doesn't support it.  It'd also be much more expensive to store a HashSet.

runtime/src/registry.rs
xla: I don't think this is a valid english sentence. It should be `identical` or `equal`.

runtime/tests/user_registration.rs
xla: It's only used in this file so far. Would move it if we need it in more places.

core/src/user_id.rs
xla: The consolidation of org and user concerns is planned for a follow-up to not extend the scope of this PR too much.

runtime/src/registry.rs
xla: Yeah, it's a an open point that needs to be addressed.

runtime/src/registry.rs
NunoAlexandre: Yeah, you might be right. What it tries to say is that `input == store::Orgs::decode_key · store::Orgs::storage_map_final_key $ input`.

cli/src/main.rs
```diff
@@ -69,6 +69,7 @@ enum Command {
     UnregisterOrg(UnregisterOrg),
     RegisterProject(RegisterProject),
     RegisterUser(RegisterUser),
+    UnreigsterUser(UnregisterUser),
```
NunoAlexandre:
```suggestion
    UnregisterUser(UnregisterUser),
```

runtime/src/registry.rs
```diff
@@ -189,6 +194,55 @@ decl_module! {
...
+            let sender = ensure_signed(origin)?;
+
+            if store::Users::get(message.user_id.clone()).is_some() {
+                return Err(RegistryError::DuplicateUserId.into())
+            }
+
+            // TODO(xla): This is a naive first version of the check to see if an account is
+            // already associated to a user. While fine for small dataset this needs to be reworked
+            // in the future.
+            for user in store::Users::iter() {
```
xla: This is a naive implementation of what we want to achieve, I'd like to get consensus if we are okay with it for now and how we plan to address mid-term? @cloudhead @NunoAlexandre @geigerzaehler

cloudhead: This is pretty sensible to me.

NunoAlexandre: @xla what you can do is to have a `impl store::Users` that does this in a dedicated function. See if it's helpful. If this is really short lived it's fine as is 👍

xla: What are you proposing to move into a dedicated impl?

NunoAlexandre: The check of whether there is already a user with the given account id so that you don't have to think about that implementation detail while following this handler.

geigerzaehler: An alternative solution would be to add an `AccountUser` storage.
~~~
    pub AccountUser: map hasher(twox_256) AccountId => Option<UserId>;
}
~~~
This would limit the transaction time complexity to `log(#users)`. The downside is that we need to ensure consistency when creating users. Nevertheless, I’d prefer this method.

xla: That was my first approach, but @cloudhead mentioned that we are more concerned about on chain space. I'd like to have consensus what we are optimising for at this stage.

NunoAlexandre: I'd prefer to optimize storage than computation, especially since that storage will be a difficult thing to handle in terms of forwards compatibility. So if this can be done computationally, I'd opt for that.

cloudhead: Is @geigerzaehler 's solution stored in the merkleized state trie (ie. "on chain")? If not, then it seems like a good solution.

xla: AFAIK it is and more so anyone given they know of an account can inspect the state to test if there is an associated user. Is this something we should be concerned about?

geigerzaehler: Good point regarding forwards compatibility. We might figure out a way to cache this information which would not be a breaking change if we start with this implementation. Let’s go with it!

geigerzaehler: > Is @geigerzaehler 's solution stored in the merkleized state trie (ie. "on chain")? If **not**, then it seems like a good solution.

The state is stored on chain.

> anyone given they know of an account can inspect the state to test if there is an associated user. Is this something we should be concerned about?

This is also true without storing it explicitly: Anybody can just iterate over the users. This is nothing we need to be concerned about.

xla: Great, will leave the todo and created #260.

cli/src/commands.rs
```diff
@@ -243,6 +243,64 @@ impl CommandT for RegisterProject {
     }
 }
 
+#[derive(StructOpt, Debug, Clone)]
+/// Register a user.
+pub struct RegisterUser {
+    /// Id of the user to register.
```
geigerzaehler: Since this comment is also displayed as the help text for this option it would be great if we explain what a valid user ID looks like.

client/examples/user_registration.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use radicle_registry_client::{ed25519, message, Client, ClientT, Error, UserId};
+
+#[async_std::main]
+async fn main() {
+    env_logger::init();
+    let mut runtime = tokio::runtime::Runtime::new().unwrap();
+
+    runtime.block_on(Compat::new(go().boxed())).unwrap();
+    runtime.shutdown_now().compat().await.unwrap();
```
geigerzaehler: If you are creating a client with `Client::create_with_executor()` we don’t need this tokio runtime.

client/tests/end_to_end.rs
```diff
@@ -113,6 +113,48 @@ async fn register_org() {
...
 
+#[async_std::test]
+#[serial]
+async fn register_user() {
+    let _ = env_logger::try_init();
+    let node_host = url::Host::parse("127.0.0.1").unwrap();
+    let client = Client::create_with_executor(node_host).await.unwrap();
+    // Must be distinct sender from other user registrations in tests to avoid
+    // AccountUserAssociated errors.
+    let sender = ed25519::Pair::from_string("//alice", None).unwrap();
```
geigerzaehler:
```suggestion
    let sender = ed25519::Pair::from_string("//Alice", None).unwrap();
```
This works now but is going to fail once we introduce fees and “alice” does not have any funds.

core/src/user_id.rs
```diff
@@ -0,0 +1,215 @@
...
+/// https://github.com/radicle-dev/registry-spec/blob/master/body.tex#L110
+use alloc::prelude::v1::*;
+use core::convert::{From, Into, TryFrom};
+use parity_scale_codec as codec;
+
+#[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+pub struct UserId(String);
+
+impl UserId {
+    fn from_string(input: String) -> Result<Self, InvalidUserIdError> {
```
geigerzaehler: This code is now duplicated three times. (The other places are `OrgId` and `ProjectId`.) Could you create an issue to address this?

core/src/user_id.rs
```diff
@@ -0,0 +1,215 @@
...
+impl From<&'static str> for InvalidUserIdError {
+    #[cfg(feature = "std")]
+    fn from(s: &'static str) -> Self {
+        Self(s)
+    }
+
+    #[cfg(not(feature = "std"))]
+    fn from(s: &'static str) -> Self {
+        InvalidUserIdError(s)
+    }
```
geigerzaehler: These methods look the same to me.

test-utils/src/lib.rs
```diff
@@ -94,6 +94,18 @@ pub fn random_project_name() -> ProjectName {
     ProjectName::try_from(random_string(size).to_lowercase()).unwrap()
 }
 
+fn random_user_id() -> UserId {
+    let size = rand::thread_rng().gen_range(1, 33);
+    UserId::try_from(random_string(size).to_lowercase()).unwrap()
```
geigerzaehler: Could you add add a comment that `random_string()` only produces alphanumeric characters and there are always valid user IDs? Or we could rename `random_string` to `random_alnum_string`.

core/src/user_id.rs
xla: There is already a todo item for this in #250 - lemme know if that is insufficient.

PR #251, opened 2020-03-11
Improve logging output

geigerzaehler:
We replace substrate’s default logger with our own. The goal is to make the logs easily readable for development purposes.

* We only show the local time and omit the date.
* We log to stdout instead of stderr since there is no other stdout output.
* We format the non-message part of the logs in grey to simplify reading.

### DIFF
--- a/node/src/logger.rs
+++ b/node/src/logger.rs
@@ -1,0 +1,53 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Provides [init] to initialize our custom logger.
+use env_logger::fmt::Color;
+use std::io::Write as _;
+
+/// Initializes [env_logger] using the `RUST_LOG` environment variables and our custom formatter.
+pub fn init() {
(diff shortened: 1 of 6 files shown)


node/src/logger.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut builder = env_logger::Builder::from_default_env();
+
+    builder.format(|formatter, record| {
+        let time = time::OffsetDateTime::now_local();
+        let context = format!(
+            "{time}.{ms:03} {level:<5} {target}",
+            time = time.format("%H:%M:%S"),
+            ms = time.millisecond(),
+            target = record.target(),
+            level = record.level(),
```
NunoAlexandre: This is really nice 👌

node/src/logger.rs
```diff
@@ -0,0 +1,33 @@
...
+        );
+        writeln!(
+            formatter,
+            "{context}  {msg}",
+            context = formatter
+                .style()
+                .set_color(Color::Black)
+                // Results in gray
```
NunoAlexandre: Don't get the comment?

node/src/logger.rs
```diff
@@ -0,0 +1,33 @@
+use env_logger::fmt::Color;
+use std::io::Write as _;
+
+/// Initializes [env_logger] using the `RUST_LOG` environment variables and our custom formatter.
+pub fn init() {
+    let mut builder = env_logger::Builder::from_default_env();
+
+    builder.format(|formatter, record| {
+        let time = time::OffsetDateTime::now_local();
```
NunoAlexandre: nit: move the formatting part to a separate function that you'd then pass here or use in the closure:

```suggestion
        builder.format(OurCustomFormatter);
```

node/src/logger.rs
```diff
@@ -1,33 +1,38 @@
+//! Provides [init] to initialize our custom logger.
 use env_logger::fmt::Color;
 use std::io::Write as _;
 
 /// Initializes [env_logger] using the `RUST_LOG` environment variables and our custom formatter.
 pub fn init() {
-    let mut builder = env_logger::Builder::from_default_env();
+    env_logger::Builder::from_default_env()
```
NunoAlexandre: beautiful 💯

PR #254, opened 2020-03-12
Log main error instead of just printing

geigerzaehler:


geigerzaehler:
@NunoAlexandre: Could you take a look at this?

node/src/main.rs
```diff
@@ -30,7 +30,7 @@ mod pow;
 
 pub use sc_cli::{error, VersionInfo};
 
-fn main() -> Result<(), error::Error> {
```
NunoAlexandre: Why not keep this? Or why was it there in the first place?

geigerzaehler: If we keep this the Rust runtime will print this error as is to stdout. This messes up the log formatting and will make it difficult for logging agents to properly parse this. There’s no need to keep it since we’re logging the error message with the `log` facility now.

NunoAlexandre: Cool! 👍

node/src/main.rs
```diff
@@ -44,5 +44,11 @@ fn main() -> Result<(), error::Error> {
         copyright_start_year: 2019,
     };
 
-    command::run(version)
+    match command::run(version) {
+        Ok(_) => (),
+        Err(error) => {
+            log::error!("{}", error);
+            std::process::exit(1);
```
CodeSandwich: Does `exit` flush the logger output?

PR #255, opened 2020-03-12
Feature/transaction fees v2

NunoAlexandre:
Draft // Work in Progress.

### What

This new solution proposal introduces `PayTxFee`, an implementation of `SignedExtension` which is made part of `SignedExtra` in the place of the default `ChargeTransactionPayment`. `

PayTxFee`, running as a pre-dispatch validation step, attempts to charge the tx fees from the correct account (tx author or involved org) and sets the tx priority according to the tip provided by the tx author. It rejects the transaction when the payer account does not have enough funds to cover the proposed bid.

This solution checks all the requirements:

✅  R0: Transactions must be validated pre-dispatch, where the transaction author is verified to have enough funds to cover the respective transaction fees.

✅ R1: Allow tx authors to specify the total fee they are willing to pay.
This bid should cover all internal fees and the remainder be used as a tip to get the tx priority.

✅ R2: Be able to control who pays the tx fee
For org-related transactions, we want the org’s account to be the one paying for the transaction.

✅ R3: Only charge base_fee and tip
In the future, we might include other sub fees, such as weights, but for now, we go-ahead with a
base_fee and tip.

✅ R4: From the withdrawn amount from the transaction payee, burn only a small percentage. (todo but seamlessly supported and already in place)

✅ R5: Transactions must only run if the fee can be withdrawn successfully

### Changes

- runtime: Add `PayTxFee` impl of `SignedExtension`.
- runtime: Add payment pipeline to `runtime::fees::payment`
- client: Add `fee` field to `TransactionExtra`
- cli: Add `fee` field to `CommandContext`
- tests: Grant enough funds to accounts used in the tests to pass the tx fee payment step

### Discussion

- [x] 1. 💬On https://github.com/radicle-dev/radicle-registry/pull/243#discussion_r391606253, extracting `members.contains` is not helpful since we need to have the `org` in the first place to get its `account_id`. I suggest that we leave this bit as it is and once we have projects both under `User` and `Org` then we think better how to abstract it.

- [x] 2. ❓Could we `Bid::new` earlier, at the phase of building a tx? I think so since `Bid` lives in the runtime. That would grant us that a tx with an insufficient bid doesn't even enter the system at all. 

- [x] 3. 💬Now, for unauthorized transactions involving orgs, we charge the bid to the tx author but we admit the transaction. It would be easier on the network to discard them right away, but, at the same time, I don't think it should be `TxPayFee` taking that decision. We can leave it as is and create a separate issue to think about this matter. Same as comment 1, I'd prefer to wait until we have `User` and `Orgs` as project owners to think wider about this.

- [x] 4. 💬I prefer us to address the TODO around rewarding the block author in a separate PR.


runtime/src/fees/payment.rs
```diff
@@ -13,20 +13,91 @@
...
+            RegistryCall::transfer_from_org(m) => TxFeePayer::Org(m.org_id),
+            RegistryCall::set_checkpoint(m) => TxFeePayer::Org(m.org_id),
+
+            // Transactions paid by the author
+            RegistryCall::create_checkpoint(_) => TxFeePayer::Author,
+            RegistryCall::register_org(_) => TxFeePayer::Author,
+            RegistryCall::transfer(_) => TxFeePayer::Author,
+
+            // Match arm required by the compiler.
+            crate::registry::Call::__PhantomItem(_, _) => TxFeePayer::Author,
```
geigerzaehler:
```suggestion
            // The runtime does not admit this call. It is required to satisfy the compiler.
            RegistryCall::__PhantomItem(_, _) =>
                panic!("Impossible registry::Call::__PhantomItem variant."),
```

runtime/src/registry.rs
```diff
@@ -280,6 +280,18 @@ decl_module! {
             ));
             Ok(())
         }
+
+        #[weight = SimpleDispatchInfo::InsecureFreeNormal]
+        pub fn transfer(origin, message: message::Transfer) -> DispatchResult {
```
geigerzaehler: If we replicate this transaction here I believe we should disallow any external call to the balances module. We don’t need to fix this here but creating an issue would be nice.

runtime/src/lib.rs
```diff
@@ -226,7 +228,7 @@ pub type SignedExtra = (
     frame_system::CheckEra<Runtime>,
     frame_system::CheckNonce<Runtime>,
     frame_system::CheckWeight<Runtime>,
-    pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
```
geigerzaehler: Could you also remove `pallet_transaction_payment` from `Cargo.toml`?

test-utils/src/lib.rs
```diff
@@ -127,3 +151,30 @@ pub fn random_string(size: usize) -> String {
         .take(size)
         .collect::<String>()
 }
+
+pub fn random_balance() -> Balance {
+    rand::thread_rng().gen_range(20, 100)
+}
+
+pub async fn grant_funds(
```
geigerzaehler: This function is just a thin wrapper around a money transfer that ensures that it has worked. I believe this was always the purpose of `submit_ok`. I’d prefer to change `submit_ok` that it ensures that `tx_applied.result` is `Ok` and call `submit_ok` instead of `grant_funds`.

test-utils/src/lib.rs
```diff
@@ -29,9 +29,10 @@ pub async fn submit_ok<Message_: Message>(
     client: &Client,
     author: &ed25519::Pair,
     message: Message_,
+    bid: Balance,
```
geigerzaehler: We almost always call `submit_ok` with a random balance. Maybe it would make sense to not require the bid argument. If we need it we can either call `sign_and_submit_message` directly or introduce `submit_ok_with_bid`.

runtime/src/fees/payment.rs
NunoAlexandre: Thanks. Where do you get this knowledge from?

runtime/src/registry.rs
NunoAlexandre: Created #257

test-utils/src/lib.rs
NunoAlexandre: It's just about increasing readability: `Grant funds` sounds better and clearer than "submit a transfer message". I can change `submit_ok` to have that check but I'd rather keep this fn.

test-utils/src/lib.rs
NunoAlexandre: yeah, I was thinking the same actually. Will do.

NunoAlexandre:
> How do you feel about calling this just the “fee”? We’re using the terms “bid” and “fee” for basically the same thing and this might lead to some confusion.

To me, especially given all the background work and learning on the subject, "bid" and "fee" are different things. I see "bid" as an amount specified by an author to cover for the ~~costs~~ fees imposed by the network. Looking at their meaning in the dictionary, they don't seem to different: 

>  Bid
> _noun_
> an offer of a price, especially at an auction.

> Fee
> _noun_
> an amount of money paid for a particular piece of work or for a particular right or service


Now, I'd still find confusing that a tx author submits a fee, where the fees are defined by us, in the runtime. It'd be "A user submits a fee to pay for the fees". 🤷‍♂ 

@cloudhead what do you think?

runtime/src/lib.rs
NunoAlexandre: Yes, done.

cli/src/main.rs
```diff
@@ -35,6 +35,17 @@ struct Args {
     )]
     author_key: ed25519::Pair,
 
+    /// Bid to cover all tx fees and gain priority.
+    /// See [radicle_registry_runtime::fees::Bid] for more information.
```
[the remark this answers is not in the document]
NunoAlexandre: @geigerzaehler replied at https://github.com/radicle-dev/radicle-registry/pull/255#issuecomment-598712632.

runtime/src/fees/payment.rs
geigerzaehler: I can’t remember really. I think I picked it up while exploring the substrate codebase.

test-utils/src/lib.rs
geigerzaehler: If you adjust `submit_ok` than `grant_funds` just looks like this:
~~~rust
pub async fn grant_funds(
    client: &Client,
    donator: &ed25519::Pair,
    recipient: AccountId,
    value: Balance,
) {
    submit_ok(
        &client,
        &donator,
        message::Transfer {
            recipient,
            balance: value,
        },
        1,
    )
}
~~~
In my opinion this crates more confusion. We now basically have two ways to transfer funds and it’s not clear which one we should use. Also what is the difference between “granting” funds and “transferring” them?.

NunoAlexandre:
@geigerzaehler 

> * Does it make sense to use the tip as the priority? We not use the total value?

As for the last point, it makes as much sense to use the `tip` as the priority to use the total bid. Both cut it the same way since the fees are the same for everyone. Since the beginning we agreed to use `bid - sum [fees...]` to set the priority, but again, it's the same in the end.

> Are we sure that we need Bid at all? If we replace all occurrences of Bid with the Balance the code becomes much more readable and there’s is less indirection:
> 
> * We can check the minimum fee in validate() instead of Bid::new().
> * We can hard code the withdraw reason which makes more sense because it does not actually 
> * depend on the value of the bid and does not need to be a member.


I don't get this part:

> and does not need to be a member.

> We can check the minimum fee in validate() instead of Bid::new().

What advantage do you see in that?  I still like to have the Bid type since it follows the [Parse, Don't Validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) idea. If we have a `Bid` value, we _know_ it is a valid bid, we can encapsulate what it makes it so (being >= than all mandatory fees), we can calculate its tip (maybe no longer necessary), can test what withdraw reasons it states, etc.

By just having `Balance`, we not only have to validate and keep working with `Balance` but we leak what it makes a valid bid in a place that should only care about _if_ it's valid, now _what_/_how_ it's validated.

> * We can hard code the withdraw reason which makes more sense because it does not actually 

I am fine with doing this since we only have one `Bid` type, although I still see value in keeping `fn withdraw_reasons()` since it's self documentation as to for what reasons a Bid should be withdrawn, instead of - again - leaking and hardcoding that in a withdraw function.

geigerzaehler:
> Now, I'd still find confusing that a tx author submits a fee, where the fees are defined by us, in the runtime. It'd be "A user submits a fee to pay for the fees". man_shrugging

I think this sentence would be much clearer as “A users submits the the fee he is willing to pay”. This is also the terminology used by [XRPL](https://xrpl.org/transaction-common-fields.html).

> Bid
> _noun_
> an offer of a price, especially at an auction.

This might be potentially confusing for users because a bid might also be an offer in a negotiation where you don’t need to pay the full bid in the end.

cloudhead:
> > How do you feel about calling this just the “fee”? We’re using the terms “bid” and “fee” for basically the same thing and this might lead to some confusion.
> @cloudhead what do you think?

I would opt for `fee` simply because that is the common term introduced by bitcoin, and well understood:

From the original whitepaper:
> The incentive can also be funded with transaction fees. If the output value of a transaction is
less than its input value, the difference is a transaction fee that is added to the incentive value of
the block containing the transaction.

The way to rationalize it for me is that with a blockchain, the bid and the fee end up being the same exact thing, due to the nature of transactions: your bid *becomes* the fee as soon as the tx is in a block, and it is then forever a fee.

cloudhead:
> Now, I'd still find confusing that a tx author submits a fee, where the fees are defined by us, in the runtime. It'd be "A user submits a fee to pay for the fees". man_shrugging

This is not really correct, besides the `base_fee` though: the user actually defines most of the fee, not the runtime.

geigerzaehler:
Here are the e2e test cases we need to cover

* For org messages
  - User is authorized
    + not enough org funds => fails
    + enough org funds and tx ok => amount is withdrawn
    + enough org funds and tx fails => amount is withdrawn
  - User is not authorized
    + not enough user funds => fails
    + enough user funds => amount is withdrawn
* For other messages
  + not enough user funds => fails
  + enough user funds => amount is withdrawn
* For all messages
  + rejected if not minium fee

client/Cargo.toml
```diff
@@ -34,10 +34,6 @@ rev = "09abd3b436ea568a47ec4fc47f933728d8cf466b"
 git = "https://github.com/paritytech/substrate"
 rev = "09abd3b436ea568a47ec4fc47f933728d8cf466b"
 
-[dependencies.pallet-transaction-payment]
```
geigerzaehler: Please also remove this dependency from the runtime crate.

test-utils/src/lib.rs
geigerzaehler: Could you please take another look at this comment?

runtime/tests/project_registration.rs
```diff
@@ -27,8 +27,18 @@ async fn register_project() {
     let register_org = random_register_org_message();
     submit_ok(&client, &alice, register_org.clone()).await;
 
+    // The org needs funds to submit transactions.
```
geigerzaehler: I don’t think it’s worth repeating this comment all the time.

runtime/src/fees/mod.rs
```diff
@@ -17,62 +17,60 @@
...
 
-pub mod bid;
 pub mod payment;
 
-pub trait Fee {
-    /// The associated [crate::Balance].
-    fn value(&self) -> Balance;
+/// The base fee serves as a disincentive to stop bad actors
+/// from spamming the network in a DoS attack.
+const BASE_FEE: Balance = 1;
```
geigerzaehler: Maybe “minimum fee” would be a more approriate term?

client/tests/end_to_end.rs
```diff
@@ -140,3 +168,43 @@ async fn invalid_transaction() {
...
+    let client = Client::create_with_executor(node_host).await.unwrap();
+    let tx_author = key_pair_from_string("Alice");
+    let insufficient_fee: Balance = 0;
+
+    let whatever_message = random_register_org_message();
+    let tx_applied = client
+        .sign_and_submit_message(&tx_author, whatever_message, insufficient_fee).await;
+
+    assert!(
+        tx_applied.is_err(),
```
geigerzaehler: Is it possible to check against a concrete error variant? This would prevent false positives when the tx fails for some other reason.

NunoAlexandre: That's what I was reporting. I don't think there is any useful variant to check against that would prevent that but I will have another look.

runtime/src/fees/mod.rs
NunoAlexandre: Yeah, indeed. 👍

runtime/tests/project_registration.rs
NunoAlexandre: ok!

test-utils/src/lib.rs
NunoAlexandre: Adjusting `submit_ok` breaks a big number of calls, so I'd rather not do that in this PR. 

I don't see how this creates any more confusion that `submit_ok` being a wrapper around `sign_and_submit` does, or `key_pair_from_string` as a wrapper of `ed25519::Pair::from_string`. They are all wrapper that creates cleaner calls. 

Would renaming `grant_funds` to `transfer` or `transfer_funds` help?

runtime/tests/project_registration.rs
NunoAlexandre: Actually, I disagree, @geigerzaehler. Most of the tests right now are not documented and have a lot of things happening without much explanation. It would be really unclear _why_ we are transferring funds from an account to another out of the blue when it's unrelated to the test case. Having these comments suffices _why_ and I rather keep them.

client/tests/end_to_end.rs
NunoAlexandre: The best we can do is match on `client::error::Error::Other`. Did that, but yeah, sub-optimal anyways.

test-utils/src/lib.rs
geigerzaehler: > Would renaming `grant_funds` to `transfer` or `transfer_funds` help?

Yes, I think this would prevent some confusion.

client/tests/end_to_end.rs
geigerzaehler: I was just wondering. If it’s not helpful I’m ok with this approach, too.

NunoAlexandre:
@geigerzaehler It's ready 😎 

I have also added an entry to the changelog reporting the new `fee` field in both the client and the cli api.

PR #261, opened 2020-03-16
Also expose WebSocket RPC with CLI flag

geigerzaehler:
The `--unsafe-rpc-external` flag now also exposes the WebSocket RPC API publicly. This is in line with the documentation for the CLI flag.

The idea is that we want the HTTP and WebSocket APIs to behave the same.


geigerzaehler:
> Was there a specific trigger to do this or was it just a quick thing to do that caught your eye?

Yes. External access to our devnet through WebSockets stopped working.

PR #266, opened 2020-03-18
Fix failing low transfers

### LINKED ISSUE #244: Can't transfer low enough funds
I ran into learning that `Transfer` fails when the value being transferred is low enough. So far, the lowest amount I've seen being accepted as 1000.

NunoAlexandre:
Fixes #244 

Change the 'ExistencialDeposit' parameter passed when defining the
runtime. Test that we can transfer any amount within a reasonable range
to cover ourselves from re-introducing this bug again.

### Discussion

It's a shame that `ExistencialDeposit` is used both as the minimum amount required to keep an account open and as a minimum transfer amount. I set it from 500 to 1 but maybe something in between would be better to keep the accounts cleaner.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -169,7 +169,10 @@ impl pallet_timestamp::Trait for Runtime {
 }
 
 parameter_types! {
-    pub const ExistentialDeposit: u128 = 500;
+    /// The minimum amount required to keep an account open.
+    /// Transfers leaving the recipient with less than this
+    /// value fail.
+    pub const ExistentialDeposit: u128 = 1;
     pub const TransferFee: u128 = 0;
     pub const CreationFee: u128 = 0;
     pub const TransactionBaseFee: u128 = 0;
--- a/runtime/tests/transfer.rs
+++ b/runtime/tests/transfer.rs
@@ -25,6 +25,33 @@ async fn transfer_fail() {
     assert!(tx_applied.result.is_err());
 }
 
+// Test that we can transfer any amount within a reasonable range.
+// Affected by the [crate::ExistentialDeposit] parameter.
+#[async_std::test]
+async fn transfer_any_amount() {
+    let client = Client::new_emulator();
+    let donator = key_pair_from_string("Alice");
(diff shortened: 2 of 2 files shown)


runtime/src/lib.rs
```diff
@@ -160,7 +160,9 @@ impl pallet_timestamp::Trait for Runtime {
 }
 
 parameter_types! {
-    pub const ExistentialDeposit: u128 = 500;
+    /// The minimum amount required to keep an account open.
+    /// It also defines the minimum allowed transfer.
```
geigerzaehler: I don’t think it is true that this is the minimum transfer amount. I had a look at the code and this is not checked anywhere.

NunoAlexandre: Then it does play a role indirectly because I could effectively move the minimum allowed transfer by changing this value. 🤔

NunoAlexandre: @geigerzaehler found it:

https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L892

In an older version of the substrate repo I have locally, in the `fn transfer` in the `ballances` it was doing:

``` rust
if would_create && value < T::ExistentialDeposit::get() {
  Err(Error::<T, I>::ExistentialDeposit)?
}
```

Now looking at the latest master, the only close thing I can find is the following, which is different:

https://github.com/paritytech/substrate/blob/dd97b1478b31a4715df7e88a5ebc6664425fb6c6/frame/balances/src/lib.rs#L966-L967


Maybe this was a bug now fixed. Will rebase and see.

NunoAlexandre: @geigerzaehler I now got the problem. So, it's not that this value is used as a minimum transfer amount per se, but say that when transferring `x ` amount from account A to account B, if doing so leaves `B` with fewer funds than this `ExistencialDeposit` amount, then it fails. I verified this by _first_ transferring an amount above the `ExistencialDeposit` value and then below, which works, but not the way around. Rephrased the comment.

geigerzaehler: Yes, that sounds correct.

PR #267, opened 2020-03-18
Feature/new cli commands

NunoAlexandre:
These changes come from the needs faced while preparing the tx fees demo.


cli/src/commands.rs
```diff
@@ -95,26 +99,62 @@ pub struct ShowProject {
...
+            .client
+            .get_org(self.org_id.clone())
+            .await?
+            .ok_or(CommandError::OrgNotFound {
+                org_id: self.org_id.clone(),
+            })?;
+
+        println!("id: {}", org.id.clone());
+        println!("account_id: {}", org.account_id.clone());
+        println!("members: {:?}", org.members.clone());
```
CodeSandwich: Do we need these `clone`s?

cli/src/commands.rs
```diff
@@ -426,3 +466,26 @@ impl CommandT for ShowBalance {
         Ok(())
     }
 }
+
+#[derive(StructOpt, Debug, Clone)]
+/// Generate a new account with the given name, i.e,
+/// generate a key pair for the given name and return
+/// the publick key. Used for development purposes only.
+pub struct GenerateAccount {
```
CodeSandwich: I think that we should include `Dev`, `ForTest` or something similar in the name to make it obviously dev-only tool. It's too easy and too dangerous to miss that detail.

At some point we probably will want to expose a generator for regular users with password and stuff and we will end up with a blocked good name.

Or we could add a `--dev`, `--test-account` or some similar flag and for now require having it set. Then in the future we can add means of passing a password and stop requiring this flag.

cli/src/commands.rs
```diff
@@ -426,3 +466,26 @@ impl CommandT for ShowBalance {
...
+    name: String,
+}
+
+#[async_trait::async_trait]
+impl CommandT for GenerateAccount {
+    async fn run(&self, _command_context: &CommandContext) -> Result<(), CommandError> {
+        let key_pair =
+            ed25519::Pair::from_string(format!("//{}", self.name).as_str(), None).unwrap();
+        println!(
+            "Generated account with name \"{}\" and public key {}",
```
CodeSandwich: Wouldn't it be sometimes useful for the devs to get a private key?

cli/src/main.rs
```diff
@@ -109,8 +113,10 @@ async fn run(args: Args) -> Result<(), CommandError> {
         Command::UnregisterUser(cmd) => cmd.run(&command_context).await,
         Command::ShowBalance(cmd) => cmd.run(&command_context).await,
         Command::ShowGenesisHash(cmd) => cmd.run(&command_context).await,
+        Command::ShowOrg(cmd) => cmd.run(&command_context).await,
         Command::ShowProject(cmd) => cmd.run(&command_context).await,
         Command::Transfer(cmd) => cmd.run(&command_context).await,
         Command::TransferOrgFunds(cmd) => cmd.run(&command_context).await,
+        Command::GenerateAccount(cmd) => cmd.run(&command_context).await,
```
CodeSandwich: Probably a bad place to discuss that: don't we need a command hierarchy? We have `list-orgs`, `show-org`, `register-org` and `unregister-org`, same with `user`, similar with `project` and the list is growing.

As a user I would prefer having subcommands, e.g. `org list`, `org show`, `org register` and `org unregister`, this way it would be easier for me to discover commands by having list of general fields of interest in top-level `--help` and list of topic-specific actions in `org --help`.

If you agree or at least don't oppose, I'll turn it into a GitHub issue.

NunoAlexandre: That's something I have started to think about given the work for the Wallet CLI, indeed. Feel free to create an issue, although it's something I am already looking at.

CodeSandwich: Awesome! Here's the issue: https://github.com/radicle-dev/radicle-registry/issues/268

cli/src/commands.rs
NunoAlexandre: I _think_ the build would fail otherwise.

cli/src/commands.rs
NunoAlexandre: `GenerateTestAccount`? I am also fine with making this a rust script that we can run with `cargo run`, whatever provides the functionality really. Your concern is valid indeed, and this doesn't need to be a cli command per se, so open for other suggestions!

geigerzaehler: I wouldn’t worry too much about adding `Dev` or `ForTest`. A lot of what is in the CLI is dev related and if we make this available to users we’ll review the commands.

However we should clarify the documentation and terminology a bit. This command does not quite generate an new account from a name. It prints the account ID for the key pair that is deterministicly derived from the input string. The key derivation mechanism is specified in detail [here](https://substrate.dev/rustdocs/v1.0/substrate_primitives/crypto/trait.Pair.html#method.from_string). It is also worth pointing out that the input string is the same as for the [`--author-key` option](https://github.com/radicle-dev/radicle-registry/blob/9a751cbc2941ce0ead6195b118c150b017d63c8a/cli/src/main.rs#L29-L36).

If the purpose of this command is to be get an account ID to be able to show the balance of an account we could also consider making the account ID argument for `show-balance` optional and using the `--author-key` value instead.

NunoAlexandre: > If the purpose of this command is to be get an account ID to be able to show the balance of an account we could also consider making the account ID argument for show-balance optional and using the --author-key value instead.

This is also a good idea. The need for me to have a public key to play around was also to `transfer` funds. There we can't follow that approach. However, since the pub key is deterministically generated from the seed, we could change such commands that receive an `SS58` address format param to receive a seed string instead and we parse it into a public key, aka `AccountId`. That would be more user friendly since the entities involved are named. 

``` bash
cargo run -p radicle-registry-cli -- transfer 5CRrGJDs8SKWwnuXxKjVazS8oK7oWe2EcYg2mrzXQQJxtAb4 10
``` 

would become

``` bash
cargo run -p radicle-registry-cli -- transfer bob 10
```

cli/src/commands.rs
NunoAlexandre: Yeah, indeed. I didn't have that need for the demo but it would be useful nonetheless. I can change it to be `GenerateKeyPair` and have it print both keys.

cli/src/commands.rs
geigerzaehler: > However, since the pub key is deterministically generated from the seed, we could change such commands that receive an `SS58` address format param to receive a seed string instead and we parse it into a public key, aka `AccountId`

This would mean that those commands are only usable if you know the _private_ key of the recipient which seems very limiting.

I believe having this command makes sense. We should just improve on the naming and documentation. What do you think about the following:

````
show-address <SEED_STRING>

Shows the SS58 address and hex encoded account ID for the key pair derived from SEED_STRING. See https://substrate.dev/rustdocs/v1.0/substrate_primitives/crypto/trait.Pair.html#method.from_string for more information on how the seed string is interpreted.
```

It would be great if the command would output both the SS58 address and the hex account ID so we can use the flexibly.

NunoAlexandre: > This would mean that those commands are only usable if you know the private key of the recipient which seems very limiting.

hm how is that? 🤔 

> I believe having this command makes sense. We should just improve on the naming and documentation. What do you think about the following:

Seems good to me! I have some questions though:

1. For the [`Wallet CLI`](https://github.com/radicle-dev/radicle-registry/issues/275), one of the use cases is "As a user, I can print an account address to use it for mining". Would this command be it then?

2. I am not sure I understand the concept of "address". For the use case above, I thought the address referred to an IP address but now it seems related to this. Is an address in this context the combination SS58 address and hex-encoded account ID?

3. What is the difference between the SS58 address (public key, thus account id (?)) and the hex encoded account ID?

geigerzaehler: > > This would mean that those commands are only usable if you know the private key of the recipient which seems very limiting.
> 
> hm how is that?

The seed string, like “Alice”, is used to derive the _private_ key. This means that anybody how has the seed string also has the private key and this is something you definitely don’t want to give away.

> 1. For the [`Wallet CLI`](https://github.com/radicle-dev/radicle-registry/issues/275), one of the use cases is "As a user, I can print an account address to use it for mining". Would this command be it then?

Yes.




> 2. I am not sure I understand the concept of "address". For the use case above, I thought the address referred to an IP address but now it seems related to this. Is an address in this context the combination SS58 address and hex-encoded account ID?
> 3. What is the difference between the SS58 address (public key, thus account id (?)) and the hex encoded account ID?

The account ID is the 32 bytes of the public key. An [SS58 address](https://substrate.dev/docs/en/overview/ss58-address-format) is a way of encoding the 32 bytes into a string with a checksum. Check out the [API documentation](https://substrate.dev/rustdocs/master/sp_core/crypto/trait.Ss58Codec.html). In the context of crypto currencies an [address](https://en.bitcoin.it/wiki/Address) is usually something you need to uniquely identify the recipient of a transaction. Both the account ID and the SS58 address are addresses in that sense.

NunoAlexandre: > The seed string, like “Alice”, is used to derive the private key. This means that anybody how has the seed string also has the private key and this is something you definitely don’t want to give away.

Ah! I thought the `seed` was used to deterministically derive the `public` key. So that's for sure something we don't want to give away. Should have checked the docs.

> The account ID is the 32 bytes of the public key. An SS58 address is a way of encoding the 32 bytes into a string with a checksum. Check out the API documentation. In the context of crypto currencies an address is usually something you need to uniquely identify the recipient of a transaction. Both the account ID and the SS58 address are addresses in that sense.

Thank you very much 👌

cli/src/commands.rs
NunoAlexandre: So, following the discussion above, we agreed to rename this to `ShowAddress`. Generating a key pair is indeed useful but I will address that separately as part of #275, since that's one of its use cases.

README.md
```diff
@@ -118,6 +118,13 @@ use radicle-registry-client::{ed25519, CryptoPair};
 let alice = ed25519::Pair::from_string("//Alice", None);
 ```
 
+Alternatively, you can use `radicle-registry-cli` to generate
+you a public key given an account name:
+
+``` bash
+cargo run -p radicle-registry-cli -- generate-account <name>
```
geigerzaehler: Could you update this too?

NunoAlexandre: Done 👍

Issue #258, opened 2020-03-13
Split `registry.rs` module by its concerns


NunoAlexandre:
It's getting cumbersome and difficult to go through. We should think of how to better split it into separate modules based on different concerns.

xla:
For clarification are you referring to `runtime/registry.rs`?

NunoAlexandre:
Yes! The only `registry.rs` we have in this project. Will adjust the title anyway.

PR #270, opened 2020-03-19
Implement block rewards

geigerzaehler:
We add a block author inherent to the runtime and credit the author with block rewards and transaction fees.

We don’t implement setting the block author from the CLI yet.

The code enforces some invariants with `panic!`s and `assert!`s. This follows common substrate practice. (See [`pallet_timestamp`][1] for example.)

[1]: https://github.com/paritytech/substrate/blob/1ea615c0a422fdbc7ecdbc7357bdf7f5bbdc421b/frame/timestamp/src/lib.rs#L149

### DIFF
--- a/node/src/inherent_data.rs
+++ b/node/src/inherent_data.rs
@@ -1,0 +1,31 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Provides [InherentDataProviders] for the registry [InherentData].
+use radicle_registry_runtime::registry::InherentData;
+use radicle_registry_runtime::AccountId;
+use sp_inherents::InherentDataProviders;
+
+/// Return [InherentDataProviders] that provides [InherentData] for registry blocks required by
+/// full nodes.
+pub fn new_full_providers() -> InherentDataProviders {
+    let providers = InherentDataProviders::new();
+    let data = InherentData {
+        block_author: AccountId::from_raw([0u8; 32]),
+    };
+    // Can only fail if a provider with the same name is already registered.
+    providers.register_provider(data).unwrap();
+    providers
+}
--- a/runtime/src/fees/payment.rs
+++ b/runtime/src/fees/payment.rs
@@ -17,7 +17,7 @@ use crate::registry::store;
 use crate::{AccountId, Call, DispatchError, RegistryCall};
 use radicle_registry_core::*;
 
-use frame_support::storage::StorageMap as _;
+use frame_support::storage::{StorageMap as _, StorageValue as _};
 use frame_support::traits::{Currency, ExistenceRequirement, WithdrawReason};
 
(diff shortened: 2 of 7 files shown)


CodeSandwich:
Were you able to test these changes? Shouldn't existing e2e tests be affected?

runtime/src/registry.rs
```diff
@@ -376,6 +476,21 @@ decl_module! {
                 ExistenceRequirement::KeepAlive
             )
         }
+
+        #[weight = SimpleDispatchInfo::FixedOperational(10_000)]
+        fn set_block_author(origin, author: AccountId) -> DispatchResult {
```
CodeSandwich: Is this function ever called? Where?

geigerzaehler: It is called when the `Call::set_block_author` inherent is dispatched. This inherent is created as part of the `ProvideInherent` implementation of `Module`.

geigerzaehler:
> Were you able to test these changes? Shouldn't existing e2e tests be affected?

Good point. I was able to test this manually but this is obviously not enough. I’ll create an issue to follow up on this and write some tests.

node/src/inherent_data.rs
```diff
@@ -0,0 +1,31 @@
...
+use sp_inherents::InherentDataProviders;
+
+/// Return [InherentDataProviders] that provides [InherentData] for registry blocks required by
+/// full nodes.
+pub fn new_full_providers() -> InherentDataProviders {
+    let providers = InherentDataProviders::new();
+    let data = InherentData {
+        block_author: AccountId::from_raw([0u8; 32]),
+    };
+    // Can only fail if a provider with the same name is already.
```
NunoAlexandre: the sentence seems unfinished.

runtime/src/fees/payment.rs
```diff
@@ -27,7 +27,15 @@ type NegativeImbalance = <crate::Balances as Currency<AccountId>>::NegativeImbal
...
 pub fn pay(author: AccountId, fee: Balance, call: &Call) -> Result<(), DispatchError> {
     let payer = payer_account(author, call);
-    let _withdrawn_fee = withdraw(fee, &payer)?;
+    let withdrawn_fee = withdraw(fee, &payer)?;
+
+    // The block author is only available when this function is run as part of the block execution.
+    // If this function is run as part of transaction validation the block author is not set. In
+    // that case we don’t need to credit the block author.
+    if let Some(block_author) = store::BlockAuthor::get() {
+        crate::Balances::resolve_creating(&block_author, withdrawn_fee);
```
NunoAlexandre: We should burn a percentage of the `withdrawn_fee`. Are we doing that in a separate PR? I recall you mentioning a substrate type/way of doing it?

runtime/src/registry.rs
```diff
@@ -376,6 +476,21 @@ decl_module! {
                 ExistenceRequirement::KeepAlive
             )
         }
+
+        #[weight = SimpleDispatchInfo::FixedOperational(10_000)]
```
NunoAlexandre: Do we really want this weight here or it is copy-pasted?

runtime/src/registry.rs
```diff
@@ -127,6 +136,98 @@ fn descends_from_initial_checkpoint(
     false
 }
 
+const INHERENT_IDENTIFIER: InherentIdentifier = *b"registry";
```
NunoAlexandre: Would it be an idea to already have this stuff in a separate module, given the [discussion around splitting this module](#258)?

runtime/src/registry.rs
geigerzaehler: We need _some_ weight. The value is not really important at the moment but it is definitely an operational dispatch.

runtime/src/fees/payment.rs
geigerzaehler: Good point. I’d like to postpone this. It is a bit more involved than I thought and I’d like to write some unit tests for this.

NunoAlexandre: sounds good to me! Can you capture that piece in a separate issue then?

PR #272, opened 2020-03-19
Add CLI option for block author

geigerzaehler:
We add the `--block-author` CLI flag to specify the block author for miners.


node/src/cli.rs
```diff
@@ -78,6 +79,13 @@ pub struct Arguments {
     /// If the file does not exist, it is created with a newly generated secret key.
     #[structopt(long, value_name = "FILE")]
     node_key_file: Option<std::path::PathBuf>,
+
+    /// Account to credit block rewards and transaction fees for mined blocks.
+    ///
+    /// If not provided the zero account is used as the block author. Account address must be given
+    /// in SS58 format.
```
NunoAlexandre: nit:

```suggestion
    /// Account to credit block rewards and transaction fees for mined blocks.
    ///
    /// If not provided the zero account is used as the block author. 
    /// Account address must be given in SS58 format.
```

node/src/cli.rs
```diff
@@ -78,6 +79,13 @@ pub struct Arguments {
...
+    #[structopt(long, value_name = "SS58_ADDRESS", parse(try_from_str = parse_ss58_account_id))]
```
NunoAlexandre: 💬  So each node should set the `SS58_ADDRESS`, which is it's own (Block Author) account. Makes sense 👍

node/src/cli.rs
```diff
@@ -129,6 +137,11 @@ impl Arguments {
             ..run_cmd
         }
     }
+
+    pub fn block_author(&self) -> AccountId {
+        self.block_author
+            .unwrap_or_else(|| AccountId::from_raw([0; 32]))
+    }
```
NunoAlexandre: suggestion:

```suggestion
        let account_zero = AccountId::from_raw([0; 32]);
        self.block_author.unwrap_or(account_zero)
    }
```

I'd like to name the account zero and maybe even have it aside and explained, if you see the value. I happen to know hat it is from a conversation with Alexis, wouldn't have a clue otherwise.

And no need to `unwrap_or_else(|| ...)`, `unwrap_or` is sufficient.

node/src/cli.rs
```diff
@@ -143,3 +156,7 @@ fn parse_chain(name: &str) -> Result<Chain, String> {
         Err(format!("Invalid chain {}", name))
     }
 }
+
+fn parse_ss58_account_id(data: &str) -> Result<AccountId, String> {
+    sp_core::crypto::Ss58Codec::from_ss58check(data).map_err(|err| format!("{:?}", err))
+}
```
NunoAlexandre: `suggestion`:

Move this `fn` to `core` and dry it here and in the `cli::commands.rs`, `parse_account_id`.

node/src/command.rs
```diff
@@ -28,6 +28,9 @@ pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
     let chain_spec = args.chain.spec();
     let spec_factory = |_: &str| Ok(Box::new(chain_spec) as Box<_>);
 
+    let block_author = args.block_author();
+    let new_full_service = move |config| service::new_full(config, block_author);
```
NunoAlexandre: `suggestion`: 

For me, this could just be the following, since `block_author` is only useful for `new_full`. 

```suggestion
    let new_full_service = move |config| service::new_full(config, args.block_author());
```

geigerzaehler: This does not work unfortunately. With your suggestion `args` will be moved to the closure and the closure will take ownership. However `args` is used later.

node/src/cli.rs
geigerzaehler: This code is very specific to CLI parsing and very simple. I don’t think `core` is the right place for this. We should avoid treating the `core` package as a collection of all common code. Usually this code would be inlined but to make it work with `structopt` we need to declare it as a function.

NunoAlexandre: Yeah, I am fine with that and I agree with your points about not turning `core` into a classic `utils`.

node/src/command.rs
NunoAlexandre: Ah, right. You would need to `args.clone().block_author()`, I guess? yeah, let's not do that then.

node/src/service.rs
```diff
@@ -191,5 +193,16 @@ pub fn new_light(config: Configuration) -> Result<impl AbstractService, ServiceE
...
+    let inherent_data_providers = InherentDataProviders::new();
+    Ok(new_full_start!(config, inherent_data_providers).0)
+}
+
+/// Return [InherentDataProviders] that provides [InherentData] for registry blocks required by
+/// full nodes.
+fn new_full_inherent_data_providers(block_author: AccountId) -> InherentDataProviders {
+    let providers = InherentDataProviders::new();
+    let data = InherentData { block_author };
+    // Can only fail if a provider with the same name is already.
```
NunoAlexandre: This comment is wrong but it's fixed in the other PR, for the record.

NunoAlexandre:
@geigerzaehler since we have the cli cli and the node cli, could you prefix your commit with `node: `?

PR #273, opened 2020-03-19
Build docs and check links on CI

### LINKED ISSUE #192: Check documentation on CI
CI builds should fail if rust [doc links](https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html) (like `/// [core::fmt::Display`) cannot be resolved. This would increase the quality of the documentation and make it less likely that we forget to update documentation when things change.

I’m not sure how to achieve this but we need to deny the `intra_doc_link_resolution_failure` lint.

geigerzaehler:
We now check that the intra doc links in the code docs are all valid. You can see that it works from the first, failing commit.

Fixes #192

### DIFF
--- a/client/src/backend/emulator.rs
+++ b/client/src/backend/emulator.rs
@@ -29,12 +29,13 @@ use crate::interface::*;
 /// [backend::Backend] implementation using native runtime code and in memory state through
 /// [sp_io::TestExternalities] to emulate the ledger.
 ///
-/// # Differences with real [Client]
+/// # Differences with real backend
 ///
-/// The [Emulator] does not produce blocks. In particular the `blocks` field in
-/// [TransactionApplied]` always equals `Hash::default` when returned from [Client::submit].
+/// * The [Emulator] does not produce blocks. In particular the `blocks` field in
+///   [TransactionApplied]` always equals `Hash::default` when returned from
+///   [ClientT::submit_transaction].
 ///
-/// The responses returned from the client never result in an [Error].
+/// * The responses returned from the client never result in an [Error].
 #[derive(Clone)]
 pub struct Emulator {
     test_ext: Arc<Mutex<sp_io::TestExternalities>>,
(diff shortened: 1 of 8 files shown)


client/src/backend/emulator.rs
```diff
@@ -29,10 +29,11 @@ use crate::interface::*;
 /// [backend::Backend] implementation using native runtime code and in memory state through
 /// [sp_io::TestExternalities] to emulate the ledger.
 ///
-/// # Differences with real [Client]
+/// # Differences with real backend
 ///
 /// The [Emulator] does not produce blocks. In particular the `blocks` field in
-/// [TransactionApplied]` always equals `Hash::default` when returned from [Client::submit].
+/// [TransactionApplied]` always equals `Hash::default` when returned from
```
NunoAlexandre: Should we make these differences bullet points to make it clear at any point that it's an enumeration?

CodeSandwich:
> How are you approaching this?

I just wrote a script running all the non-`cargo test` checks and use it whenever I need to `¯\_(ツ)_/¯` It's too slow for a git hook.

NunoAlexandre:
> It's too slow for a git hook.

Indeed. It's also time-consuming to have the ci running for a while to only then learn that a `clone` is unnecessary, but probably not as much. I wish these two things could be either verified at the time of writing or only verified at the time of merging, to avoid spending so much time during development where I'd rather have my mind focused on more important things.

geigerzaehler:
@NunoAlexandre I hear your concern. There’s a lot of things to get right to make a PR pass. Code editors should, with the appropriate configuration, be able to support checking the formatting, compilation, and lints. The rust language servers only use clippy by default if there is a `!#[warn(clippy::all)]` or similar attribute in the code. But it is possible to configure the language server to allways check for clippy lints I believe. It would also be ok to add the attribute to all crates.

I don’t know of a good solution for the documentation issue. But I believe that is important to have that check to ensure that the documentation stays meaningful.

PR #276, opened 2020-03-23
Burn part of tx fees

geigerzaehler:
We burn a 1% of transaction fees. We also add tests for the `pay` function.

### DIFF
--- a/runtime/src/fees/payment.rs
+++ b/runtime/src/fees/payment.rs
@@ -18,9 +18,13 @@ use crate::{AccountId, Call, DispatchError, RegistryCall};
 use radicle_registry_core::*;
 
 use frame_support::storage::{StorageMap as _, StorageValue as _};
-use frame_support::traits::{Currency, ExistenceRequirement, WithdrawReason};
+use frame_support::traits::{Currency, ExistenceRequirement, Imbalance, WithdrawReason};
+use sp_runtime::Permill;
 
 type NegativeImbalance = <crate::Balances as Currency<AccountId>>::NegativeImbalance;
+
+/// Share of a transaction fee that is burned rather than credited to the block author.
+const BURN_SHARE: Permill = Permill::from_percent(1);
 
 /// Pay Fees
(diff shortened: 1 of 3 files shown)


runtime/src/fees/payment.rs
```diff
@@ -18,22 +18,28 @@ use crate::{AccountId, Call, DispatchError, RegistryCall};
...
+const BURN_SHARE: Permill = Permill::from_percent(1);
+
 /// Pay Fees
 /// Given a tx author, their fee, and a RegistryCall they are submitting,
 /// charge the tx fees to the right account, which depends on the `registry_call`.
 pub fn pay(author: AccountId, fee: Balance, call: &Call) -> Result<(), DispatchError> {
     let payer = payer_account(author, call);
     let withdrawn_fee = withdraw(fee, &payer)?;
+    let (burn, reward) = withdrawn_fee.split(BURN_SHARE * fee);
+    drop(burn);
```
NunoAlexandre: TIL

runtime/src/fees/payment.rs
```diff
@@ -95,3 +101,50 @@ fn org_payer_account(author: AccountId, org_id: &OrgId) -> AccountId {
...
+#[cfg(test)]
+mod test {
+    use core::convert::TryFrom;
+    use frame_support::traits::Currency;
+    use sp_core::{crypto::Pair, ed25519};
+    use sp_runtime::BuildStorage;
+
+    use crate::{Balances, GenesisConfig};
+
+    use super::*;
```
NunoAlexandre: `nit`

I recall us talking to first have the local imports and then the libraries. Doing as suggested below also makes these imports a bit more compact in terms of vertical space.

```suggestion
    use super::*;
    use crate::{Balances, GenesisConfig};
    
    use core::convert::TryFrom;
    use frame_support::traits::Currency;
    use sp_core::{crypto::Pair, ed25519};
    use sp_runtime::BuildStorage;    
```

PR #278, opened 2020-03-23
Simplify runtime types

geigerzaehler:
We eliminate unnecessary runtime types

### DIFF
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -30,19 +30,17 @@ use std::sync::Arc;
 use url::Url;
 
 use radicle_registry_runtime::{
-    opaque::Block as OpaqueBlock, BlockNumber, Event, EventRecord, Hash, Hashing, Header,
+    Block, BlockNumber, Event, EventRecord, Hash, Hashing, Header, UncheckedExtrinsic,
 };
 
 use crate::backend::{self, Backend};
 use crate::interface::*;
 
-type ChainBlock = SignedBlock<OpaqueBlock>;
-
 /// Collection of substrate RPC clients
 #[derive(Clone)]
 struct Rpc {
     state: StateClient<BlockHash>,
-    chain: ChainClient<BlockNumber, Hash, Header, ChainBlock>,
+    chain: ChainClient<BlockNumber, Hash, Header, SignedBlock<TransportBlock>>,
     author: AuthorClient<Hash, BlockHash>,
 }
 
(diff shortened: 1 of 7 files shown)


CodeSandwich:
> Looks good to me but I would like to understand what changed to make this comment no longer the case:

These are types exactly the same as the ones defined below the `opaque` module. They are opaque too.

geigerzaehler:
This is blocked by https://github.com/paritytech/substrate/issues/5359

geigerzaehler:
Could you take another look at this? Since there’s nothing happening on the upstream issue I created a small fix for this.

client/src/backend/remote_node.rs
```diff
@@ -239,3 +237,36 @@ fn extract_transaction_events(
...
+        D: serde::Deserializer<'a>,
+    {
+        let bytes: Vec<u8> = Vec::<u8>::deserialize(de)?;
+        let uncheck_extrinsic = Decode::decode(&mut &bytes[..])
+            .map_err(|e| serde::de::Error::custom(format!("Decode error: {}", e)))?;
+        Ok(TransportExtrinsic(uncheck_extrinsic))
+    }
+}
+
+fn unwrap_transport_block(transport_block: TransportBlock) -> Block {
```
CodeSandwich, line 266: This name suggests possibility of panicking, which is not the case. It also took me quite some time to resolve all the type aliases and figure out that this function just unpacks `Block`'s `UncheckedExtrinsics` from `TransportExtrinsics`. Maybe it's just me or maybe it's not so obvious and needs some documentation?

client/src/backend/remote_node.rs
```diff
@@ -143,11 +141,11 @@ impl RemoteNode {
...
-        let block = opt_signed_block
-            .ok_or_else(|| {
-                Error::from("Block that should include submitted transaction does not exist")
-            })?
-            .block;
+        let opt_block =
+            opt_signed_block.map(|signed_block| unwrap_transport_block(signed_block.block));
+        let block = opt_block.ok_or_else(|| {
+            Error::from("Block that should include submitted transaction does not exist")
+        })?;
```
CodeSandwich, line 148: Maybe `unwrap_transport_block` could be renamed to `signed_block_to_block` and accept a signed block? Then this could clear up:
```suggestion
        let block = opt_signed_block.map(signed_block_to_block)
            .ok_or("Block that should include submitted transaction does not exist")?;
```

client/src/backend/remote_node.rs
geigerzaehler: Good point. I’ve renamed the function to `unpack_transport_block` and added some documentation.

client/src/backend/remote_node.rs
geigerzaehler: I’d like to keep `unpack_transport_block` because it makes the one-to-one-correspondence between `TransportBlock` and `Block` more obvious and it clearly separates the workaround for the serialization from the actual logic.

CodeSandwich: That's fair

Issue #268, opened 2020-03-19
Create CLI command hierarchy


CodeSandwich:
### Problem
We have too many long, specific CLI commands. E.g. for `-org` we have `list-orgs`, `show-org`, `register-org` and `unregister-org`, it's the same with `-user` and similar with `-project`. The list is growing quickly and so does the user confusion.

### Solution
As a user I would prefer having subcommands, e.g. `org list`, `org show`, `org register` and `org unregister`. This way it would be easier for me to discover commands by having a list of general topics of interest in a top-level `--help` and a list of topic-specific actions in `org --help`.

NunoAlexandre:
@CodeSandwich thanks for starting this discussion. As part of the [Wallet CLI](#275) work, this was something I started to look into.

I have a few alternatives I'd like us to consider, where yours is one of them :)

### Commands

First, let's enumerate the commands we have atm:

``` rust
enum Command {
    ListOrgs(ListOrgs),
    ListProjects(ListProjects),
    RegisterOrg(RegisterOrg),
    UnregisterOrg(UnregisterOrg),
    RegisterProject(RegisterProject),
    RegisterUser(RegisterUser),
    UnregisterUser(UnregisterUser),
    ShowBalance(ShowBalance),
    ShowGenesisHash(ShowGenesisHash),
    ShowOrg(ShowOrg),
    ShowProject(ShowProject),
    Transfer(Transfer),
    TransferOrgFunds(TransferOrgFunds),
    ShowAddress(ShowAddress),
}
```

## Alternative A - Organized by domains

Organize the commands by their domain. `Org`-related commands go under `org`, `Project`-related commands go under `project`, and so on.

Following this alternative, the commands would look like this:

``` bash
> cli org list
> cli org register
> cli org unregister
> cli org show
> cli org transfer

> cli project list
> cli project register
> cli project show

> cli user register
> cli user unregister

> cli account address
> cli account transfer
> cli account balance

> cli genesis show
> 
```

### Pros

+ Natural categorization
+ Very readable


## Alternative B - CQRS-inspired

Some of the _commands_ translate to transactions, others are just _queries_.
 

``` bash

> cli submit org-register
> cli submit org-unregister
> cli submit org transfer

> cli query org-list
> cli query org-show

> cli query project-list
> cli query project-show

> cli submit project-register

> cli submit user register
> cli submit user unregister

> cli query address
> cli query balance
> cli submit transfer

> cli query genesis-show
```

### Pros

[...]

PR #279, opened 2020-03-23
Improve CLI structure

NunoAlexandre:
Fixes https://github.com/radicle-dev/radicle-registry/issues/268

### Changes

0. Group and clean commands by domain (#268)  

   So that the list of commands becomes:  

      ``` bash 
       > cli account address
       > cli account transfer
       > cli account balance

       > cli genesis hash       

       > cli org list
       > cli org register
       > cli org unregister
       > cli org show
       > cli org transfer

       > cli project list
       > cli project register
       > cli project show

       > cli user register
       > cli user unregister
      ```

1. Split executable and library components

      Introduce a `lib.rs` and have the `main.rs` only as an entry point. 
    Doing so better separates concerns and makes the CLI easier to navigate. 

2. Split `commands.rs` by domain

      Instead of having all the commands structs and implementations in this single module,  
    split this module by domain, where we have a dedicated module per domain.  

    ``` bash  
    cli
    ├── Cargo.toml
    └── src
        └── main.rs
        ├── lib.rs
        ├── command
        │   ├── mod.rs
        │   ├── account.rs
        │   ├── genesis.rs
        │   ├── org.rs
        │   ├── project.rs
        │   └── user.rs
    ```

3. Implement `CommandT` for all command types

    Implement CommandT for all command types, from top to bottom in the
    hierarchy. Doing so allows us to delegate by cascading, instead of
    having a massive match at the top level. Another benefict is that it
    provides a perfect separation between adding high-level and
    domain-specific commands. For instance, adding a new command related to
    org means only changing command::org, nothing else.
   
4. Refactor and rename Args as CommandLine plus further renaming

    What we now call 'Args' really is the _model_ of our command-line. I was
    initially confused with this type, thinking it would capture the
    arguments of commands. This change makes it all clear through naming but
    also through structure, where it's clear that our command-line is made
    of a command and options. Furthermore, have the now 'CommandLine' offer
    a `run` function for improved encapsulation, where `main.rs` really only
    has to connect the dots and know nothing about the internals. At last,
    improve other naming and reduce unnecessary verbosity.

### Discussion

Two things I'd like to consider and discuss

1. `cli genesis hash` and `cli account address` could be moved into a [`flatten`](https://docs.rs/structopt/0.3.12/structopt/#flattening-subcommands) Command case `Misc/Other/etc` if we don't find the new structure useful for these two commands. They would then be `cli show-genesis` and `cli show-address` again.

2. now `--help` only shows the high-level commands, not the nested ones. One can always run `--help` on a specific subcommand but I'd like to see all right away.


cli/src/commands.rs
```diff
@@ -13,82 +13,14 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Defines [CommandT] trait, structs for all commands and their [CommandT] implementations.
+//! Defines all the [CommandT] implementations.
```
geigerzaehler: This file is not needed anymore.

NunoAlexandre: Oh, I must have re-introduced it while cleaning the git history I had locally. Thanks

geigerzaehler:
> 1. `cli genesis hash` and `cli account address` could be moved into a [`flatten`](https://docs.rs/structopt/0.3.12/structopt/#flattening-subcommands) Command case `Misc/Other/etc` if we don't find the new structure useful for these two commands. They would then be `cli show-genesis` and `cli show-address` again.

`account address` is fittingly grouped under `account`. As for the other command I’d prefer to have it just as `cli genesis-hash`. Having groups with only one item makes things slight less discoverable.

NunoAlexandre:
> Having groups with only one item makes things slight less discoverable.

yeah, I agree. If we'd foresee more commands genesis-related in the pipeline I'd prefer to keep it but in this case, I'd rather have it flatten.

cli/src/lib.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub command: Command,
+
+    #[structopt(flatten)]
+    pub options: CommandLineOptions,
+}
+
+impl CommandLine {
+    pub async fn run(self) -> Result<(), CommandError> {
+        let command = self.command.clone();
+        let context = CommandContext::new(self.options.clone()).await?;
```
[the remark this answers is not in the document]
NunoAlexandre: Sick!

cli/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+/// The commands are grouped by domain.
+#[derive(StructOpt, Debug, Clone)]
+pub enum Command {
+    Account(account::Command),
+    Org(org::Command),
+    Project(project::Command),
+    User(user::Command),
+
+    #[structopt(flatten)]
+    Other(other::Command),
```
geigerzaehler: Suggestion: To remove a bit of cruft you could just use `ShowGenesisHash` here. This would eliminate the `other::Command` enum and you could also move `ShowGenesisHash` to its own module. It would make the command easier to find.

cli/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+#[async_trait::async_trait]
+pub trait CommandT {
+    async fn run(&self, ctx: &CommandContext) -> Result<(), CommandError>;
+}
+
+/// Error returned by [CommandT::run].
+///
+/// Implements [From] for client errors.
+#[derive(Debug, derive_more::From)]
+pub enum CommandError {
```
CodeSandwich: This enum would probably benefit from `ThisError`

cli/src/lib.rs
NunoAlexandre: Thanks for the suggestion. I see your point but I prefer to keep this structure to already have the setup in place to add new commands that don't fit within any specific domain. Shall we find issues finding the command in future work we can always reconsider.

cli/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+/// Context for running commands.
+/// Created from [CommandLineOptions].
+pub struct CommandContext {
+    pub tx_author: ed25519::Pair,
+    pub client: Client,
+    pub tx_fee: Balance,
+}
```
CodeSandwich: Do we need these options in every single command? Why are we ask users for them when they are unnecessary?

cli/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl CommandLineOptions {
+    fn parse_key_pair(s: &str) -> Result<ed25519::Pair, String> {
+        ed25519::Pair::from_string(s, None).map_err(|err| format!("{:?}", err))
+    }
+}
+
+/// The supported [CommandLine] commands.
+/// The commands are grouped by domain.
+#[derive(StructOpt, Debug, Clone)]
```
CodeSandwich: `Clone` is probably unnecessary, but it's not a big deal

cli/src/lib.rs
NunoAlexandre: We can address that as part of the error-handling improvement work we have planned 👍

cli/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+            tx_fee: options.tx_fee,
+        })
+    }
+}
+
+/// The trait that every command must implement.
+#[async_trait::async_trait]
+pub trait CommandT {
+    async fn run(&self, ctx: &CommandContext) -> Result<(), CommandError>;
+}
```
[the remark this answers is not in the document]
NunoAlexandre: This trait already existed before. Answering your question: to have a common interface for all commands. So far we haven't had the need to have a less rigid setup, but it's something to consider shall we need it. For now, I think the trait is sensible. Feel free to create a follow-up issue if you would like discussing this further now.

cli/src/lib.rs
NunoAlexandre: Note that this already was the way it is before, I just moved them to the new `CommandLineOptions` type. Regarding your question, the answer is no. That's part of the reason why I thought of grouping the commands by `query`, `transaction`, etc. This does need some work since the overall setup leads to things like #281

cli/src/lib.rs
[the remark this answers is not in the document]
CodeSandwich: It seems that this trait is already too rigid, it forces us to request users for many unnecessary parameters only to fill `CommandContext` with them.

cli/src/lib.rs
CodeSandwich: Ok, I'll just add this point to #281

cli/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: I wouldn't look at it that way. Most likely, the first commands added to the cli were transactions, which did require the context to run. So it made sense to have that in. Things changed in the meantime and ignoring the context-param wasn't a clear show-stopper. That's all fine, now that we have everything more organized we have the room clean to better spot and think of these improvements.

PR #280, opened 2020-03-23
Validate transaction call type

### LINKED ISSUE #250: Consolidate orgs and users as top-level entities
Following up on #245 and to complete radicle-dev/registry-spec#11 we need to consolidate orgs and uesrs and support the unified registration of projects under either as well as guarantee a shared namespace for their ids.

- [x] consolidate id namespace for orgs and users
- [ ] support project registration under both orgs and users https://github.com/radicle-dev/radicle-registry/issues/389

### LINKED ISSUE #269: Validate inherent extrinsics
At the moment nodes allow inherent calls in signed extrinsics. For example a user can submit a signed extrinsic with a `set_timestamp` call to a node. The node will accept that extrinsic, broadcast it, and it will eventually be included in a block. Executing that intrinsic will have no effect but it waste space, time, and bandwidth. The problem also applies to other inherents that we will add, like block rewards.

To address this we should add code that ensures that certain calls are only valid for unsigned extrinsics. This validation needs to be run by the tx pool and the block importer.

geigerzaehler:
To fix #269 and fix #250 we forbid certain extrinsic calls from entering the transaction pool and being executed.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -30,13 +30,15 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 extern crate alloc;
 
 use frame_support::weights::Weight;
-use frame_support::{construct_runtime, parameter_types, traits::Randomness};
+use frame_support::{construct_runtime, ensure, fail, parameter_types, traits::Randomness};
 use sp_core::{ed25519, OpaqueMetadata};
 use sp_runtime::traits::{BlakeTwo256, Block as BlockT};
 use sp_runtime::{
-    create_runtime_str, generic, transaction_validity::TransactionValidity, ApplyExtrinsicResult,
-    Perbill,
+    create_runtime_str, generic,
+    transaction_validity::{InvalidTransaction, TransactionValidity},
+    ApplyExtrinsicResult, Perbill,
 };
+
 use sp_std::prelude::*;
 
 use sp_api::impl_runtime_apis;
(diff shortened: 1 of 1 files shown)


NunoAlexandre:
@geigerzaehler this approach looks good, although - before seeing code - I'd expect us to to do this using a signed extension where this would run at pre-dispatch. Did you think of that? Would it be a good idea, or advantageous?

geigerzaehler:
> I'd expect us to to do this using a signed extension where this would run at pre-dispatch

I’ve considered this but it requires substantially more machinery and is less transparent. We would need to create an empty `struct` just for the purpose of implementing `SignedExtension`. We would need to add that struct to `SignedExtra` and also add it to `transaction_extra_to_runtime_extra` in the client. Moreover `SignedExtension` forces us to separate validation of signed and unsigned extrinsics which gives us less flexibility in structuring the code.

Also note that hardcoding the `InvalidTransaction` values in this manner is not ideal. This is something to pick up on when we tackle the errors.

runtime/src/lib.rs
```diff
@@ -243,6 +245,36 @@ pub type Executive = frame_executive::Executive<
     AllModules,
 >;
 
+const SIGNED_INHERENT_CALL_ERROR: InvalidTransaction = InvalidTransaction::Custom(1);
+const FOBIDDEN_CALL_ERROR: InvalidTransaction = InvalidTransaction::Custom(2);
+const UNSGINED_CALL_ERROR: InvalidTransaction = InvalidTransaction::Custom(3);
+
+/// Validate that the call of extrinsic is allowed.
```
NunoAlexandre:
```suggestion
/// Validate that the call of the extrinsic is allowed.
```

NunoAlexandre: or

```suggestion
/// Validate that the extrinsic call is allowed.
```

runtime/src/lib.rs
```diff
@@ -243,6 +245,36 @@ pub type Executive = frame_executive::Executive<
...
+///
+/// * We forbid calls reserved for inherents when the extrinsic is not signed.
+/// * We forbid any calls to the [Balances] or [System] module.
+/// * We ensure that the extrinsic is signed for non-inherent calls.
+///
+fn validate_extrinsic_call(xt: &UncheckedExtrinsic) -> Result<(), InvalidTransaction> {
+    match xt.function {
+        // Inherents are only allowed if there unsigned.
```
NunoAlexandre:
```suggestion
        // Inherents are only allowed if they are unsigned.
```

PR #284, opened 2020-03-25
node: Add CLI flag to expose prometheus endpoint externally

geigerzaehler:
Add CLI flag to expose prometheus endpoint externally


xla:
Instead of a boolean flag this should take an interface/address so that operators can control where to expose their telemetry endpoints, i.e. a management network/interface.

geigerzaehler:
> Instead of a boolean flag this should take an interface/address so that operators can control where to expose their telemetry endpoints, i.e. a management network/interface.

Good point. Unfortunately with the current setup we’re restricted to what the substrate CLI provides. Implementing you’re suggestion would require us to somehow gain access to the internal configuration. This might be possible but will require some effort

PR #286, opened 2020-03-26
Clarify the PR process

CodeSandwich:
Resolves PR reviewers' race condition. It also cleans up the "branch" section of `CONTRIBUTING.MD`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,17 +11,25 @@ may not use any other domain-specific abbreviations.
 
 # Git Flow
 
-TL;DR: This repository follows the Git rebase flow for the most part. Each feature branch must
-be reviewed and approved in a pull request. Once ready to merge, consider squashing the branch's
-commits when the separate commits don't add value, rebase it, force push with lease, and merge it
-via the GitHub UI.
+This repository follows the Git rebase flow for the most part.
 
-## Branches
+## Branches and pull requests
 
 1. Create a separate branch for each issue your are working on
 2. Do your magic
-3. Keep your branch up to date by rebasing it from its base branch
-4. Delete the branch after its been both approved and merged. GitHub does this automatically.
+3. Create a pull request and assign reviewers.
+  - Assign the registry team as the reviewer when one review from whoever member reviews first is
+    sufficient. The reviewer who arrives first must overwrite themselves as the reviewer to signal
+    ownership and avoid racing reviews.
+  - Assign multiple reviewers individually when you expect a review from everyone listed.
+4. Keep your branch up to date by rebasing it from its base branch
+5. Make sure that all the assigned reviewers approve the changes
+6. Consider squashing the branch's commits when the separate commits don't add value
+7. Merge it via the GitHub UI
+8. Delete the branch after its been merged. GitHub does this automatically.
+
+If you were asked to review a pull request by being member of a team that was assigned
+as a reviewer, always assign yourself and unassign the team before you start the review.
 
 ## Commits
 


CONTRIBUTING.md
```diff
@@ -11,17 +11,18 @@ may not use any other domain-specific abbreviations.
 
 # Git Flow
 
-TL;DR: This repository follows the Git rebase flow for the most part. Each feature branch must
-be reviewed and approved in a pull request. Once ready to merge, consider squashing the branch's
-commits when the separate commits don't add value, rebase it, force push with lease, and merge it
-via the GitHub UI.
+This repository follows the Git rebase flow for the most part.
 
 ## Branches
```
geigerzaehler: Should we rename this section to “Pull Requests”?

CONTRIBUTING.md
```diff
@@ -11,17 +11,18 @@ may not use any other domain-specific abbreviations.
...
 
 1. Create a separate branch for each issue your are working on
 2. Do your magic
-3. Keep your branch up to date by rebasing it from its base branch
-4. Delete the branch after its been both approved and merged. GitHub does this automatically.
+3. Create a pull request and assign reviewers
```
geigerzaehler: Maybe worth mentioning that you can assign the team to just get one review from somebody.

CONTRIBUTING.md
CodeSandwich: Renamed it to `Branches and pull requests`

CONTRIBUTING.md
CodeSandwich: Fixed

CodeSandwich:
> in the case of setting a team as the reviewer to mean "whoever reviews first", the problem that originated this discussion in the first place remains

That's a good point. In the Grafana PR the whole Radicle team was assigned. To mark that I want to review it I've assigned myself on top of Radicle. I could also remove Radicle team to make it even more clear. Is it a good strategy? If so, I'll write it down in `CONTRIBUTING.md`.

NunoAlexandre:
Yeah, that sounds like a good idea 👍

CodeSandwich:
Fixed

CONTRIBUTING.md
```diff
@@ -11,17 +11,22 @@ may not use any other domain-specific abbreviations.
...
+This repository follows the Git rebase flow for the most part.
 
-## Branches
+## Branches and pull requests
 
 1. Create a separate branch for each issue your are working on
 2. Do your magic
-3. Keep your branch up to date by rebasing it from its base branch
-4. Delete the branch after its been both approved and merged. GitHub does this automatically.
+3. Create a pull request and assign reviewers. Remember that you can assign a whole team,
```
NunoAlexandre:
```suggestion
3. Create a pull request and assign reviewers. 
    Assign the registry team as the reviewer when one review from whoever member reviews first is sufficient (*).
    Assign multiple, individual reviewers when everyone listed is expected to review.
```

CONTRIBUTING.md
```diff
@@ -11,17 +11,22 @@ may not use any other domain-specific abbreviations.
...
+which means that only one review is requested from all its members.
+4. Keep your branch up to date by rebasing it from its base branch
+5. Make sure that all the assigned reviewers approve the changes
+6. Consider squashing the branch's commits when the separate commits don't add value
+7. Merge it via the GitHub UI
+8. Delete the branch after its been merged. GitHub does this automatically.
+
+If you were asked to review a pull request by being member of a team that was assigned
+as a reviewer, always assign yourself and unassign the team before you start the review.
```
NunoAlexandre: Sorry, I find this sentence difficult to understand. How about:

```suggestion
(*) To avoid racing and duplicated reviews, change the team as the reviewer to yourself before you start your review.
```

CONTRIBUTING.md
CodeSandwich: For me this version has more convoluted structure and is harder to read. I don't have strong feelings though, so if you're convinced that your version is better, I'll include it.

CONTRIBUTING.md
NunoAlexandre: Actually, this should just live next to the place where it's referenced.

```suggestion
```

CONTRIBUTING.md
NunoAlexandre: I wouldn't have suggested it otherwise. Anyway, I have rewards some parts and included the part that was at the end of the file here to avoid the unnecessary lookup.

NunoAlexandre: btw @CodeSandwich, do we get a notification when one overwrites the selected PR reviewer?

CONTRIBUTING.md
CodeSandwich: Applied

CONTRIBUTING.md
CodeSandwich: Fixed.

I don't know, why is it important?

NunoAlexandre: Because it's the whole point of this setup: avoid duplicated/racing reviews. If two people open the PR and change the reviewer without noticing the other did the same, the problem still exists. A notification would decrease the damage. But let's go ahead like this and see how it goes in practice.

CodeSandwich: If two people try to change the reviewer of the same PR, GitHub will serve as a synchronous source of truth. You'll see that something is wrong when removal of `Registry` fails and not only you are a reviewer, but also somebody else. But the probability of such clash is very low.

NunoAlexandre: Yeah 👍

PR #287, opened 2020-03-26
Add Grafana and Prometheus to local devnet

geigerzaehler:
To allow us to play with metrics for the local devnet we add Grafana and Prometheus to our local devnet. We also provision a basic dashboard with some network metrics.

### DIFF
--- a/local-devnet/docker-compose.yaml
+++ b/local-devnet/docker-compose.yaml
@@ -4,7 +4,7 @@ version: "3.7"
 x-node: &node
   image: ubuntu:eoan-20191101
   volumes:
-  - ../target/debug/radicle-registry-node:/usr/local/bin/radicle-registry-node:ro
+  - ../target/release/radicle-registry-node:/usr/local/bin/radicle-registry-node:ro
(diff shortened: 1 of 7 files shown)


local-devnet/docker-compose.yaml
```diff
@@ -4,7 +4,7 @@ version: "3.7"
 x-node: &node
   image: ubuntu:eoan-20191101
   volumes:
-  - ../target/release/radicle-registry-node:/usr/local/bin/radicle-registry-node:ro
+  - ../target/debug/radicle-registry-node:/usr/local/bin/radicle-registry-node:ro
```
CodeSandwich: Now we should also update `scripts/build-dev-node` to build a debug, otherwise `local-devnet/README.md` instructions are invalid. I think that would be an overall good step, because the official way of building would become faster. We probably also need `scripts/build-release-node`.

geigerzaehler: This was just a local change that I’ll revert. The debug build might suffer from performance problems that make it impossible to author blocks. This is why we need the release build.

PR #297, opened 2020-03-27
node: Update version info

geigerzaehler:
We ensure that the node includes the commit hash. This is helpful for identifying the node build with monitoring and through logs.

We also update the support URL and use `ff` (for Friends&Family-Net) for the version.

### DIFF
--- a/node/src/main.rs
+++ b/node/src/main.rs
@@ -28,14 +28,12 @@ mod service;
 fn main() {
     let version = sc_cli::VersionInfo {
         name: "Radicle Registry Node",
-        commit: "<none>",
-        // commit: env!("VERGEN_SHA_SHORT"),
-        // version: env!("CARGO_PKG_VERSION"),
-        version: "unstable",
+        commit: env!("VERGEN_SHA_SHORT"),
+        version: "ff.0",
         executable_name: "radicle-registry",
         author: "Monadic GmbH",
         description: "Radicle Registry Node",
-        support_url: "support.anonymous.an",
+        support_url: "http://github.com/radicle-dev/radicle-registry/issues",
         copyright_start_year: 2019,
     };
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3610,6 +3610,7 @@ dependencies = [
  "time 0.2.9",
  "tokio 0.2.13",
  "trie-root 0.15.2",
+ "vergen",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


node/src/main.rs
```diff
@@ -28,14 +28,12 @@ mod service;
...
-        // commit: env!("VERGEN_SHA_SHORT"),
-        // version: env!("CARGO_PKG_VERSION"),
-        version: "unstable",
+        commit: env!("VERGEN_SHA_SHORT"),
+        version: "ff",
         executable_name: "radicle-registry",
         author: "Monadic GmbH",
         description: "Radicle Registry Node",
-        support_url: "support.anonymous.an",
+        support_url: "http://radicle.community",
```
NunoAlexandre: I remember us talking about creating a "network" section in the radicle community. It'd be better to have this link pointing to it to better guide users.

node/src/main.rs
```diff
@@ -28,14 +28,12 @@ mod service;
 fn main() {
     let version = sc_cli::VersionInfo {
         name: "Radicle Registry Node",
-        commit: "<none>",
-        // commit: env!("VERGEN_SHA_SHORT"),
-        // version: env!("CARGO_PKG_VERSION"),
-        version: "unstable",
+        commit: env!("VERGEN_SHA_SHORT"),
+        version: "ff",
```
NunoAlexandre: Are we sure we won't release any update during the ff net phase? I'd prefer to be safe and start with `registry.ff.0` or something like that.

node/src/main.rs
geigerzaehler: There’s no “network” section yet. Maybe we can link to the radicle channel on freenode?

node/src/main.rs
geigerzaehler: > Are we sure we won't release any update during the ff net phase?

No, we’ll probably release an update. But instead of coming up with a release and versioning scheme right now that might become obsolete I’d rather stick with a simple indicator. `unstable` was just a bit scary.

node/src/main.rs
NunoAlexandre: I know, but maybe we could add it and point the link to it.

node/src/main.rs
NunoAlexandre: yeah! but ok, since we will release an update during ff, we should version it at least that far, right? "ff.0"?

node/src/main.rs
geigerzaehler: This is Abbey’s domain and as far as I know we don’t want to have that section quite yet. With what we have right now, what would be the best place? Maybe Github issues actually?

NunoAlexandre: Yeah maybe 🤔 I just don't want to guide users to a place where they either get lost or will be posting in the wrong place.

geigerzaehler: I’ve changed it to the Github issues URL. If we have a better target we can always update it.

PR #302, opened 2020-03-27
Disable CORS check when --unsafe-rpc-external is enabled

### LINKED ISSUE #301: CORS for external connections to nodes
Currently nodes reject external RPC API connections with a 403 code even with the `--unsafe-external-rpc` option. The reason is that the CORS option for the APIs is not set properly.

geigerzaehler:
We disable the RPC API server side check for connections. Fixes #301.

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -38,9 +38,14 @@ pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
             subcommand.run(config, service::new_for_command)
         }
         None => {
+            let unsafe_rpc_external = args.unsafe_rpc_external;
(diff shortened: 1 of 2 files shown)


node/src/command.rs
```diff
@@ -38,9 +38,14 @@ pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
             subcommand.run(config, service::new_for_command)
```
CodeSandwich: Don't we need to set `unsafe_rpc_external` when we're running a subcommand?

geigerzaehler: No. The “subcommands” are all one-shot commands that run neither the node nor the RPC API.

Issue #275, opened 2020-03-20
Create CLI Wallet


NunoAlexandre:
We want users to interact with the FFnet and send transactions. Miners also will want to verify that they receive block rewards. To this end, we need to provide a simple wallet.

The wallet needs to satisfy the following use cases:

- [ ] C0. As a user, I can generate a key pair to start using an account
- [ ] C1. As a user, I want to manage multiple accounts and identify them by name (It’s not necessary to encrypt the stored keys with passwords)

- [x] C2. As a user, I want to see the balance for an account
        Run `radicle-registry-cli account balance <account_id>`
- [x] C3. As a user, I want to transfer tokens to another account
       Run `radicle-registry-cli account transfer <recipient> <balance>`

- [x] C4. As a user, I can print an account address to use it for mining
        Run `radicle-registry-cli account address <seed>`

- [x]  C5. As a user, I want to use different nodes to interact with the chain
         Set `--node_host` cmd-line option

**note:** For now, `Account` is the name given to the public key of a key pair, which is unique.

NunoAlexandre:
Regarding `C0` and `C1`, that are functionally intertwined, I forsee us offering the following:

- `C0`: `cli account create <unique_name> <seed>`

The original use case says "I want to generate a key pair to start using an account". Generating a key pair and storing it in the local filesystem seems to be unnecessary. Other CLI commands either receive an address (encoded pub key) or the seed from which key pairs can be derived, so we never have the case where we need to specify a key pair directly. Therefore, storing the key pair doesn't seem to add value.

The simplest solution for FFnet would be, as the output of this command, to store the `seed` in a file named after the `unique_name`. It would look something like this:

``` 
.radicle-cli
    ├── accounts
    │   ├── unique_name_a
    │   ├── unique_name_b
    │   ├── unique_name_c
```

Each `unique_name_x` file would only contain the `seed`. Having this, we can fulfil `C1`.

- `C1`: `cli account list`

This command will iterate over the local accounts created at `C0` and print the following info for each:

``` bash
   Account 'unique_name' (file name)
     address: 'abc' (computed)
     seed: 'xyz' (read from the file) (or maybe not? see the *concern* below)
```

This setup grants us both use cases while not storing any more than what we need. 

There is a remaining **concern** though.

[...]

PR #303, opened 2020-03-27
cli: Add account generate and list commands

NunoAlexandre:
Completes the functionality required by #275 

Adds `cli account generate <name>` and `cli account list`.

**Note**:
`cli account generate <name>` generates a random key pair using `ed25519::Pair::generate()`, instead of us generating one from a given seed. I switched to this approach since we will no longer need the users to remember the seed. By doing so, we delegate security and form concerns to `ed25519::Pair::generate()` and offload our users from such unnecessary and critical responsibility.

### DIFF
--- a/cli/src/account_storage.rs
+++ b/cli/src/account_storage.rs
@@ -1,0 +1,112 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Manages accounts stored in the filesystem,
+//! providing ways to store and retrieve them.
+
+use directories::BaseDirs;
+use sp_core::serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fs::File;
+use thiserror::Error as ThisError;
+
+use std::io::Error as IOError;
+use std::path::PathBuf;
+
+/// The data that is stored in the filesystem relative
+/// to an account. The account name is used as the key
+/// to this value, therefore not included here.
+#[derive(Serialize, Deserialize)]
+pub struct AccountData {
+    pub seed: Seed,
+}
+
+/// The seed from which an account's key pair
+/// can be determiniscally generated.
+type Seed = [u8; 32];
+
+#[derive(Debug, ThisError)]
+pub enum Error {
+    /// An Account with the given name already exists
+    #[error("An account with the given name already exists")]
+    AlreadyExists(),
+
+    /// An IOError occurred
+    #[error("An IO error occured: {0}")]
+    IOError(IOError),
+
+    /// A serde json Error occurred
+    #[error("A serde json error occured: {0}")]
+    JsonError(serde_json::Error),
+}
+
+/// Add an account to the storage.
+///
+/// Fails if an account with the given `name` already exists.
+/// It can also fail from IO and Serde Json errors.
+pub fn add(name: String, data: AccountData) -> Result<(), Error> {
+    let mut accounts = list()?;
+    if accounts.contains_key(&name) {
+        return Err(Error::AlreadyExists());
+    }
+
+    accounts.insert(name, data);
+    update(accounts)
+}
+
+/// List all the stored accounts
+///
+/// It can fail from IO and Serde Json errors.
+pub fn list() -> Result<HashMap<String, AccountData>, Error> {
+    let path_buf = get_or_create_path()?;
+    let file = File::open(path_buf.as_path()).map_err(Error::IOError)?;
+    let accounts: HashMap<String, AccountData> =
+        serde_json::from_reader(&file).map_err(Error::JsonError)?;
+    Ok(accounts)
+}
+
+fn update(accounts: HashMap<String, AccountData>) -> Result<(), Error> {
+    let path_buf = get_or_create_path()?;
+    let new_content = serde_json::to_string(&accounts).map_err(Error::JsonError)?;
+    std::fs::write(path_buf.as_path(), new_content.as_bytes()).map_err(Error::IOError)
+}
+
+const FILE: &str = "accounts.json";
+
+// Get the path to the accounts file on disk.
+//
+// If the file does not yet exist, create it and initialize
+// it with an empty object so that it can be deserialized
+// as an empty HashMap<String, AccountData>.
+fn get_or_create_path() -> Result<PathBuf, Error> {
+    let dir = dir()?;
+    let path_buf = dir.join(FILE);
+    let path = path_buf.as_path();
+
+    if !path.exists() {
+        std::fs::write(path, b"{}").map_err(Error::IOError)?
+    }
+
+    Ok(path_buf)
+}
+
+fn dir() -> Result<PathBuf, Error> {
+    let dir = BaseDirs::new()
+        .unwrap()
+        .data_dir()
+        .join("radicle-registry-cli");
+    std::fs::create_dir_all(&dir).map_err(Error::IOError)?;
+    Ok(dir)
+}
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -16,12 +16,15 @@
 //! Define the commands supported by the CLI related to Accounts.
 
 use super::*;
+use crate::account_storage;
 
 /// Account related commands
 #[derive(StructOpt, Debug, Clone)]
 pub enum Command {
     Address(ShowAddress),
     Balance(ShowBalance),
+    Generate(Generate),
+    List(List),
     Transfer(Transfer),
 }
 
@@ -31,16 +34,18 @@ impl CommandT for Command {
         match self {
             Command::Address(cmd) => cmd.run(ctx).await,
             Command::Balance(cmd) => cmd.run(ctx).await,
+            Command::Generate(cmd) => cmd.run(ctx).await,
+            Command::List(cmd) => cmd.run(ctx).await,
             Command::Transfer(cmd) => cmd.run(ctx).await,
         }
     }
 }
 
-#[derive(StructOpt, Debug, Clone)]
 /// Show the SS58 address for the key pair derived from `seed`.
 ///
 /// For more information on how the seed string is interpreted see
 /// <https://substrate.dev/rustdocs/v1.0/substrate_primitives/crypto/trait.Pair.html#method.from_string>.
+#[derive(StructOpt, Debug, Clone)]
 pub struct ShowAddress {
     seed: String,
 }
@@ -55,8 +60,8 @@ impl CommandT for ShowAddress {
     }
 }
 
-#[derive(StructOpt, Debug, Clone)]
 /// Show the balance of an account
+#[derive(StructOpt, Debug, Clone)]
 pub struct ShowBalance {
     #[structopt(
         value_name = "account",
@@ -75,8 +80,47 @@ impl CommandT for ShowBalance {
     }
 }
 
+/// Generate a local account and store it on disk.
+///
+/// Fail if there is already an account with the given `name`.
 #[derive(StructOpt, Debug, Clone)]
+pub struct Generate {
+    /// The name that uniquely identifies the account locally.
+    name: String,
+}
+
+#[async_trait::async_trait]
+impl CommandT for Generate {
+    async fn run(&self, _ctx: &CommandContext) -> Result<(), CommandError> {
+        let (_, seed) = ed25519::Pair::generate();
+        account_storage::add(self.name.clone(), account_storage::AccountData { seed })?;
+        println!("✓ Account generated successfully");
+        Ok(())
+    }
+}
+/// list all the local accounts
+#[derive(StructOpt, Debug, Clone)]
+pub struct List {}
+
+#[async_trait::async_trait]
+impl CommandT for List {
+    async fn run(&self, _ctx: &CommandContext) -> Result<(), CommandError> {
+        let accounts = account_storage::list()?;
+
+        println!("Accounts ({})", accounts.len());
+        for (name, data) in accounts {
+            println!("Account '{}'", name);
+            println!(
+                "\taddress: {}",
+                ed25519::Pair::from_seed(&data.seed).public().to_ss58check()
+            );
+        }
+        Ok(())
+    }
+}
+
 /// Transfer funds to recipient
+#[derive(StructOpt, Debug, Clone)]
 pub struct Transfer {
     #[structopt(parse(try_from_str = parse_account_id))]
     /// Recipient Account in SS58 address format.
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -18,6 +18,8 @@
 use radicle_registry_client::*;
 use structopt::StructOpt;
 
+pub mod account_storage;
+
 mod command;
 use command::{account, org, other, project, user};
 
@@ -147,6 +149,7 @@ pub enum CommandError {
         project_name: ProjectName,
         org_id: OrgId,
     },
+    AccountStorageError(account_storage::Error),
 }
 
 impl core::fmt::Display for CommandError {
@@ -162,6 +165,9 @@ impl core::fmt::Display for CommandError {
                 project_name,
                 org_id,
             } => write!(f, "Cannot find project {}.{}", project_name, org_id),
+            CommandError::AccountStorageError(error) => {
+                write!(f, "Account storage error: {}", error)
+            }
         }
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "blake2b_simd"
+version = "0.5.10"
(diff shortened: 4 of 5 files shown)


cli/src/command/account.rs
```diff
@@ -31,11 +33,38 @@ impl CommandT for Command {
...
+/// Fail if there is already an account with the given `name`
+/// locally.
+pub struct Create {
+    name: String,
+}
+
+#[async_trait::async_trait]
+impl CommandT for Create {
+    async fn run(&self, _ctx: &CommandContext) -> Result<(), CommandError> {
+        let (kp, _) = ed25519::Pair::generate();
```
NunoAlexandre: should do this instead:

```suggestion
        let (, seed) = ed25519::Pair::generate();
```

cli/src/command/account.rs
```diff
@@ -110,3 +161,107 @@ impl CommandT for Transfer {
...
+                name: name.clone(),
+                seed: account.seed,
+            });
+        }
+
+        accounts
+    }
+
+    fn get_file() -> File {
+        //TODO(nuno): find a better and more qualified path.
```
NunoAlexandre: 🤔 not sure where to place this. Suggestions?

cli/src/command/account.rs
```diff
@@ -31,11 +33,38 @@ impl CommandT for Command {
...
             Command::Balance(cmd) => cmd.run(ctx).await,
+            Command::Create(cmd) => cmd.run(ctx).await,
             Command::Transfer(cmd) => cmd.run(ctx).await,
+            Command::List(cmd) => cmd.run(ctx).await,
         }
     }
 }
 
+#[derive(StructOpt, Debug, Clone)]
+/// Create an account identified by `name`.
```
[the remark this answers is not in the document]
geigerzaehler: How about calling the command `generate` and using this documentation.
```suggestion
/// Generate a key pair for an account and store it on disk.
```
You could remove the other two sentences and add them to the `name` argument. Would be much more concise.

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
         Ok(())
     }
 }
+
+/// Accounts Storage
+///
+/// Manages [Account]s stored in the filesystem,
+/// providing ways to store and retrieve them.
+pub mod storage {
```
geigerzaehler: This is a concern that is cross cutting for different commands once they use it too. How about moving it to a separate file, say `src/account_storage.rs`?

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
...
+    use directories::BaseDirs;
+    use serde_json as json;
```
geigerzaehler: Could we avoid renaming crates? This only saves a few characters but a reader always needs to look it up when they see it in the code.

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
...
+        json.insert(account.name.clone(), StoredAccount { seed: account.seed });
+        overwrite(&mut file, &json)?;
+        Ok(())
+    }
+
+    /// List all the stored accounts
+    pub fn list() -> Vec<Account> {
+        let file = get_file();
+        let json: HashMap<String, StoredAccount> =
+            json::from_reader(&file).ok().unwrap_or({ HashMap::new() });
```
geigerzaehler: We’re swallowing errors when the account storage cannot be read. I’d prefer to show fail and show an error to the user.

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
...
+        let mut accounts = Vec::new();
+        for (name, account) in json.iter() {
```
geigerzaehler: Suggestion: Use `json.iter().map(...).collect()`.

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
...
+            .open(path)
+            .expect("Failed to open accounts file")
+    }
+
+    fn overwrite<T: ?Sized>(file: &mut File, value: &T) -> Result<(), Error>
+    where
+        T: Serialize,
+    {
+        file.seek(SeekFrom::Start(0)).map_err(Error::IOError)?;
+        file.set_len(0).map_err(Error::IOError)?;
```
geigerzaehler: How about using [`std::fs::write`](https://doc.rust-lang.org/std/fs/fn.write.html). Then we don’t need to deal with `OpenOptions` and file seeking.

cli/src/command/account.rs
```diff
@@ -55,6 +84,28 @@ impl CommandT for ShowAddress {
     }
 }
 
+#[derive(StructOpt, Debug, Clone)]
+/// Show the balance of an account
```
geigerzaehler: That’s not the right documentation

cli/src/lib.rs
```diff
@@ -162,6 +165,20 @@ impl core::fmt::Display for CommandError {
...
+        }
+    }
+}
+
+use account::storage::Error as AccountStorageError;
+impl std::convert::From<AccountStorageError> for CommandError {
+    fn from(error: AccountStorageError) -> CommandError {
+        match error {
+            AccountStorageError::AlreadyExists() => CommandError::DuplicatedAccount(),
+            AccountStorageError::JsonError(_) => CommandError::BadJson(),
```
geigerzaehler: Is there a reason why the error names are different? Why not use the same name but prefix it with `Account`?

cli/src/lib.rs
```diff
@@ -162,6 +165,20 @@ impl core::fmt::Display for CommandError {
                 project_name,
                 org_id,
             } => write!(f, "Cannot find project {}.{}", project_name, org_id),
+            CommandError::BadJson() => write!(f, "Bad json"),
```
geigerzaehler: If the user see’s this they will have no idea what JSON is “bad” and why. I think we should add some context.

cli/src/command/account.rs
NunoAlexandre: yeah, was going to do that 👍

cli/src/command/account.rs
NunoAlexandre: The thing is that showing such errors to users adds no value whatsoever. This should only error if someone but the cli is messing up with the file. If that happens, we just basically start fresh. It seems like fair resilience behaviour.

cli/src/lib.rs
NunoAlexandre: Didn't keep them consistent, sorry. About `CommandError::BadJson`, not sure how to go about that. For users, since they don't pass json in, the only way for this error to happen is if they or someone messes up the accounts file, which we handle anyway. As a user it would be weird to me to get a bad json error when creating or listing accounts 🤔

cli/src/lib.rs
NunoAlexandre: yeah. see https://github.com/radicle-dev/radicle-registry/pull/303#discussion_r399370528.

cli/src/command/account.rs
NunoAlexandre: Do you ever see value in qualifying imports?

geigerzaehler: I’m not sure I understand what you mean. In this case I would remove `use serde_json as json` and use `serde_json` explicitly everywhere. Then it is unambiguous what `serde_json::from_reader` does.

cli/src/command/account.rs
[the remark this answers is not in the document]
NunoAlexandre: The renaming is fine, the docs I am not convinced because it's not true. We do generate a key_pair but only to get the seed out. (Maybe we could just compute the seed instead). What we store on disk is the association name - seed.

What about this?

```suggestion
/// Generate a local account identified by `name` and store it on disk.
```

cli/src/command/account.rs
```diff
@@ -110,3 +161,106 @@ impl CommandT for Transfer {
...
+            });
+        }
+
+        accounts
+    }
+
+    fn get_file() -> File {
+        //TODO(nuno): find a better and more qualified path.
+        let dir = BaseDirs::new().unwrap();
+        let path = dir.home_dir().join("accounts.json");
```
geigerzaehler: On Linux this resolves to `$HOME/accounts.json` which does not seem to be a good idea. How about
```suggestion
        let path = dir.data_dir().join("radicle-registry-cli", "keys.json");
```
On linux this would resolve to `~/.local/share/radicle-registry-cli/keys.json`.

NunoAlexandre: I'm on mac and here `data_dir` translates to `$HOME/Library/Application Support` which the cli fails to create in due to permissions. I guess we want to support mac & windows?

cli/src/lib.rs
geigerzaehler: > As a user it would be weird to me to get a bad json error when creating or listing accounts thinking

I think the opposite would be rather confusing. Assume your account storage get’s corrupted and you run `account list` and you get nothing. Then you as a user ask yourselve what is going. All tools that I can think of print an error when there databases are corrupted.

NunoAlexandre: Even ff tools? We can always pass that info upstream but then we should provide a way for users to fix it. maybe `cli account reset`? Losing accounts has no happy ending.

cli/src/command/account.rs
geigerzaehler: > I'm on mac and here `data_dir` translates to `$HOME/Library/Application Support` which the cli fails to create in due to permissions. I guess we want to support mac & windows?

Does `$HOME/Library/Application Support` not exist? If so the CLI should probably create it recursively.

cli/src/lib.rs
geigerzaehler: > Even ff tools? We can always pass that info upstream but then we should provide a way for users to fix it. maybe `cli account reset`?

I don’t think we necessarily need to provide a way to fix this. And it does not cost us a lot providing the necessary information. I think this is especially important that users can report these errors in the early phase.

cli/src/command/account.rs
NunoAlexandre: I meant that I understand your point, but I also see value in qualifying imports, so I wondered if you ever see value in doing so. In this case, using `serde_json` doesn't bother me, but rust is already so verbose so I like to have the code a little less crowded.

cli/src/lib.rs
NunoAlexandre: I disagree but I am ok with waiting for user feedback in this first phase.

cli/src/command/account.rs
NunoAlexandre: Ok, so I had seen that `write` creates the file passed if it doesn't exist so I thought it would do the same to the path's directory but apparently it doesn't. Fixed by ensuring the directory exists.

cli/src/command/account.rs
NunoAlexandre: Discussed aside, now the errors are propagated properly 👍

cli/src/command/account.rs
NunoAlexandre: Yeah, that's good! thanks

NunoAlexandre:
@geigerzaehler all addressed, appreciate your second take 🛹

cli/src/account_storage.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const FILE: &str = "accounts.json";
+
+// Get the path to the accounts file on disk.
+//
+// If the file does not yet exist, create it and initialize
+// it with an empty object so that it can be deserialized
+// as an empty HashMap<String, AccountData>.
+fn get_or_create_path() -> Result<PathBuf, Error> {
```
NunoAlexandre: This function, as well as `fn dir()` return `PathBuf` instead of `Path` because paths are only operated as references and we [can't return a reference to a variable created in a function](https://stackoverflow.com/questions/32682876/is-there-any-way-to-return-a-reference-to-a-variable-created-in-a-function). I tried a lot of different things but hit a wall here. After some reading, it seems like it's not possible to do it at all and the previous link shows. These two following links seem to provide a way to make it happen but that doesn't seem better to me.

- https://stackoverflow.com/questions/36706429/is-it-possible-to-return-either-a-borrowed-or-owned-type-in-rust
- https://doc.rust-lang.org/std/borrow/enum.Cow.html

cli/src/account_storage.rs
```diff
@@ -0,0 +1,120 @@
...
+fn dir() -> Result<PathBuf, Error> {
+    let dir = BaseDirs::new()
+        .unwrap()
+        .data_dir()
+        .join("radicle-registry-cli");
+    std::fs::create_dir_all(dir.clone()).map_err(Error::IOError)?;
+    Ok(dir)
+}
+
+impl core::fmt::Display for Error {
```
geigerzaehler: Suggestion: You could use [thiserror](https://docs.rs/thiserror/1.0.11/thiserror/) as recommended in [radicle decisions](https://github.com/radicle-dev/radicle-decisions/blob/master/proposals/0001.md) to simplify adding the display implementation.

cli/src/command/account.rs
```diff
@@ -16,12 +16,15 @@
 //! Define the commands supported by the CLI related to Accounts.
 
 use super::*;
+use crate::account_storage as storage;
```
geigerzaehler: I’d like to avoid this aliasing. It makes the code less understandable without knowing the imports and only saves a couple of characters.

cli/src/account_storage.rs
```diff
@@ -0,0 +1,120 @@
...
+    accounts.insert(name, data);
+    update(accounts)
+}
+
+/// List all the stored accounts
+///
+/// It can fail from IO and Serde Json errors.
+pub fn list() -> Result<HashMap<String, AccountData>, Error> {
+    let path_buf = get_or_create_path()?;
+    let file = File::open(path_buf.as_path()).map_err(Error::IOError)?;
```
geigerzaehler: nit:
```suggestion
    let file = File::open(&path_buf).map_err(Error::IOError)?;
```

cli/src/account_storage.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use directories::BaseDirs;
+use sp_core::serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fs::File;
+
+use std::io::Error as IOError;
+use std::path::PathBuf;
+
+/// AccountData
```
geigerzaehler: Rustdoc displays the first paragraph (or sentence) next to the item in the module overview. If it just repeats the struct name the documentation reader does not get any additional information from that. Why not just use the first sentence from below?

cli/src/account_storage.rs
```diff
@@ -0,0 +1,120 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Accounts Storage
```
geigerzaehler: See my other comment: This is displayed next to the module in the documentation and does not provide any additional information

cli/src/account_storage.rs
NunoAlexandre: Ah, interesting! Is this possible because of https://doc.rust-lang.org/std/path/struct.Path.html#impl-Borrow%3CPath%3E?

cli/src/account_storage.rs
NunoAlexandre: I forgot that, will amend. I just generated the docs to see the result of all this and this doesn't seem to be included 🤔 It only gives us this:

Screenshot 2020-03-30 at 10 31 52

geigerzaehler: You can see it with `--document-private-items`.

cli/src/account_storage.rs
geigerzaehler: This is possible because of the `Deref<Target=Path>` implementation. See the [“coercion” section](https://doc.rust-lang.org/core/ops/trait.Deref.html#more-on-deref-coercion) of deref.

cli/src/command/account.rs
NunoAlexandre: I don't mean neither tend to be savvy, but "a couple characters" sum up to many characters, even though that's not my motivation to do this. We are already in the context of "account", so I'd just avoid the stutter. 

Now, I am ok with dropping this, but so far everyone has an opinion and equally convinced of their position. @xla we had a discussion about stuttering in a PR you owned, which motivated me to do this when I see fit. Can we agree on some criteria?

cli/src/account_storage.rs
NunoAlexandre: Thanks. I now see https://doc.rust-lang.org/std/path/struct.PathBuf.html#deref-methods.

How does it know this `Path` implements it if we don't import the `Deref` trait? Is it implicitly imported?

cli/src/account_storage.rs
NunoAlexandre: Thanks. We could also make `mod account_storage;` `pub mod account_storage;` but I think this is fine.

cli/src/account_storage.rs
geigerzaehler: > Is it implicitly imported?

The [prelude](https://doc.rust-lang.org/std/prelude/index.html) doesn’t mention it and I don’t think it’s possible to call `path_pub.deref()`. The coercion is working though so it must be done deeper in the compilation.

cli/src/command/account.rs
geigerzaehler: At the moment we don’t use import aliases only to disambiguate and for modules and crates we don’t use them at all. In fact I [previously removed](https://github.com/radicle-dev/radicle-registry/pull/12) all the aliases coming from substrate because the code was unreadable for the uninitiated.

I’d propose to add this as a guideline to our contributing guide.

NunoAlexandre: I am fine with that, all for avoiding this kind of discussion. 👍 Will update the contributing file.

PR #305, opened 2020-03-30
Operate on account names

### LINKED ISSUE #304: Adjust CLI to receive account name instead of address or seed
After #275, we have local accounts identified by unique names. Having that allows us to switch from expecting seeds or addresses as command args / options to receive account names instead, whenever possible.

NunoAlexandre:
Closes #304 

- Receive account name instead of seed for `tx-author` cmd line option
- Drop `account address` command since `account list` supersedes that functionality

##Note

I was thinking of also changing `account balance` to receive either an account name or an SS58 address (as it does now). Doing so would make it more user friendly to get the balance of an owned account. I want to think more about this but I'd prefer to keep it aside from this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "blake2b_simd"
+version = "0.5.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


NunoAlexandre:
Need to amend the PR and can't make it a draft, closing it for now.

Issue #292, opened 2020-03-26
FFnet RAD distribution


geigerzaehler:
For users to participate in the FFnet they need to acquire RAD.

- We need to determine the process for acquiring
- We need to determine who receives the tokens in genesis.
- We want to uncover additional requirements for RAD distribution
- We should look at other networks for reference
- We want to write down the process so that it can be made available to FFnet participants

geigerzaehler:
I propose the following solution: Every registry developer generates a key and we put that in to the ffnet genesis spec. I would like to avoid having a publicly known Alice account with all the funds to prevent people from abusing it, even if it is just the ffnet. We should also keep the sudo key private.

Issue #283, opened 2020-03-24
Fix CLI leaked abstraction


NunoAlexandre:
Addresses #281 

### Scenario

- Some commands don't need to interact with the chain at all. e.g: `account address`
- Some commands _query_ the chain
   No transactions involved, only calling client functions that return data.
   Here, only `node_host` is needed.

- Some commands submit transactions to the chain
   Here, all current `CommandLineOptions` are needed: `node_host`, `tx_author`, and `tx_fee`.

### Problems

We have two intertwined problems to resolve:

P0. `CommandLineOptions` are currently global, even though they are only useful to some commands.

P1. All commands (implementations of `CommandT`) receive a `CommandContext`(build from `CommandLineOptions`) even though some don't need to. That means that a `CommandContext` is always created independently of whether the selected command interacts with the chain or not. Commands that don't interact with the chain fail if the `CommandContext.client` fails to connect to the specified `node_host`, which - again - is unnecessary in the first place.  (See #281)

### Goals

G0. Offer cmd-line options based on each command

G1. Only instantiate `Client` when strictly necessary, i.e, only when the selected command needs to use it.

### Solution

Have each command define the options that it needs to run.
Should be done by `flatten`ing shared options, one for the `node_host` and the other for the transaction-related options.

Note: I need to think a bit more about this.

PR #306, opened 2020-03-30
Operate on account names

### LINKED ISSUE #304: Adjust CLI to receive account name instead of address or seed

NunoAlexandre:
Closes #304 

- Receive account name instead of seed for `tx-author` cmd line option
- Drop `account address` command since `account list` supersedes that functionality

## Notes

1. No more `//Alice`, _the_ account with funds. We need to think about how we will survive like this. (#292)

2. I was thinking of also changing `account balance` to receive either an account name or an SS58 address (as it does now). Doing so would make it more user friendly to get the balance of an owned account. I want to think more about this but I'd prefer to keep it aside from this PR.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -47,16 +47,12 @@ impl CommandLine {
 /// The accepted command-line options
 #[derive(StructOpt, Clone)]
 pub struct CommandLineOptions {
-    /// Value to derive the key pair for signing transactions.
-    /// See
-    /// <https://substrate.dev/rustdocs/v1.0/substrate_primitives/crypto/trait.Pair.html#method.from_string>
-    /// for information about the format of the string
+    /// The name of the local account to be used to sign transactions.
     #[structopt(
         long,
-        default_value = "//Alice",
         env = "RAD_TX_AUTHOR",
-        value_name = "key_pair",
-        parse(try_from_str = Self::parse_key_pair)
+        value_name = "account_name",
+        parse(try_from_str = Self::lookup_account)
     )]
     pub tx_author: ed25519::Pair,
 
@@ -76,8 +72,12 @@ pub struct CommandLineOptions {
 }
 
 impl CommandLineOptions {
-    fn parse_key_pair(s: &str) -> Result<ed25519::Pair, String> {
-        ed25519::Pair::from_string(s, None).map_err(|err| format!("{:?}", err))
+    fn lookup_account(name: &str) -> Result<ed25519::Pair, String> {
+        let accounts = account_storage::list().map_err(|e| format!("{}", e))?;
+        match accounts.get(&name.to_string()) {
+            Some(account) => Ok(ed25519::Pair::from_seed(&account.seed)),
+            None => Err(format!("Could not find local account named '{}'", name)),
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


cli/src/lib.rs
```diff
@@ -47,16 +47,13 @@ impl CommandLine {
 /// The accepted command-line options
 #[derive(StructOpt, Clone)]
 pub struct CommandLineOptions {
-    /// Value to derive the key pair for signing transactions.
-    /// See
-    /// <https://substrate.dev/rustdocs/v1.0/substrate_primitives/crypto/trait.Pair.html#method.from_string>
-    /// for information about the format of the string
+    /// The local account name to be used to sign a transaction.
+    /// TODO(nuno)
```
geigerzaehler: What is the todo here? I’d also prefer to have these todos as issues. Otherwise it becomes difficulty to decide what to work on.

cli/src/lib.rs
```diff
@@ -47,16 +47,13 @@ impl CommandLine {
...
     #[structopt(
         long,
-        default_value = "//Alice",
         env = "RAD_TX_AUTHOR",
-        value_name = "key_pair",
-        parse(try_from_str = Self::parse_key_pair)
+        value_name = "account_name",
+        parse(try_from_str = Self::lookup_account)
     )]
     pub tx_author: ed25519::Pair,
```
geigerzaehler: We should probably change the name of this option to `--account`.

cli/src/lib.rs
```diff
@@ -76,8 +73,12 @@ pub struct CommandLineOptions {
...
 
 impl CommandLineOptions {
-    fn parse_key_pair(s: &str) -> Result<ed25519::Pair, String> {
-        ed25519::Pair::from_string(s, None).map_err(|err| format!("{:?}", err))
+    fn lookup_account(name: &str) -> Result<ed25519::Pair, String> {
+        let accounts = account_storage::list().map_err(|e| format!("{}", e))?;
+        match accounts.get(&name.to_string()) {
+            Some(account) => Ok(ed25519::Pair::from_seed(&account.seed)),
+            None => Err(format!("Could not find local account named '{}'", name)),
+        }
```
geigerzaehler: Suggestion: We’re putting side effects and state dependencies into the parsing logic here. Alternatively we could make this return just a string and implement an `account_pair()` method on `CommandContext`

cli/src/lib.rs
NunoAlexandre: I wanted to improve the doc and forgot to do it before raises the PR, my bad. I rarely leave TODO's cross merge, so I don't think the suggestion applies here.

cli/src/lib.rs
NunoAlexandre: Well, before this option received a `seed` and it was called `tx-author`, I don't think it makes less sense to keep being named that way. The option stands to define the tx author, based on a local account name. As a user, I'd prefer to have it as `tx-author`.

cli/src/lib.rs
NunoAlexandre: Yeah, I thought the same about the mix but at the same time, this is pretty seamless and done in one place instead of accepting whatever string as `tx-author` to only then figure it isn't a valid one. In #283 I plan to rethink the CommandContext, so if you'd be alright with that I would prefer to keep this as is and rethink it based on the new structure.

PR #307, opened 2020-03-30
contributing: Add 'Imports' section

NunoAlexandre:
Find the motivating discussion at https://github.com/radicle-dev/radicle-registry/pull/303#discussion_r400000226.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,13 @@ as a reviewer, always assign yourself and unassign the team before you start the
 - Leave additional comments explaining 'why' rather than 'how', strive to have the code clean
   and elegant to make this aim viable.
 
+# Code
+
+## Imports
+
+- Only use import aliases to disambiguate, be it types, modules, or crates (in the case of
+  having two coexisting versions).
+
 # Changelog
 
 We use `./CHANGELOG.md` to record all changes visible to users of the client.


CONTRIBUTING.md
```diff
@@ -45,6 +45,12 @@ via the GitHub UI.
   and cryptocurrencies.
 - Leave additional comments explaining 'why' rather than 'how', strive to have the code clean
   and elegant to make this aim viable.
+  
+# Code
+
+### Imports
+
+- Only use import aliases to disambiguate from local types.
```
geigerzaehler: “Local types” might be too limiting. What I had in mind was this, which is fairly common”
~~~rust
use abc::foo::Error as FooError
use xyz::bar::Error as BarError
~~~
I think it also wouldn’t hurt to explicitly call out that we should not alias crates and modules at all.

NunoAlexandre: As for the first point, it makes sense. I viewed it the way I wrote it given the context of, for example, me doing `std::io::Error as IOError` when I already had an `Error` type locally. 

It sounds all good, will adjust.

NunoAlexandre: @CodeSandwich, regarding [your comment](https://github.com/radicle-dev/radicle-registry/pull/307#pullrequestreview-383681477), can you have a look at this discussion and at the latest changes?

geigerzaehler:
> I'm against it. This is a matter of a clean code and this is not something that can be enforced with strict rules. Sure, aliasing for the sake of aliasing **usually** reduces ease of reading, but this isn't something to set in stone.

Can you give an example where using an alias is cleaner? Having this guideline doesn’t mean that there are no exceptions. It just ensures consistency and establishes a baseline that avoids style discussions.

CodeSandwich:
@NunoAlexandre I did, before writing this comment ;)
@geigerzaehler I don't know when aliasing would make sense, maybe there are such cases, but I can't come up with any, which is why I agree, that aliasing without conflicts is usually bad. This guideline is very strict, there is no space for exceptions and no commit with aliasing will be accepted while it's in power, which is probably fine. If both of you feel that this common good practice is really worth writing down as a hard, official rule, then I'm fine with it.

CONTRIBUTING.md
```diff
@@ -46,6 +46,20 @@ via the GitHub UI.
 - Leave additional comments explaining 'why' rather than 'how', strive to have the code clean
   and elegant to make this aim viable.
 
+# Code
+
+### Imports
```
CodeSandwich: Why `###`, wouldn't `##` be enough?

CONTRIBUTING.md
```diff
@@ -46,6 +46,20 @@ via the GitHub UI.
...
+
+- Aliases
+
+  Only use import aliases to disambiguate, as shown below.
+  Do not alias crates and modules.
```
CodeSandwich: Sometimes aliasing crates makes sense for disambiguating, e.g. when there are 2 versions of the same crate. It's rare though.

CodeSandwich: I'm really not sure about modules, I can think of a reasonable case, e.g. `crate1::service::worker::types` and `crate2::service::worker::types`. It would probably make sense to have aliases to disambiguate `types` modules, I'm not sure. What's your opinion?

geigerzaehler: I think you’re right. In your example one would probably want `crate1_types` and `crate2_types`. I think this neatly falls under “disambiguate” exception so I’d suggest just removing “modules” from the sentence.

geigerzaehler:
@CodeSandwich As outline above, I believe there is value in having such a guideline. As for the strictness I believe it’s preferable to establish clear guidelines to ensure that the guidelines are followed by default. This doesn’t mean that we can’t treat them flexibly on a case by case basis or that we can’t amend them.

I’d suggest we start from this and see if we need to extend the possible exceptions. This is an approach that we already have been following with the abbreviations.

CONTRIBUTING.md
NunoAlexandre: You are right. The reason is that usually, a page should only have one `h1`, so I am use to have `###` after `##`. The first one I got right, the second was the power of habit.

CONTRIBUTING.md
NunoAlexandre: Fixed.

NunoAlexandre:
> @NunoAlexandre I did, before writing this comment ;)

My comment was regarding the amends post your comment, so it can't be your had written it before.

> As for the strictness I believe it’s preferable to establish clear guidelines to ensure that the guidelines are followed by default. This doesn’t mean that we can’t treat them flexibly on a case by case basis or that we can’t amend them.

I’d suggest we start from this and see if we need to extend the possible exceptions. This is an approach that we already have been following with the abbreviations.

I completely agree. Rules are there to be broken and common sense is a thing.

CONTRIBUTING.md
CodeSandwich: Great! What do you think about crates?

CodeSandwich:
> My comment was regarding the amends post your comment, so it can't be your had written it before.

I thought you were referring the discussion that lead to this PR.

CONTRIBUTING.md
geigerzaehler: I think we should not allow crate aliases. I can’t think of a good reason to have them. Even if the crates are badly named I’d prefer the keep them. It makes the code just more accessible since you don’t have to learn how we are calling the crate. And in the worst case our code base is inconsistent on how we name crates which makes it even harder to understand.

CodeSandwich: I mean the case with usage of 2 versions of the same crate. At least one of them must be aliased in `Cargo.toml` in order to make them distinguishable.

geigerzaehler: > I mean the case with usage of 2 versions of the same crate. At least one of them must be aliased in `Cargo.toml` in order to make them distinguishable.

Excellent point. I believe this also falls under the “disambiguate exception”. So let’s just drop the “do not alias ...” sentence.

CONTRIBUTING.md
```diff
@@ -46,6 +46,19 @@ via the GitHub UI.
...
+## Imports
+
+- Aliases
+
+  Never alias crates.
+  Only use import aliases to disambiguate:
+  ``` rust
+  use abc::foo::Error as FooError;
+  use xyz::bar::Error as BarError;
+  ```
```
NunoAlexandre:
```suggestion
- Only use import aliases to disambiguate:
  ``` rust
  use abc::foo::Error as FooError;
  use xyz::bar::Error as BarError;
  ```
```

CONTRIBUTING.md
NunoAlexandre: Done.

CONTRIBUTING.md
```diff
@@ -54,6 +54,15 @@ as a reviewer, always assign yourself and unassign the team before you start the
...
 
+# Code
+
+## Imports
+
+- Only use import aliases to disambiguate:
+  ``` rust
+  use abc::foo::Error as FooError;
+  use xyz::bar::Error as BarError;
+
```
CodeSandwich: The code block is open

CONTRIBUTING.md
```diff
@@ -54,6 +54,15 @@ as a reviewer, always assign yourself and unassign the team before you start the
 - Leave additional comments explaining 'why' rather than 'how', strive to have the code clean
   and elegant to make this aim viable.
 
+# Code
+
+## Imports
+
+- Only use import aliases to disambiguate:
```
CodeSandwich: Maybe it would be good to include examples of modules and crate versions disambiguations. Just 4 lines of code.

NunoAlexandre:
@CodeSandwich feel free to leave suggestions that can be committed.

CodeSandwich:
> @CodeSandwich feel free to leave suggestions that can be committed.

I can't suggest any changes in a proper way, because all of them contain triple back-ticks. Here are examples to copy-paste.

```rust
use abc::foo;
use xyz::foo;
```
```rust
foo_1 = { package = "foo", version = "1.0.0" }
foo_2 = { package = "foo", version = "2.0.0" }
```

NunoAlexandre:
@CodeSandwich regarding your example, how does `abc` and `xyz` related to `foo_1` and `foo_2`?

NunoAlexandre:
@CodeSandwich I have rebased, fixed the missing code block (```) and rephrased the sentence to be comprehensive. If you feel strongly about having comprehensive examples, feel free to commit to this PR :+1:

CodeSandwich:
> If you feel strongly about having comprehensive examples

I don't, if we don't include any, I'm fine with it.

PR #310, opened 2020-03-30
Update Rust nightly and add Git to CI

geigerzaehler:
We update the Rust nightly version and rebuild the CI base image. We also add Git to the CI base image so that `vergen` can inject the commit sha properly.


cli/src/command/org.rs
```diff
@@ -74,9 +74,9 @@ impl CommandT for Show {
                     org_id: self.org_id.clone(),
                 })?;
 
-        println!("id: {}", org.id.clone());
-        println!("account_id: {}", org.account_id.clone());
-        println!("members: {:?}", org.members.clone());
+        println!("id: {}", org.id);
+        println!("account_id: {}", org.account_id);
+        println!("members: {:?}", org.members);
```
NunoAlexandre: How is it that this no longer needs clones?

geigerzaehler: The clone was redundant before, clippy just didn’t catch it. The new version of clippy does though.

NunoAlexandre: Ok :) But I'm surprised that it's redundant. Wouldn't `org.field` borrow `org`?

Issue #226, opened 2020-02-17
Missing impl for std::error::Error


xla:
Currently it's not possible to use the `try!` macro / `?` operator on the `client::error::Error` in a context that returns a `std::error::Error`. Which makes it quite unergonomic.

geigerzaehler:
Could you explain what you mean by “returns a `std::error::Error`” since that is a trait? An example would be nice. Is implementing `std::error::Error` for `client::Error` enough?

xla:
@geigerzaehler Yes, that's enough. An example is a commonly found main function with the following signature: `fn main() -> Result<(), Box<dyn std::error::Error>>`

geigerzaehler:
Thanks for the clarification. If this is not urgent I’d like to tackle this as part of an error revamp that we plan to do. Otherwise you’re welcome to add the trait implementation as a quick fix.

xla:
This has no urgency. :)

PR #311, opened 2020-03-30
Implement std Error for client::Error

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/226

There are a few changes in the API to make it more conventional. @xla please confirm if they are OK with you.

### DIFF
--- a/client/src/error.rs
+++ b/client/src/error.rs
@@ -13,20 +13,37 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use failure::{Compat, Fail};
 use jsonrpc_core_client::RpcError;
 use parity_scale_codec::Error as CodecError;
 
-/// Error that may be returned by any of the [crate::ClientT] methods.
-#[derive(Debug, derive_more::From, derive_more::Display, derive_more::TryInto)]
+/// Error that may be returned by any of the [crate::ClientT] methods
+#[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Decoding data received from failed.
-    Codec(CodecError),
-    /// Error from the underlying RPC connection.
-    Rpc(RpcError),
+    /// Decoding the received data failed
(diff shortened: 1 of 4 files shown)


CodeSandwich:
> Is there a technical reason to drop the period marks at the end of the affected sentences?

No, it just feels right when single sentence lists don't have dots.

client/src/error.rs
```diff
@@ -13,22 +13,40 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+use failure::{Compat, Fail};
 use jsonrpc_core_client::RpcError;
 use parity_scale_codec::Error as CodecError;
+use thiserror::Error;
```
NunoAlexandre, line 19: following to the discussion around ambiguity, I'd suggest us applying that here already:

```suggestion
use thiserror::Error as ThisError;
```

client/src/error.rs
```diff
@@ -13,22 +13,40 @@
...
 
-/// Error that may be returned by any of the [crate::ClientT] methods.
-#[derive(Debug, derive_more::From, derive_more::Display, derive_more::TryInto)]
+/// Error that may be returned by any of the [crate::ClientT] methods
+#[derive(Debug, Error)]
```
NunoAlexandre, line 22:
```suggestion
#[derive(Debug, ThisError)]
```

client/src/error.rs
CodeSandwich: Interesting case of ambiguity, unclear for humans, but easy for the compiler, good catch! I'd rather avoid aliasing if not necessary.

Fixed, but differently.

client/src/error.rs
CodeSandwich: Fixed, but differently

client/src/error.rs
NunoAlexandre: ahah indeed, it;s usually the way around. Your fix is good :+1:

PR #312, opened 2020-03-30
Upgrade Substrate

CodeSandwich:
The usual routine. This time with our issue addressed and emojis in logs!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,9 +231,9 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.26"
+version = "0.1.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


NunoAlexandre:
@CodeSandwich, in this case, it makes sense to upgrade, but how often do you plan to do this? Every upgrade means a substantial slow down in build times without much benefit if done too often.

CodeSandwich:
Sorry about that, but the more often the better, I'm trying to keep it around once a week. Substrate keeps adding breaking changes every few days and if they pile up, they will be very hard to apply at once. The last big update was a big effort and besides bugs a lot of work went into just fixing the changes. The few minutes of full compilation once a week is a small price in comparison.

NunoAlexandre:
Ok, it sounds like a fair trade-off then :) 👍

Issue #285, opened 2020-03-26
Expose block height using client interface


CodeSandwich:
We need to get block height based on block hash and the current chain height using the client interface.

geigerzaehler:
Why do we need to get the block height for the block hash? Is it because only the block hash is part of the transaction response? If so, we should probably include the block height in the transaction response to save us a roundtrip.

xla:
> Why do we need to get the block height for the block hash? Is it because only the block hash is part of the transaction response? If so, we should probably include the block height in the transaction response to save us a roundtrip.

I think we mean the same when we say height for the hash, from the consumer perspective it's just important that whenever we return a block from anywhere it includes the height, which is extremely vital to interaction with that info.

geigerzaehler:
> from the consumer perspective it's just important that whenever we return a block from anywhere it includes the height,

That is good to know. If you get a block or a block header the height is automatically included. But we dont’t have any APIs that return these.

It’s unclear to me though if you also need this information if the API only gives you a _reference_ to a block, say, in the form of a block hash. In this case we also have to answer the question if we want to include the height as part of the response or add a method that allows you to retrieve the header (including the height) from the hash. (I personally would lean towards the latter to keep the client as simple as possible. If retrieving the height is a common pattern consumers can build their own abstractions.)

geigerzaehler:
Superseded by #320

PR #313, opened 2020-03-31
Return block number when sending message

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,3 +53,4 @@ Upcoming
   * `ClientT::genesis_hash()`
   * `Transaction`
   * `Transaction::new_signed()`
+* Add `block_number` to `TransactionApplied` returned from client API
--- a/client/src/backend/emulator.rs
+++ b/client/src/backend/emulator.rs
@@ -133,6 +133,7 @@ impl backend::Backend for Emulator {
             backend::TransactionApplied {
                 tx_hash,
                 block: header.hash(),
+                block_number: header.number,
                 events,
             },
         ))))
--- a/client/src/backend/mod.rs
+++ b/client/src/backend/mod.rs
@@ -16,7 +16,7 @@
 //! Define trait for client backends and provide emulator and remote node implementation
 use futures::future::BoxFuture;
 
-pub use radicle_registry_runtime::{Hash, UncheckedExtrinsic};
+pub use radicle_registry_runtime::{BlockNumber, Hash, UncheckedExtrinsic};
 
 use crate::interface::*;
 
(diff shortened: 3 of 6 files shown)


CodeSandwich:
> Looks good overall, my only remark is to rename `number` to `block_number`.

Good idea :+1:

> We could also group the block-related fields in a `BlockInfo` or some type alike that would have `hash` and `number`. Not important now.

While I agree that it would add some value, it would also make this change a breaking one for the API.

NunoAlexandre:
> While I agree that it would add some value, it would also make this change a breaking one for the API.

Yeah, it adds no value at the moment, but probably a good idea if we add a 3rd block-related field to it.

xla:
> Fixes #285

It doesn't entirely address the issue as we still need the ability to get the current height.

CodeSandwich:
> Using `number` instead of `height` doesn't really match the expectation of the domain. Never seen it to be referred to anything but the `height`. Is there a particular reason we call it differently?

That's a Substrate naming convention. I find it confusing and would rather use `height`, but having two names would be worse.

xla:
> That's a Substrate naming convention. I find it confusing and would rather use `height`, but having two names would be worse.

Having substrate dictate our public API seems rather odd, I'd push to find what makes sense for our (`radicle Registry`) domain understanding and if this is `height` use that.

CodeSandwich:
> It doesn't entirely address the issue as we still need the ability to get the current height.

Ouch, sorry about that. Could you respond to this comment? https://github.com/radicle-dev/radicle-registry/issues/285#issuecomment-604317880 This way we can discuss a proper solution.

CodeSandwich:
> Having substrate dictate our public API seems rather odd, I'd push to find what makes sense for our (`radicle Registry`) domain understanding and if this is `height` use that.

@geigerzaehler Has it ever been discussed? If not, that's very much worth settling.

The ugly part is that this particular naming convention may have already leaked through type names, e.g. `BlockNumber`.

xla:
> Ouch, sorry about that. Could you respond to this comment? [#285 (comment)](https://github.com/radicle-dev/radicle-registry/issues/285#issuecomment-604317880) This way we can discuss a proper solution.

Followed up there, but I think it that comment isn't concerned with fetching the current chain height.

geigerzaehler:
> > Having substrate dictate our public API seems rather odd, I'd push to find what makes sense for our (`radicle Registry`) domain understanding and if this is `height` use that.
> 
> @geigerzaehler Has it ever been discussed? If not, that's very much worth settling.
> 
> The ugly part is that this particular naming convention may have already leaked through type names, e.g. `BlockNumber`.

We haven’t discussed this yet. There’s at least one other instance where there’s a mismatch in terminology: “extrinsic” vs “transaction”. In that case we’re also using the more common term “transaction” over the substrate specific “extrinsic”.

I think in general we should lean towards the more common terms in our public APIs. This might of course create a mismatch of terms in our code base but it’s easier to make this clear to the developers than to the consumers.

Issue #281, opened 2020-03-23
CLI commands that don’t require a node fail if a node is not running


geigerzaehler:
The `cli show-address` command fails if the CLI cannot connect to a node even though the command does not need a to connect.

CodeSandwich:
This is an opportunity to solve another small issue: the CLI has a rigid set of options, which in many sub-commands are unnecessary.

NunoAlexandre:
See also https://github.com/radicle-dev/radicle-registry/pull/279#discussion_r396992671.

NunoAlexandre:
@radicle-dev/registry I'm taking on this one if that's alright with you.

NunoAlexandre:
Fixed by #315

PR #315, opened 2020-03-31
Cli/fix leaked abstraction

### LINKED ISSUE #314: Bootstrapping CLI accounts is not working
When using the CLI for the first and no account exists yet it is impossible to create an account. The following command fails

```
$ radicle-registry-cli --tx-author default account generate default
error: Invalid value for '--tx-author <account_name>': Could not find local account named 'default'
```

NunoAlexandre:
Fixes #314 #283 #281 

As explained in #283, have each command define the options that it needs to run.
Done by flattening shared options, one for the network and the other for the transaction-related options.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -23,8 +23,7 @@ pub mod account_storage;
 mod command;
 use command::{account, org, other, project, user};
 
-/// The type that captures the command line,
-/// composed by a [Command] and [CommandLineOptions].
+/// The type that captures the command line.
 #[derive(StructOpt, Clone)]
 #[structopt(max_term_width = 80)]
 pub struct CommandLine {


cli/src/command/account.rs
```diff
@@ -19,41 +19,45 @@ use super::*;
 use crate::account_storage;
 
 /// Account related commands
-#[derive(StructOpt, Debug, Clone)]
```
geigerzaehler: Could we keep the `Debug` implementation. I know it’s not necessary but its intended for debugging after all. It will become very annoying if you want to `println!` a command and you first need to add `Debug` to it and all its constituents.

NunoAlexandre: So I had to remove it because `ed25519::Pair` does not implement it, so any type including it won't be able to derive it. A way I see around this is to have the `TxOptions` `author` field be an account name and have each command get the key pair from it, as you suggested in a previous PR. We could have this one as is and I address that change in as separate PR?

CodeSandwich:
I've removed all the `Box`es and it's compiling without any warnings.

CodeSandwich:
Ok, it's a clippy warning. Ugh.

NunoAlexandre:
That is a good observation, @CodeSandwich. I like that more in the given context. @geigerzaehler do you agree?

CodeSandwich:
Maybe we should annotate the whole cli crate, it contains no high performance code or structures and this is going to kick us over and over again.

NunoAlexandre:
@CodeSandwich @geigerzaehler I added `#![allow(clippy::large_enum_variant)]` to the cli `lib.rs` instead of `Box`ing each hurting command. Nothing else to address here.

PR #317, opened 2020-03-31
cli: use thiserror for CommandError Display

### LINKED ISSUE #308: cli: Use `thiserror` for all error types
Follow the example in place for `radicle_registry_cli::account_storage::Error` and do the same for all other CLI errors.

Note: scoping this to the `cli` only as I have and am working on mostly at the moment, avoiding touching the whole project but start by improving the cli now instead.

NunoAlexandre:
Closes #308 

I thought we had more error types in the CLI in need of this but it turns out that only `CommandError` was missing.

### DIFF
--- a/cli/src/account_storage.rs
+++ b/cli/src/account_storage.rs
@@ -45,37 +45,36 @@ pub enum Error {
 
     /// Failed to write to the accounts file
     #[error("Failed to write to the accounts file: {0}")]
-    FailedWrite(WritingError),
+    FailedWrite(#[from] WritingError),
 
     /// Failed to read the accounts file
     #[error("Failed to read the accounts file: {0}")]
-    FailedRead(ReadingError),
+    FailedRead(#[from] ReadingError),
 }
-
 
 /// Possible errors when writing to the accounts file.
 #[derive(Debug, ThisError)]
 pub enum WritingError {
-    #[error("{0}")]
+    #[error(transparent)]
     IO(IOError),
 
-    #[error("{0}")]
+    #[error(transparent)]
     Serialization(serde_json::Error),
 }
 
 /// Possible errors when reading the accounts file.
 #[derive(Debug, ThisError)]
 pub enum ReadingError {
-    #[error("{0}")]
+    #[error(transparent)]
     IO(IOError),
 
-    #[error("{0}")]
+    #[error(transparent)]
     Deserialization(serde_json::Error),
 }
 
 /// Add an account to the storage.
 ///
-/// Fails if an account with the given `name` already exists_.
+/// Fails if an account with the given `name` already exists.
 /// It can also fail from IO and Serde Json errors.
 pub fn add(name: String, data: AccountData) -> Result<(), Error> {
     let mut accounts = list()?;
@@ -87,25 +86,22 @@ pub fn add(name: String, data: AccountData) -> Result<(), Error> {
     update(accounts)
 }
 
-/// List all the stored accountsr
+/// List all the stored accounts.
 ///
 /// It can fail from IO and Serde Json errors.
 pub fn list() -> Result<HashMap<String, AccountData>, Error> {
     let path_buf = get_or_create_path()?;
-    let file = File::open(path_buf.as_path())
-        .map_err(|e| Error::FailedRead(ReadingError::IO(e)))?;
+    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
     let accounts: HashMap<String, AccountData> =
-        serde_json::from_reader(&file)
-            .map_err(|e| Error::FailedRead(ReadingError::Deserialization(e)))?;
+        serde_json::from_reader(&file).map_err(ReadingError::Deserialization)?;
     Ok(accounts)
 }
 
 fn update(accounts: HashMap<String, AccountData>) -> Result<(), Error> {
     let path_buf = get_or_create_path()?;
-    let new_content = serde_json::to_string(&accounts)
-        .map_err(|e| Error::FailedWrite(WritingError::Serialization(e)))?;
-    std::fs::write(path_buf.as_path(), new_content.as_bytes())
-        .map_err(|e| Error::FailedWrite(WritingError::IO(e)))
+    let new_content = serde_json::to_string(&accounts).map_err(WritingError::Serialization)?;
+    std::fs::write(path_buf.as_path(), new_content.as_bytes()).map_err(WritingError::IO)?;
+    Ok(())
 }
 
 const FILE: &str = "accounts.json";
@@ -121,8 +117,7 @@ fn get_or_create_path() -> Result<PathBuf, Error> {
     let path = path_buf.as_path();
 
     if !path.exists() {
-        std::fs::write(path, b"{}")
-            .map_err(|e| Error::FailedWrite(WritingError::IO(e)))?;
+        std::fs::write(path, b"{}").map_err(WritingError::IO)?;
     }
 
     Ok(path_buf)
@@ -133,6 +128,6 @@ fn dir() -> Result<PathBuf, Error> {
         .unwrap()
         .data_dir()
         .join("radicle-registry-cli");
-    std::fs::create_dir_all(&dir).map_err(|e| Error::FailedRead(ReadingError::IO(e)))?;
+    std::fs::create_dir_all(&dir).map_err(ReadingError::IO)?;
     Ok(dir)
 }
--- a/cli/src/account_storage.rs
+++ b/cli/src/account_storage.rs
@@ -39,22 +39,43 @@ type Seed = [u8; 32];
 
 #[derive(Debug, ThisError)]
 pub enum Error {
-    /// An Account with the given name already exists
+    /// An account with the given name already exists
     #[error("An account with the given name already exists")]
     AlreadyExists(),
 
-    /// An IOError occurred
-    #[error("An IO error occured: {0}")]
-    IOError(IOError),
+    /// Failed to write to the accounts file
+    #[error("Failed to write to the accounts file: {0}")]
+    FailedWrite(WritingError),
 
-    /// A serde json Error occurred
-    #[error("A serde json error occured: {0}")]
-    JsonError(serde_json::Error),
+    /// Failed to read the accounts file
+    #[error("Failed to read the accounts file: {0}")]
+    FailedRead(ReadingError),
+}
+
+
+/// Possible errors when writing to the accounts file.
+#[derive(Debug, ThisError)]
+pub enum WritingError {
+    #[error("{0}")]
+    IO(IOError),
+
+    #[error("{0}")]
+    Serialization(serde_json::Error),
+}
+
+/// Possible errors when reading the accounts file.
+#[derive(Debug, ThisError)]
+pub enum ReadingError {
+    #[error("{0}")]
+    IO(IOError),
+
+    #[error("{0}")]
+    Deserialization(serde_json::Error),
 }
 
 /// Add an account to the storage.
 ///
-/// Fails if an account with the given `name` already exists.
+/// Fails if an account with the given `name` already exists_.
 /// It can also fail from IO and Serde Json errors.
 pub fn add(name: String, data: AccountData) -> Result<(), Error> {
     let mut accounts = list()?;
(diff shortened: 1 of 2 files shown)


cli/src/lib.rs
```diff
@@ -135,39 +136,26 @@ pub trait CommandT {
 /// Error returned by [CommandT::run].
 ///
 /// Implements [From] for client errors.
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, derive_more::From, ThisError)]
 pub enum CommandError {
+    #[error("Client error: {0}")]
```
CodeSandwich: AFAIK the error shouldn't print its cause in `Display` if it's `Error` too, that's what `source` is for.

cli/src/lib.rs
```diff
@@ -135,39 +136,26 @@ pub trait CommandT {
...
+    #[error("Transaction {tx_hash:?} failed in block {block_hash:?}")]
     FailedTransaction {
         tx_hash: TxHash,
         block_hash: BlockHash,
     },
-    OrgNotFound {
-        org_id: OrgId,
-    },
+
+    #[error("Cannot find org {org_id:?}")]
```
CodeSandwich: We don't need to print debug for any of the error messages in all variants. Regular display is cleaner.

cli/src/lib.rs
```diff
@@ -135,39 +136,26 @@ pub trait CommandT {
...
-            CommandError::ProjectNotFound {
-                project_name,
-                org_id,
-            } => write!(f, "Cannot find project {}.{}", project_name, org_id),
-            CommandError::AccountStorageError(error) => {
-                write!(f, "Account storage error: {}", error)
-            }
-        }
-    }
+    #[error("Account storage error: {0}")]
```
geigerzaehler: This now gives an error like this
~~~
ERROR: Account storage error: A serde json error occured: invalid length 31, expected an array of length 32 at line 3 column 75
~~~
This is not super readable. I’d suggest 
```suggestion
    #[error("{0}")]
```
and then adjust the message of the account error so that they are understandable to the user. Ideally the message would be something like this
~~~
ERROR: Failed to read accounts file: invalid length 31, expected an array of length 32 at line 3 column 75
~~~

NunoAlexandre: Ok, sounds good :)

Given that IOErrors and Json errors can happen on read/write or deserialize/serialize, I will add corresponding variants to `account_storage::Error` in order to provide that more detailed kind of error message.

cli/src/lib.rs
```diff
@@ -135,39 +136,26 @@ pub trait CommandT {
 /// Error returned by [CommandT::run].
 ///
 /// Implements [From] for client errors.
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, derive_more::From, ThisError)]
```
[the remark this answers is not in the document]
NunoAlexandre: @CodeSandwich what is `source`? #rustnewbie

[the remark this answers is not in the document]
CodeSandwich: It's part of an std error: https://doc.rust-lang.org/std/error/trait.Error.html. It's the error that is wrapped inside another error, because it's its cause. `thiserror` has excellent support for wrapping and then providing source errors: https://docs.rs/thiserror/1.0.9/thiserror/

cli/src/lib.rs
CodeSandwich: This is the concept of source: each layer of error only knows how to format itself and doesn't care about its source. If one is interested in its source, there's a `source` method, it can be then printed too and again called `.source()` on.

CodeSandwich:
It's good to have a function that catches errors on the top-most level in `main.rs` and nicely formats them in a loop unpacking and printing `source`s, e.g. each layer of causality in one line, maybe prefixed with `caused by` or a growing indentation or an arrow.

cli/src/lib.rs
NunoAlexandre: We want to print a `CommandError` and give all useful info to the users. How does that suggestion bring value to users? Honest question, I am not getting it.

cli/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Thanks!

cli/src/lib.rs
CodeSandwich: This is the first step for a conventional, universal solution: https://github.com/radicle-dev/radicle-registry/pull/317#issuecomment-607072775. This way we can stop caring about what's inside and stop worrying that we haven't dug deep enough in our `Display` to tell user what's wrong. Other library authors follow this convention too and they don't display their `source`s in `Display`s, we, as final binary authors, mast take care of this.

NunoAlexandre: > This way we can stop caring about what's inside and stop worrying that we haven't dug deep enough in our Display

That's not a worry I have at all 🤷‍♂ 

> Other library authors follow this convention too and they don't display their sources in Displays,

In this case, this is our library and our binary. We gain nothing from the separation, or do we? How do we bring value to users by following this convention/approach?

CodeSandwich: > That's not a worry I have at all man_shrugging

It should! The users may end up with a generic `Client has failed` error and not the real root cause (e.g. timeout error burried deep in HTTP library error)

> In this case, this is our library and our binary. We gain nothing from the separation, or do we? How do we bring value to users by following this convention/approach?

We gain universal clarity. When both `Serde` errors and our errors follow the same convention, it's easy to properly process errors. Theoretically we could ducktape a decent user experience without following any conventions, but at what cost?

NunoAlexandre: Let's see if we get this right. We are not understanding each other.

> You should! The users may end up with a generic Client has failed error and not the real root cause (e.g. a timeout error burried deep in an HTTP library error)

I absolutely care that users get useful error messages, that is not what I said. I am just not " worrying that we haven't dug deep enough in our Display to tell user what's wrong".

Now, I am quite sure I am not following your suggestion.
What I get you saying is that when implementing Display, we shouldn't print the inner error because of some convention since it's so easy to forget and that it will fill us with worry. And that the way then is to have something like:

```` rust
#[error("data store disconnected")]
Disconnect(#[from] io::Error),
```

instead of:

```` rust
#[error("data store disconnected: {0}")]
Disconnect(#[from] io::Error),
```

My question then was, how do we also show the inner io::Error (in this example)? What I understand from you is that we should have a dedicated function in `main` that handles the `CommandError`, matches each variant as finally displays the whole, complete error message where it unwraps the `source` error too? That would mean that we'd have to ways to print the error, via print that uses the Display impl or via that second function. If that is the suggestion, I fail to see how that helps with those extreme worries or how it provides "universal clarity".

The way this Display impl works now (on master) provides exactly the functionality we need. This PR simplify offers the same but using `thiserror`. Now, I am new to Rust, so I am unfamiliar with most conventions and idealism, but I would like to focus on delivering palpable value to the project and our users.

Could you help me understand or correct me if I am not getting your point and suggestion? An example would help a lot.

CodeSandwich: Take a closer look at std error: it derives from `Display`. Anything that implements `Error` can be printed, even if used as `&dyn std::error::Error`. Whatever is returned from `source`, can be both printed and provided a deeper layer `source`. There's no need to match on `CommandError`, it implements `std::error::Error`.

CodeSandwich: Take a look at this example: https://github.com/input-output-hk/jormungandr/blob/master/jcli/src/main.rs

NunoAlexandre: Thanks for the link, we are now on the same page. So the idea is that each error (variant) prints only itself, even it has inner errors (which can go arbitrarily deep), and that then developers should have such a function as in that example that will make sure that all errors, at all levels, are printed.

While I see the value in doing that (now I understand your worry about forgetting to print nested errors), it seems odd to me that each Rust project needs to have this function every time; shouldn't;t this be native? Or would that be a bad idea since it would be assuming and potentially get in the way?

CodeSandwich: > it seems odd to me that each Rust project needs to have this function every time; shouldn't;t this be native?

It will: https://doc.rust-lang.org/std/error/struct.Chain.html

cli/src/lib.rs
NunoAlexandre: Tbh, I just looked at `thiserror`'s docs (https://docs.rs/thiserror/1.0.14/thiserror/) and it referencing to the variant's fields in this fashion even for `String`s so I thought it was a matter of syntax and not of printing debug.

cli/src/lib.rs
NunoAlexandre: @geigerzaehler could you have a look? I added `FailedWriting` and `FailedRead` variants that wrap themselves variants for IO and json (de)serialization use cases. We don't benefit from having this extra sub error type but how else would we have a generic error type on the high-level variants?

I.e, Now we have:

``` rust
#[error("Failed to write to the accounts file: {0}")]
FailedWrite(WritingError),

 #[error("Failed to read the accounts file: {0}")]
 FailedRead(ReadingError),
```

where

``` rust
#[derive(Debug, ThisError)]
pub enum WritingError {
    #[error("{0}")]
    IO(IOError),

    #[error("{0}")]
    Serialization(serde_json::Error),
}

/// Possible errors when reading the accounts file.
#[derive(Debug, ThisError)]
pub enum ReadingError {
    #[error("{0}")]
    IO(IOError),

    #[error("{0}")]
    Deserialization(serde_json::Error),
}
```

whereas I'd prefer to just have:

``` rust
#[error("Failed to write to the accounts file: {0}")]
FailedWrite(AnyError),

 #[error("Failed to read the accounts file: {0}")]
 FailedRead(AnyError),
```

`io::Error` and `serde_json::Error` don't implement `std::error:Error`. Is there another way, besides making it:

``` rust
#[error("Failed to write to the accounts file: {0}")]
FailedWrite(String),

 #[error("Failed to read the accounts file: {0}")]
 FailedRead(String),
```

and format the actual error to `String` to when building `FailedWrite` and `FailedRead`?

cli/src/lib.rs
NunoAlexandre: Cool. This seems like something we should discuss cross-teams since it's quite an important matter that should not just be done in the cli of a single project. I prefer if we focus on the remaining issues of this PR and have this discussion in `radicle-decisions` and then implement that here and everywhere else.

cli/src/lib.rs
CodeSandwich: I think that `thiserror` simply copy-pastes this annotation into `Display`'s `format!`, so it's a regular Rust formatting grammar.

cli/src/lib.rs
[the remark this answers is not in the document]
NunoAlexandre: Done, very neat :ok_hand:

cli/src/lib.rs
NunoAlexandre: Fixed.

cli/src/lib.rs
CodeSandwich: > io::Error and serde_json::Error don't implement std::error:Error

I've just checked the docs and both of them do :thinking:

NunoAlexandre: Oh, can you pointed me there? https://docs.serde.rs/serde_json/struct.Error.html doesn't show any std::error::Error implementation

CodeSandwich: I can't directly link to `impl`s, so I've made a screenshot

image

NunoAlexandre: The problem with ambiguity again... Anyway, now that I have implemented `from` using `thiserror` and creating these new error variants is much cleaner I'd leave it as is. If you don't oppose, I will merge.

CodeSandwich: You're right, it DOES improve the code down the line. It turned out to be a good design.

cli/src/account_storage.rs
```diff
@@ -39,17 +39,37 @@ type Seed = [u8; 32];
...
+}
+
+/// Possible errors when reading the accounts file.
+#[derive(Debug, ThisError)]
+pub enum ReadingError {
+    #[error("{0}")]
+    IO(IOError),
+
+    #[error("{0}")]
+    Deserialization(serde_json::Error),
```
CodeSandwich, line 72: This looks like a perfect use case for `#[error(transparent)]`. It would be great to have either `#[source]` of `#[from]` here, this way this error will be able to provide its source.

NunoAlexandre: For reference, I found this on what `#[error(transparent)]` is: https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362/2. Done.

cli/src/account_storage.rs
```diff
@@ -39,17 +39,37 @@ type Seed = [u8; 32];
...
+    /// Failed to write to the accounts file
+    #[error("Failed to write to the accounts file: {0}")]
+    FailedWrite(#[from] WritingError),
 
-    /// A serde json Error occurred
-    #[error("A serde json error occured: {0}")]
-    JsonError(serde_json::Error),
+    /// Failed to read the accounts file
+    #[error("Failed to read the accounts file: {0}")]
+    FailedRead(#[from] ReadingError),
```
CodeSandwich, line 52: https://github.com/radicle-dev/radicle-registry/pull/317#discussion_r402366809 :grimacing:

cli/src/lib.rs
kim: @CodeSandwich this is new to me, too. Would you mind submitting an amendment to radicle-dev/radicle-decisions#2? I am particularly interested in how this pans out for wrapper variants (`Io(io::Error)`) only introduced for unification.

cli/src/account_storage.rs
NunoAlexandre: ? That's addressed.

CodeSandwich: This is exactly what dtolnay advices against. We both print the inner error in a message and return it from `source` to print it again.

NunoAlexandre: @CodeSandwich can we move this discussion to radicle-decisions https://github.com/radicle-dev/radicle-registry/pull/317#discussion_r402372444?

PR #321, opened 2020-04-02
Cli/fix leaked abstraction (corrected)

NunoAlexandre:
Fix lost changes from https://github.com/radicle-dev/radicle-registry/pull/315.

### DIFF
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -19,7 +19,7 @@ use super::*;
 use crate::account_storage;
 
 /// Account related commands
-#[derive(StructOpt, Debug, Clone)]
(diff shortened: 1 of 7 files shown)


cli/src/command/account.rs
```diff
@@ -19,7 +19,7 @@ use super::*;
 use crate::account_storage;
 
 /// Account related commands
-#[derive(StructOpt, Debug, Clone)]
```
CodeSandwich, line 22: Why all the `Debug`s everywhere are dropped?

NunoAlexandre: See https://github.com/radicle-dev/radicle-registry/pull/315#discussion_r400982628

PR #322, opened 2020-04-02
Add header fetching to client API

### LINKED ISSUE #320: Client API for block headers
The client API must make it possible to:
* Get block header by hash
* Get best block header

CodeSandwich:
Fixes #320

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,3 +54,4 @@ Upcoming
   * `ClientT::genesis_hash()`
   * `Transaction`
   * `Transaction::new_signed()`
+* Add block header fetching to the client API
--- a/client/src/backend/mod.rs
+++ b/client/src/backend/mod.rs
@@ -16,7 +16,7 @@
 //! Define trait for client backends and provide emulator and remote node implementation
 use futures::future::BoxFuture;
 
-pub use radicle_registry_runtime::{Hash, UncheckedExtrinsic};
+pub use radicle_registry_runtime::{Hash, Header, UncheckedExtrinsic};
 
 use crate::interface::*;
 
@@ -65,6 +65,10 @@ pub trait Backend {
         prefix: &[u8],
         block_hash: Option<BlockHash>,
     ) -> Result<Vec<Vec<u8>>, Error>;
+
+    /// Fetch the header of the given block hash.
+    /// If the block hash is `None`, fetch the header of the best chain tip.
+    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<Header, Error>;
 
     /// Get the genesis hash of the blockchain. This must be obtained on backend creation.
     fn get_genesis_hash(&self) -> Hash;
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -197,6 +197,18 @@ impl backend::Backend for RemoteNode {
         Ok(keys.into_iter().map(|key| key.0).collect())
     }
 
+    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<BlockHeader, Error> {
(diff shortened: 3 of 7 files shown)


client/src/backend/remote_node.rs
```diff
@@ -197,6 +197,18 @@ impl backend::Backend for RemoteNode {
         Ok(keys.into_iter().map(|key| key.0).collect())
     }
 
+    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<Header, Error> {
```
NunoAlexandre, line 200: The problem of using type alias is that it's never enforced, creating inconsistencies that leads to cases like this.
This to say:

```suggestion
    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<BlockHeader, Error> {
```

client/src/interface.rs
```diff
@@ -111,6 +115,10 @@ pub trait ClientT {
         account_id: &AccountId,
     ) -> Result<state::AccountTransactionIndex, Error>;
 
+    /// Fetch the header of the given block hash.
+    /// If block hash is none, fetch the header of the best chain tip.
```
NunoAlexandre, line 119: nit:

```suggestion
    /// If the block hash is `None`, fetch the header of the best chain tip.
```

client/src/backend/mod.rs
```diff
@@ -66,6 +66,10 @@ pub trait Backend {
         block_hash: Option<BlockHash>,
     ) -> Result<Vec<Vec<u8>>, Error>;
 
+    /// Fetch the header of the given block hash.
+    /// If block hash is none, fetch the header of the best chain tip.
+    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<Header, Error>;
```
NunoAlexandre, line 71: suggestion: Instead of coercing the meaning "`None` means x", have a type describing that:

``` rust
enum GetBlockHeader {
    Hash(BlockHash),
    BestChainTip,
}
```

I see that the chain operates on `Option<BlockChain>` with that meaning builtin, so leave it up to you. But this can be seen as xla's remark that substrate shouldn't dictate our api.

CodeSandwich: I don't like putting special meaning into `Option`s either. I was also considering creating two separate methods: `.block_header(block_hash)` and `.block_header_best_chain()` to keep it as simple as possible.

client/src/backend/remote_node.rs
CodeSandwich: Fixed

client/src/interface.rs
CodeSandwich: Fixed

client/src/backend/mod.rs
NunoAlexandre: I find it worth implementing either of the two but leave it up to you :+1: Will approve the PR since everything else is addressed.

CHANGELOG.md
```diff
@@ -53,3 +53,4 @@ Upcoming
   * `ClientT::genesis_hash()`
   * `Transaction`
   * `Transaction::new_signed()`
+* Add header fetching to client API
```
NunoAlexandre, line 56: sorry, one last thing:

```suggestion
* Add block header fetching to the client API
```

client/src/backend/mod.rs
geigerzaehler: I’d prefer the two methods, too. The methods become self-explanatory then

CodeSandwich: Fixed, split into 2 methods

CHANGELOG.md
CodeSandwich: Fixed

PR #326, opened 2020-04-06
Client/fee validity

### LINKED ISSUE #324: Add ability to query cost bounds for safe tranaction submission
Note that we could add a query to the registry client giving us here that minimum fee value.

_Originally posted by @NunoAlexandre in https://github.com/radicle-dev/radicle-upstream/pull/272/files_

As a consumer of the Registry client there should be the ability to validate, confirm that fees meeting the minimum configured values and that deposits are matched (once introduced).

NunoAlexandre:
Closes #324

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Upcoming
   mines if the option is given.
 
 ### Addition
+* Add `MINIMUM_FEE` to the registry client
 * The client emulator now authors blocks when a transaction is submitted.
 * Add `TransferFromOrg` message
 * Add `Client::get_org` and `Client::list_orgs`
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -47,6 +47,7 @@ mod transaction;
 
 pub use crate::interface::*;
 pub use radicle_registry_core::Balance;
+pub use radicle_registry_runtime::fees::MINIMUM_FEE;
 
 pub use backend::EMULATOR_BLOCK_AUTHOR;
 
--- a/runtime/src/fees/mod.rs
+++ b/runtime/src/fees/mod.rs
@@ -26,9 +26,9 @@ use sp_runtime::transaction_validity::{
 
 mod payment;
 
-/// The base fee serves as a disincentive to stop bad actors
-/// from spamming the network in a DoS attack.
-const MINIMUM_FEE: Balance = 1;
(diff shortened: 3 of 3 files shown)


client/src/lib.rs
```diff
@@ -318,4 +323,18 @@ mod test {
         fn is_sync_send(_x: impl Sync + Send + 'static) {}
         is_sync_send(Client::new_emulator());
     }
+
+    #[async_std::test]
```
CodeSandwich, line 327: They don't need to be async

NunoAlexandre:
@geigerzaehler @CodeSandwich apologies for the noise but can you have another look? I have changed the approach to expose the `MINIMUM_FEE`. This way, users of the client can **both** know the minimum they need to pay and/or verify whether a given amount is sufficient, which better serves the needs of the upstream.

CodeSandwich:
On one hand this simplification makes sense, but on the other we're pushing the fee validation logic to the client. Now it's trivial, but it probably will get much more complicated in the future. How about 2 methods: `mininum_fee` and `validate_fee`?

NunoAlexandre:
@CodeSandwich I agree, thus my hesitation. There are no plans to make this any more complex in the future, so I'd suggest us to go this way. Let's see what @xla has to say as a consumer.

runtime/src/fees/mod.rs
```diff
@@ -28,7 +28,7 @@ mod payment;
 
 /// The base fee serves as a disincentive to stop bad actors
 /// from spamming the network in a DoS attack.
-const MINIMUM_FEE: Balance = 1;
+pub const MINIMUM_FEE: Balance = 1;
```
geigerzaehler, line 31: Could you please update the documentation so that is understandable for users of the client. I’d suggest just saying “The minimum fee required for a transaction to be considered for inclusion in the block chain.” I’d drop the “why”.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Upcoming
   mines if the option is given.
 
 ### Addition
+* Add 'Client::fee_is_sufficient' to verify fee validity.
```
geigerzaehler, line 47: This needs to be updated, too.

NunoAlexandre: Good catch. Note to self: don't multitask, man.

Issue #291, opened 2020-03-26
Add FFnet placeholder chain spec


geigerzaehler:
We want to add a placeholder chain spec for the FFnet. The goal is to have the chain spec in place and solve the problem with runtime code explained below.

### Static Genesis Config

The genesis config includes the runtime wasm code. If the runtime code for the ffnet genesis config is build from source it is not guaranteed that the code stays the same. This results in a node built from source using a different genesis block. To avoid this we need to check the gensis config into the code and include it statically in the node binary.

See also [https://substrate.dev/docs/en/next/development/deployment/chain-spec](https://substrate.dev/docs/en/next/development/deployment/chain-spec)

geigerzaehler:
It should also be possible to use a build environment variable to make the ffnet chain spec the default for the node binary. Meaning that if the build is run with say `DEFAULT_CHAIN=ffnet` then `radicle-registry-node` connects to the `ffnet` chain

PR #327, opened 2020-04-06
Add deterministic runtime WASM building

CodeSandwich:
The first step to https://github.com/radicle-dev/radicle-registry/issues/291.


node/src/chain_spec.rs
```diff
@@ -73,7 +73,7 @@ fn dev_genesis_config() -> GenesisConfig {
     let root_key = get_from_seed::<AccountId>("Alice");
     GenesisConfig {
         system: Some(SystemConfig {
-            code: WASM_BINARY.to_vec(),
+            code: GENESIS_RUNTIME_WASM.to_vec(),
```
geigerzaehler, line 76: This variable is only used in this module. Is there a reason to expose it via `radicle_registry_runtime`? If not we could just do `include_bytes!` here.

runtime/build.rs
```diff
@@ -1,21 +1,66 @@
...
     WasmBuilder::new()
         .with_current_project()
         .with_wasm_builder_from_crates("1.0.9")
         .export_heap_base()
         .import_memory()
-        .build()
+        .build();
+}
+
+fn build_genesis_wasm_rs_file(wasm_binary_dir: &Path) {
```
geigerzaehler, line 47: Is it necessary to have this function? It only generates this one file dynamically which is already checked into source control.

DEVELOPING.md
```diff
@@ -117,6 +117,25 @@ To update the revision of substrate run
...
+.scripts/build-genesis-runtime-wasm
+~~~
+Remember that this operation should **never** be done after starting the public network!
+It will make the node incompatible with any older nodes on the network.
+
+### Building a runtime for an update
+
+To build a WASM runtime binary meant for packing into a transaction and broadcasting on the network run
+~~~
+.scripts/build-runtime-wasm OUT
```
geigerzaehler, line 136: Two things would make this command more ergonomic.
* Accept a relative path
* Make `OUT` the name of the Wasm blob and not the containing directory. Since we only generate one file we don’t need the extra indirection.

node/src/chain_spec.rs
CodeSandwich: The only reason was encapsulation, I wanted to keep it a strictly `runtime`'s problem.

Fixed.

runtime/build.rs
CodeSandwich: It was encapsulating knowledge about the physical WASM binary name and location in the build script.

Fixed, simplified code quite a bit.

DEVELOPING.md
CodeSandwich: It's possible to set only the output directory, but not the output file name in the WASM builder. In order to change that, the output file must be renamed  after it's created. Unfortunately the WASM binary is built only after the the `build.rs` is finished, `WasmBuilder::build` only schedules the work!

Fixed, pushed most of the logic to the script. Now the name of the genesis binary is hardcoded in two places, but it simplified the code even further.

DEVELOPING.md
```diff
@@ -117,6 +117,25 @@ To update the revision of substrate run
...
 where `REV` is the new Git revision SHA.
 
+Building a WASM runtime binary
+------------------------------
+
+### Building a runtime for the genesis
+
+To build a genesis WASM runtime binary run
+~~~
+.scripts/build-genesis-runtime-wasm
```
geigerzaehler, line 127:
```suggestion
./scripts/build-genesis-runtime-wasm
```
See also below.

DEVELOPING.md
```diff
@@ -117,6 +117,25 @@ To update the revision of substrate run
 ~~~
 where `REV` is the new Git revision SHA.
 
+Building a WASM runtime binary
+------------------------------
+
+### Building a runtime for the genesis
+
+To build a genesis WASM runtime binary run
```
geigerzaehler, line 125: Could you also explain which file is updated and how the file is used?

DEVELOPING.md
CodeSandwich: Fixed

DEVELOPING.md
CodeSandwich: Fixed

DEVELOPING.md
```diff
@@ -117,6 +117,28 @@ To update the revision of substrate run
...
+It will be then used as the genesis WASM runtime in all consecutive compilations of the node.
+
+Remember that this operation should **never** be done after starting the public network!
+It will make the node incompatible with any older nodes on the network.
+
+### Building a runtime for an update
+
+To build a WASM runtime binary meant for packing into a transaction and broadcasting on the network run
+~~~
+./scripts/build-runtime-wasm OUT
```
NunoAlexandre, line 139: better use the gnu syntax for cli args: 
```suggestion
./scripts/build-runtime-wasm <output_file>
```

runtime/build.rs
```diff
@@ -1,21 +1,22 @@
...
-            std::env::set_var(DUMMY_WASM_BINARY_ENV, "1");
-        }
+    if let Some(wasm_binary_dir) = env::var_os(BUILD_WASM_BINARY_OUR_DIR_ENV) {
+        build_wasm_runtime_in_dir(&wasm_binary_dir);
     }
+}
 
+fn build_wasm_runtime_in_dir(out_dir: impl AsRef<OsStr>) {
+    env::set_var("WASM_TARGET_DIRECTORY", out_dir);
+    env::set_var("WASM_BUILD_TYPE", "release");
```
NunoAlexandre, line 15: So we have code setting environment variables? Shouldn't we consume them instead? Can't this go wrong or cause silent misbehaviour given the side effects?

runtime/src/genesis_runtime_wasm/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! Do not edit! This file is generated automatically in the build script.
+
+pub const GENESIS_RUNTIME_WASM: &[u8] = include_bytes!("radicle_registry_runtime.wasm");
```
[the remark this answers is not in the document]
CodeSandwich: They aren't discontinued, they just became optional. Anyway this file doesn't exist anymore :wink:

DEVELOPING.md
CodeSandwich: Good point, fixed, I've also fixed another place in this document, where the gnu syntax wasn't followed.

runtime/build.rs
CodeSandwich: That's the official and the only way to pass these parameters to the WASM builder, so it's not up to us.

runtime/src/genesis_runtime_wasm/mod.rs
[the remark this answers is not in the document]
NunoAlexandre: I had written that comment when you open the PR :P

runtime/build.rs
NunoAlexandre: I see I see, ok!

PR #332, opened 2020-04-08
Don’t set logging filters in dev node script

geigerzaehler:
We don’t want to filter the logs by default for the dev node. In particular, we want to see the “Imported #XX” logs.


geigerzaehler:
> What's the default log level, that's going to have an effect when we remove these flags?

The default level is `info`. This change also makes this the default level for the two `substrate_consensus_slots` and `substrate` targets.

PR #334, opened 2020-04-08
Improved runtime API interface

geigerzaehler:
Before, the emulator was duplicating code from the runtime API implementations. We now offer this code from a `runtime_api` module from the runtime. We also move everything related to the Substrate runtime implementation (that is the `impl_runtime_apis!` call ) to this module to improve encapsulation.

### DIFF
--- a/client/src/backend/emulator.rs
+++ b/client/src/backend/emulator.rs
@@ -23,7 +23,7 @@ use sp_runtime::{traits::Hash as _, BuildStorage as _, Digest};
 use sp_state_machine::backend::Backend as _;
 
 use radicle_registry_runtime::{
-    registry, AccountId, BalancesConfig, Executive, GenesisConfig, Hash, Hashing, Header, Runtime,
+    registry, runtime_api, AccountId, BalancesConfig, GenesisConfig, Hash, Hashing, Header, Runtime,
 };
 
 use crate::backend;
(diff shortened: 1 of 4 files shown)


runtime/src/runtime_api.rs
```diff
@@ -0,0 +1,168 @@
...
+//! Implements Substrate runtime APIs and provide a function based interface for the runtime APIs.
+use frame_support::{ensure, fail, traits::Randomness};
+use sp_core::OpaqueMetadata;
+use sp_runtime::traits::Block as BlockT;
+use sp_runtime::{
+    transaction_validity::{InvalidTransaction, TransactionSource, TransactionValidity},
+    ApplyExtrinsicResult,
+};
+
+use super::*;
```
CodeSandwich, line 25: How hard would it be to import things manually? It's often painful to track dependencies when wildcards are in use.

client/src/backend/emulator.rs
```diff
@@ -104,25 +104,25 @@ impl backend::Backend for Emulator {
         let mut state = self.state.lock().unwrap();
         let next_header = state.next_header.clone();
         let (header, events) = state.test_ext.execute_with(move || {
-            Executive::initialize_block(&next_header);
+            runtime_api::initialize_block(&next_header);
```
CodeSandwich, line 107: Are these equivalent?

geigerzaehler: Yes, they are. The new method just provides better encapsulation and does not need the `Executive`.

PR #342, opened 2020-04-09
developing: Make a release

### LINKED ISSUE #319: Provide FFnet binaries
For users to participate in the FFnet we need to provide them with the node and CLI binaries.

* The binaries should be configured for the ffnet by default
  * The default chain for the node should be the ffnet
  * The CLI should connect to one of our full nodes by default
* The location of the binaries should be documented and discoverable so that we can share it in the getting started guide.
* It should be easy (and documented) to release a new version of the binaries

## Open questions
* Should we automatically release the latest master or should we create manual releases?

## Implementation
* Github releases with a moving `ffnet` tag? Would need to automate this with buildkite.
* Buildkite artifcats? The problem is that builds are addressed by commit at best and UUIDs at worst. This means moving releases are not possible
* GCS storage bucket as an alternative to Github releases. More lightweight

## Related

* Depends on #291 
* See also #318

NunoAlexandre:
Closes #319 in respect to how we make the binaries available.

### Decisions

- We are not automating the releases using the CLI since it's a one-minute manual step, sufficient for the stage we are in
- Automating Github releases from Buildkite requires wide write permissions to the project which we don't want
- We prefer to have the binaries available via GitHub releases on the project itself instead of having them hosted somewhere else.

This PR adds a `Make a release` section to `DEVELOPING.MD`, explaining the steps.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -14,6 +14,7 @@ The code is bootstrapped with the [`substrate-node-template`][node-template].
 - [Packages](#packages)
 - [Running checks and tests](#running-checks-and-tests)
 - [Continuous Deployment](#continuous-deployment)
+- [Make a release](#make-a-release)
 - [Local devnet](#local-devnet)
 - [Updating substrate](#updating-substrate)
 - [Updating Continuous Integration's base Docker image](#updating-continuous-integrations-base-docker-image)
@@ -99,6 +100,31 @@ Continuous Deployment
 
 We’re continuously deploying master builds of the node to the devnet. We’re
 using Kubernetes for this. You can find the code in `ci/deploy`.
+
+
+Make a release
+--------------
+
+1. Download the latest binaries for the targeted master build [in Builkite](https://buildkite.com/monadic/radicle-registry/builds?branch=master)
+    - Select the targeted build
+    - Expand the `Test ci/run` section
+    - Open the `Artifacts` tab
+    - Download the following binaries:
+        - `radicle-registry-cli`
+        - `radicle-registry-node`
+2. Create a new release
+    - Open the [Radicle Registry Github Releases page](https://github.com/radicle-dev/radicle-registry/releases)
+    - Define the tag version (see [Tags](#tags))
+    - Select the right target
+    - Add a release title
+    - Describe the changes included in this release to users
+    - Attach the binaries downloaded in step 1.
+
+### Tags
+
+Our tags are composed of the release date followed by an incremental number. For example:
+
+- `2020-04-09-0`
 
 
 Local devnet
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -122,9 +122,10 @@ Make a release
 
 ### Tags
 
-Our tags are composed of the release date followed by an incremental number. For example:
-
-- `2020-04-09-0`
+Our tags are composed of the release date followed by a number representing the
+number of releases done in that same date. For example, `2020-04-09-0` would
+be the first release done that day. Should we need to make a new release in
+that same day, it'd be tagged `2020-04-09-1`.
 
 
 Local devnet


DEVELOPING.md
```diff
@@ -101,6 +102,27 @@ We’re continuously deploying master builds of the node to the devnet. We’re
 using Kubernetes for this. You can find the code in `ci/deploy`.
 
 
+Make a release
+--------------
+
+1. Download the latest binaries for the targeted master build [in Builkite](https://buildkite.com/monadic/radicle-registry/builds?branch=master)
```
geigerzaehler, line 108: Could we use the term “artifacts” and describe how to get them in more detail? (If I remember correctly you actually struggled to find them.) We should also make the list of artifacts explicit. I.e. the node and the CLI. If we add more artifacts for development purposes, say, we don’t include them in the release.

DEVELOPING.md
```diff
@@ -101,6 +102,27 @@ We’re continuously deploying master builds of the node to the devnet. We’re
...
+    - Describe the changes included in this release to users
+    - Attach the binaries downloaded in step 1.
+
+### Tags
+
+We are currently **not** in production. Therefore, the tags reflect that by
+including the respective phases as a suffix to follow the [semVer specification](https://semver.org/).
+
+- `v0.X-ffnet`
+- `v0.X-devnet`
```
geigerzaehler, line 123: We are not doing devnet releases and we don’t have a need for them so I’d drop this.

DEVELOPING.md
```diff
@@ -101,6 +102,27 @@ We’re continuously deploying master builds of the node to the devnet. We’re
...
+--------------
+
+1. Download the latest binaries for the targeted master build [in Builkite](https://buildkite.com/monadic/radicle-registry/builds?branch=master)
+2. Open the [Radicle Registry Github Releases page](https://github.com/radicle-dev/radicle-registry/releases)
+3. Create a new release
+    - Define the tag version (see [Tags](#tags))
+    - Select the right target
+    - Add a release title
+    - Describe the changes included in this release to users
+    - Attach the binaries downloaded in step 1.
```
geigerzaehler, line 115: Suggestion: There are CLI tools ([hub](https://hub.github.com/) and the [official one](https://cli.github.com/)) that can do this. I’m ok with describing the steps for the UI but if you feel like it I’d prefer to describe this with the tools. This makes it more easily reproducible and it can then be automated.

DEVELOPING.md
```diff
@@ -101,6 +102,27 @@ We’re continuously deploying master builds of the node to the devnet. We’re
...
+
+### Tags
+
+We are currently **not** in production. Therefore, the tags reflect that by
+including the respective phases as a suffix to follow the [semVer specification](https://semver.org/).
+
+- `v0.X-ffnet`
```
geigerzaehler, line 122: Suggestion: What do you think about using the date as the tag. E.g. `v2020-04-09.0`. This would avoid the confusion with SemVer and one can easily see how old there version is.

geigerzaehler:
Although this is out of scope I would like to raise the question of _when_ to release? My suggestion would be release once a week on a fixed day.

Related is the discussion about semver. While the scheme is great for libraries it fails for applications. For example it is harder to determine what breaking changes, added features, and patches are. Most user facing software does not follow semver.

DEVELOPING.md
NunoAlexandre: Regarding your comment below

> Related is the discussion about semver. While the scheme is great for libraries it fails for applications. For example it is harder to determine what breaking changes, added features, and patches are. Most user facing software does not follow semver.

How does using a date help with that? I am fine with using a date, although I also like having `ffnet` in the tag. @CodeSandwich what would you prefer? I'm open for either one.

NunoAlexandre:
> Although this is out of scope I would like to raise the question of when to release? My suggestion would be release once a week on a fixed day.

Good point. Some follow up questions on that:

- Do we want users to be notified of a new version?
- Do we want to log anything? Like occurred errors, etc?
- Shouldn't we change the version in the Cargo.toml file for the binaries?

DEVELOPING.md
NunoAlexandre: > Could we use the term “artifacts” 

Do you mean to use "artifacts" instead of "binaries"?

> and describe how to get them in more detail? (If I remember correctly you actually struggled to find them.)

That's true, and sure will do!

and describe how to get them in more detail? (If I remember correctly you actually struggled to find them.)

> We should also make the list of artifacts explicit. I.e. the node and the CLI. If we add more artifacts for development purposes, say, we don’t include them in the release.|

:+1:

DEVELOPING.md
NunoAlexandre: This shouldn't indeed be a blocker but I agree with the reproducibility argument. Will have a look at those CLI tools :+1:

DEVELOPING.md
CodeSandwich: For the development purposes date seems more useful. We can switch to semver when we stabilize. Either way is fine by me, it's not that crucial as long as we follow any convention.

geigerzaehler: > How does using a date help with that?

The version just doesn’t look like a semver version so there is also no expectation for it to follow semver. But this is just a minor point. Keeping `ffnet` might make sense since we might have different builds depending on the network. Although we don’t need it right now I’m ok with keeping it.

geigerzaehler:
> * Do we want users to be notified of a new version?

Yes. If you use Github releases they can subscribe to the release on the page. We can also post something on radicle.community. On the other hand, if we tell users we release weekly we don’t really need notifications.

> * Do we want to log anything? Like occurred errors, etc?

I don’t follow. What process logs things?


>  * Shouldn't we change the version in the Cargo.toml file for the binaries?

Good point. We also have the version built into the binary. (It’s `ffnet.0`.) I’d suggest not touching the `Cargo.toml` version. Since we’re not publishing packages it does not really have a meaning and this would just create churn.

NunoAlexandre:
> I don’t follow. What process logs things?

None at the moment, that's my point. I'd prefer to catch panics and errors instead of waiting for users to come to us with them. We could setup [sentry.io](https://sentry.io), for example. It isn't a blocker, but it would surely be good to have at some point.

NunoAlexandre:
> We also have the version built into the binary. (It’s ffnet.0.)

@geigerzaehler not sure if get what you mean. Are you suggesting renaming the binary to like `radicle-registry-cli-ffnet-0`? That would be weird in action: `radicle-registry-cli-ffnet-0 org register`, for example. But maybe I just didn't get it?

DEVELOPING.md
NunoAlexandre: The official tool doesn't seem to offer that command, `hub` does. I'd prefer to keep this as is, to avoid introducing another tool to the table. We will be releasing often so shall anything change, we will be in sync rapidly.

DEVELOPING.md
NunoAlexandre: We can always have both, right? `v2020-04-09-ffnet-X`. It's long but makes everyone happy.

NunoAlexandre:
@geigerzaehler as for the scope of this PR, we have two discussions pending:

- [Using the term "artifacts"](https://github.com/radicle-dev/radicle-registry/pull/342#discussion_r406132199)

- [Tag convention](https://github.com/radicle-dev/radicle-registry/pull/342#discussion_r406191100)

DEVELOPING.md
geigerzaehler: > > Could we use the term “artifacts”
> 
> Do you mean to use "artifacts" instead of "binaries"?

Yes. But I don’t think this is necessary anymore because we explain in detail how to get the artifacts. I was just worried about being able to find them in the UI.

NunoAlexandre: I agree :+1:

DEVELOPING.md
geigerzaehler: > We can always have both, right? `v2020-04-09-ffnet-X`. It's long but makes everyone happy.

For simplicity we should choose one scheme that best satisfies our needs and is easy to understand. I think we should choose one or the other. I’m fine with either but would prefer the date.

geigerzaehler:
> > I don’t follow. What process logs things?
> 
> None at the moment, that's my point. I'd prefer to catch panics and errors instead of waiting for users to come to us with them. We could setup [sentry.io](https://sentry.io), for example. It isn't a blocker, but it would surely be good to have at some point.

Ah ok. So you want to get crash reports from users running the node. Haven’t thought about this either. It’s not directly related to releases but let’s keep it in mind.

> @geigerzaehler not sure if get what you mean. Are you suggesting renaming the binary to like `radicle-registry-cli-ffnet-0`? That would be weird in action: `radicle-registry-cli-ffnet-0 org register`, for example. But maybe I just didn't get it?

Sorry, I wasn’t clear: If you run the command it prints `version ff.0-0f67c46-x86_64-linux-gnu`. This might be confusing. The code is here:

https://github.com/radicle-dev/radicle-registry/blob/2138e4907b90040dbc2256c92c711ee7b4be5afd/node/src/main.rs#L32

DEVELOPING.md
NunoAlexandre: I would prefer the `ffnet` alternative since it's easier to talk about the `ffnet-X` release than the `yyyy-mm-dd-X` release, which, in a year time will become unclear as to which phase it related to, requiring a lookup. But again, I'm open. @CodeSandwich you decide, or we [toss the coin](https://justflipacoin.com/).

CodeSandwich: @NunoAlexandre I didn't change my opinion, I'm still slightly preferring the date. I think that having two interlocking numbering systems is complex and doesn't bring much information.

NunoAlexandre:
@CodeSandwich @geigerzaehler could you have another look at the `Tags` section?

DEVELOPING.md
```diff
@@ -101,6 +102,31 @@ We’re continuously deploying master builds of the node to the devnet. We’re
...
+    - Open the [Radicle Registry Github Releases page](https://github.com/radicle-dev/radicle-registry/releases)
+    - Define the tag version (see [Tags](#tags))
+    - Select the right target
+    - Add a release title
+    - Describe the changes included in this release to users
+    - Attach the binaries downloaded in step 1.
+
+### Tags
+
+Our tags are composed of the release date followed by an incremental number. For example:
```
CodeSandwich, line 125: I want to make sure that we're all on the same side: the number in the suffix is zero by default and only incremented if there are more than one releases during one day?

NunoAlexandre: That's how I saw it, altough that now reading the text it might give the idea that it's a global incremental number.

NunoAlexandre: disambiguated :heavy_check_mark:

PR #344, opened 2020-04-09
Send notification when master build fails

geigerzaehler:
Send notification when master build fails.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -31,3 +31,8 @@ steps:
     agents:
       platform: "linux"
       production: "true"
+
+notify:
+  - email: "registry-devs@monadic.xyz"
+    if: |
+      build.state == "failed" && build.branch == "master"


.buildkite/pipeline.yaml
```diff
@@ -31,3 +31,8 @@ steps:
     agents:
       platform: "linux"
       production: "true"
+
+notify:
+  - email: "thomas@monadic.xyz"
```
NunoAlexandre, line 36: Can it be sent to the team email?

Is there an integration with irc? It'd be cool to get a message in our channel.

CodeSandwich: There's only integration with email and slack. Unless we can bridge an email address or a dummy slack account with IRC, we won't be getting notifications.

NunoAlexandre: Maybe can plug something using https://ifttt.com/. Using the team email would suffice for now though.

geigerzaehler: I don’t want to send it to `registry@monadic.xyz` to not spam the non-devs that are subscribed there. What do you think about creating a `registry-notification@monadic.xyz` group?

CodeSandwich: Cool! Or `registry-devs`. Can we see all the existing groups somewhere? Maybe such a group already exists.

geigerzaehler: Done

Issue #341, opened 2020-04-09
Make ffnet the default chain for released binaries


geigerzaehler:
We want users to be able to participate in the ffnet with our provided node binaries without additional configuration. To achieve this the node binaries that we release should have the ffnet as the default `--chain` option. This should also apply to binaries build from source via our instructions (see #318).

Depends on #291.

We need to consider that the development flow relies on the `devnet` being the default for now so we should account for that.

PR #347, opened 2020-04-09
Add ffnet chain and make it default

### LINKED ISSUE #291: Add FFnet placeholder chain spec

CodeSandwich:
Fixes #291 and #341.

This adds support for the ffnet and makes it the default in CI-built node binaries. This is achieved with an optional compilation env variable, which lets override the `--chain` default `dev` and set it to `ffnet`. It doesn't affect regular builds in any way, `dev` is still the default.


node/src/chain_spec.rs
```diff
@@ -123,33 +112,28 @@ pub fn local_devnet() -> ChainSpec {
...
+        pallet_sudo: Some(SudoConfig { key: sudo_key }),
     }
 }
 
-/// Helper function to generate a crypto pair from seed
-pub fn get_from_seed<TPublic: Public>(seed: &str) -> <TPublic::Pair as Pair>::Public {
-    TPublic::Pair::from_string(&format!("//{}", seed), None)
-        .expect("static values are valid; qed")
+/// Helper function to generate an account public key from a seed
+fn account_pub_key(seed: &str) -> <<AccountId as CryptoType>::Pair as Pair>::Public {
```
geigerzaehler, line 135: Or just 
```suggestion
fn account_pub_key(seed: &str) -> AccountId {
```

CodeSandwich: We don't want the whole AccountId, we want just the public key. It could be changed into `sp_core::ed25519::Public`, but then the tight relation with `AccountId` would become just a coincidence.

geigerzaehler: AFAIK, `AccountId` is an alias for `sp_core::ed25519::Public`

CodeSandwich: You're right, it is. The level of convolution here was hilarious :D Fixed.

CodeSandwich:
> Why have both DEFAULT_CHAIN and CHAIN?

I don't quite follow what is `CHAIN`, could you elaborate?

> Also, I thought that we wanted the ffnet the default chain overall but this PR only makes it default when running in the CI.

Not quite. We want the binaries we **release** to use the `ffnet` by default. For the devlopers it's better to keep the regular `dev` as a default, so they don't need to remember to pass it every single time. The CI builds our release binaries for the users and that's why it's the only place, where the `ffnet` is set.

NunoAlexandre:
I overlooked your solution, sorry for the noise. I was not seeing that DEFAULT_CHAIN is provided at build time while CHAIN(the Chain CLI option) can only be provided at execution time. Approving ✌️

Issue #348, opened 2020-04-10
Fix "watch_extrinsic stream terminated" on failed fee payment


NunoAlexandre:
When submitting a transaction from the CLI using an account with insufficient funds or with an invalid fee, it fails with `watch_extrinsic stream terminated`.

I need a more expert pair of eyes to go through the internals of Substrate to understand where things are going wrong since I don't our side of the code silently ignoring an `InvalidTransaction` case.

FWIW, I'm pretty sure we were getting proper errors back when using `ChargeTransactionPayment`.

PR #351, opened 2020-04-13
Fix CLI reporting terminated streams on errors

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/348.
It does 2 things to improve reporting errors to the users:
1. Patches jsonrpc with the newest version from master, which handles errors correctly (temporary fix until Substrate officially releases it and upgrades Substrate)
2. Prints CLI errors recursively

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,10 @@ members = [
   "runtime",
   "test-utils"
   ]
+
+# TODO remove when Substrate upgrades jsonrpc-core-client to 14.1.0
+# https://github.com/paritytech/substrate/issues/5623
+[patch.crates-io]
+jsonrpc-pubsub = { git = "https://github.com/paritytech/jsonrpc", tag = "v14.1.0" }
+jsonrpc-core = { git = "https://github.com/paritytech/jsonrpc", tag = "v14.1.0" }
+jsonrpc-core-client = { git = "https://github.com/paritytech/jsonrpc", tag = "v14.1.0" }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -16,6 +16,7 @@
 //! The executable entry point for the Radicle Registry CLI.
 
 use radicle_registry_cli::CommandLine;
+use std::error::Error;
 use structopt::StructOpt;
 
 #[async_std::main]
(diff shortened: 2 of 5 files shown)


NunoAlexandre:
I am going to give it a try but I don't think this fixes the issue reported in #348 .

NunoAlexandre:
It actually solves the issue! However, I don't quite like the way it now prints the whole error stack, giving out more info than what would be friendly to users:

Screenshot from 2020-04-14 09-41-17

I'd prefer it to say like "Invalid Transaction: The specified fee is too low"

NunoAlexandre:
And the tests need to be fixed: https://buildkite.com/monadic/radicle-registry/builds/1948#2597de8b-6b8e-4ef0-b363-5b8bc9d4df0d

CodeSandwich:
The tests are already passing on the CI.

We're not getting the transaction invalidity cause, so all we could have is `Invalid Transaction`.

I think that limiting information in the error messages is a tricky thing to do. Are you sure that displaying only the root cause is always a good thing to do? In this case maybe, but there can be other errors, which without context don't make sense. I think that this should be done with deeper consideration and preferably in a separate PR. We've started telling the user what's wrong, next we can make it pretty.

NunoAlexandre:
> We're not getting the transaction invalidity cause, so all we could have is Invalid Transaction.

Yeah, I wonder where the actual cause is lost.

> I think that limiting information in the error messages is a tricky thing to do. Are you sure that displaying only the root cause is always a good thing to do? In this case maybe, but there can be other errors, which without context don't make sense. I think that this should be done with deeper consideration and preferably in a separate PR. We've started telling the user what's wrong, next we can make it pretty.

I agree with you in all things here. This is a great improvement, we can think of the other bits later.

cli/src/main.rs
```diff
@@ -27,7 +28,13 @@ async fn main() {
     match result {
         Ok(_) => std::process::exit(0),
         Err(error) => {
-            eprintln!("ERROR: {}", error);
+            eprintln!("Error: {}", error);
+            let mut parent = &error as &dyn Error;
+            while let Some(source) = parent.source() {
+                eprintln!("  Caused by: {}", source);
+                parent = source;
+            }
```
NunoAlexandre, line 36: Can we move this to a separate `print_error` function?

Cargo.toml
```diff
@@ -7,3 +7,9 @@ members = [
   "runtime",
   "test-utils"
   ]
+
+# TODO remove when Substrate upgrades jsonrpc-core-client beyond 14.0.5
```
geigerzaehler, line 11: Is there an issue for this that we could link?

CodeSandwich: I've asked the Substrate team directly to do that. That's a good idea, I'll create issues on jsonrpc and Substrate to link them here.

cli/src/main.rs
CodeSandwich: Done

Cargo.toml
CodeSandwich: Added

PR #353, opened 2020-04-14
node: Implement TryFrom for Chain

NunoAlexandre:
nit found while on #352

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -14,12 +14,12 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 //! Provides [Arguments] struct that represents the command line arguments.
+use crate::chain_spec::Chain;
+use core::convert::TryFrom;
 use radicle_registry_runtime::AccountId;
 use sc_cli::{RunCmd, Subcommand};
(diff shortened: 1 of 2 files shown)


node/src/cli.rs
```diff
@@ -147,15 +147,7 @@ impl Arguments {
...
-    if name == "dev" {
-        Ok(Chain::Dev)
-    } else if name == "local-devnet" {
-        Ok(Chain::DevnetLocal)
-    } else if name == "devnet" {
-        Ok(Chain::Devnet)
-    } else {
-        Err(format!("Invalid chain {}", name))
-    }
+    Chain::try_from(name.to_string())
```
geigerzaehler, line 150: I’d prefer to keep the code here. The reason is that the string values are also present in the `possible_values` annotation for the `--chain` argument. If we update, add or remove a chain we know need to touch two modules and the comment above does not make sense anymore.

In addition, we now have one more indirection which makes the code harder to follow.

CodeSandwich: Another point: this string is a strictly CLI representation, we should keep it with the CLI and not infect the inner types with that knowledge.

PR #357, opened 2020-04-14
Remove Polkadot telemetry

CodeSandwich:
Also improves the node CLI docs. This stems from comments on https://github.com/radicle-dev/radicle-run/pull/10.


NunoAlexandre:
I suggest splitting the two changes here: the drop of the polkadok telemetry and the touched docs into two separate commits.

PR #360, opened 2020-04-15
cli: Announce transactions with waiting notice

### LINKED ISSUE #356: cli: Print that transactions will take a while to return a status

NunoAlexandre:
Closes #356

### DIFF
--- a/cli/src/command/mod.rs
+++ b/cli/src/command/mod.rs
@@ -51,3 +51,8 @@ where
 fn parse_account_id(data: &str) -> Result<AccountId, String> {
     Ss58Codec::from_ss58check(data).map_err(|err| format!("{:?}", err))
 }
+
+fn announce_tx(msg: &str) {
+    println!("{}", msg);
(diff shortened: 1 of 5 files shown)


cli/src/command/mod.rs
```diff
@@ -51,3 +51,8 @@ where
 fn parse_account_id(data: &str) -> Result<AccountId, String> {
     Ss58Codec::from_ss58check(data).map_err(|err| format!("{:?}", err))
 }
+
+fn announce_tx(msg: &str) {
+    println!("{}", msg);
```
CodeSandwich, line 56: We can print `"{}..."`, the dots are a CLI-wide convention too.

NunoAlexandre: If that's ok with you, I prefer to leave that out shall we need to print a message where we don't want to suffix with `...`.

CodeSandwich: I can't think about any use case when we want to wait for something without `...`, but it's fine by me :slightly_smiling_face:

PR #362, opened 2020-04-15
Clarify Balance unit as μRAD

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/361

### DIFF
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -63,7 +63,7 @@ impl CommandT for ShowBalance {
     async fn run(self) -> Result<(), CommandError> {
         let client = self.network_options.client().await?;
         let balance = client.free_balance(&self.account_id).await?;
-        println!("{} RAD", balance);
+        println!("{} μRAD", balance);
         Ok(())
     }
 }
@@ -140,7 +140,7 @@ impl CommandT for Transfer {
         let transfered = transfer_fut.await?;
         transaction_applied_ok(&transfered)?;
         println!(
-            "✓ Transferred {} RAD to {} in block {}",
+            "✓ Transferred {} μRAD to {} in block {}",
             self.amount, self.recipient, transfered.block,
         );
         Ok(())
--- a/cli/src/command/org.rs
+++ b/cli/src/command/org.rs
@@ -203,7 +203,7 @@ impl CommandT for Transfer {
         let transfered = transfer_fut.await?;
         transaction_applied_ok(&transfered)?;
         println!(
-            "✓ Transferred {} RAD from Org {} to Account {} in block {}",
+            "✓ Transferred {} μRAD from Org {} to Account {} in block {}",
             self.amount, self.org_id, self.recipient, transfered.block,
         );
         Ok(())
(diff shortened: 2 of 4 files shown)


NunoAlexandre:
I think that most terminals will support it just fine, as utf-8 is usually set by default by most shells, at least the ones I've used. But it doesn't hurt at all to play safe and follow your suggestion! I think it's fine to use 'u' instead. @cloudhead do you object?

cloudhead:
Why not show the approximate amount in `RAD`, by using decimal points? uRADs aren't really useful for the user.

NunoAlexandre:
@cloudhead the entire registry operates in μRAD. To do that, we'd need to convert it every time we want to display it in μRAD. We can find clean ways of doing that, but it's added effort without much benefit at the moment, or we have never seen it until now. @CodeSandwich what's your take?

CodeSandwich:
I think that this is a good approach to have RAD expressed as a floating point number in the whole CLI. Obviously it would have to be parsed and stored as a regular integer. I like that idea and it shouldn't be hard to implement. I have no idea if this is important for the ffnet, it would be nice to have an ultimate-ish user experience, but on the other hand it's just an introduction of the decimal point.

NunoAlexandre:
Introducing this is not just a matter of always printing `Balance` as `RAD` (outgoing conversion) but also introducing the incoming conversion, where users should then specify RADs to be transferred, or RADs they pay as a fee. This conversion is sensitive and has consequences, unlike the outgoing one. Shouldn't then each Radicle component with the Registry follow the same approach, to avoid, say, the CLI in `RAD` and the upstream in `uRAD`?

It sounds much simpler and much less error-prone to have it all in uRAD. I don't know how other chain networks go about this though.

@geigerzaehler would highly appreciate your input!

geigerzaehler:
@cloudhead raises a good point, but I agree with @NunoAlexandre that implementing this takes some time and thought and I think that is something that is not necessary for the FFnet. I would suggest creating an issue for that so that we can properly estimate and prioritize this.

Regarding the μ/u discussion: I’d like to investigate what happens if we print unicode to non-unicode terminals. I’d think that none of the users the CLI targets actually uses a non-unicode terminal. So I’d be ok with not considering that edge case. If we do want to do that I’d prefer `microRAD` over `uRAD`. The visual approximation might not be obvious to everybody.

NunoAlexandre:
UTF8 seems to be the default encoding for most terminal for many years. I also tried printing it in a vanilla macOS terminal and it worked as well. Just didn't have a chance to try it in a PowerShell or other window terminal. Most likely our window users won't be using the PowerShell though...

cloudhead:
I would vote for using the correct unicode symbol and see if any user complains :)

In terms of the conversation on what to display: the "storage" or "internal accounting" unit (ie. uRADs) should have no consequence on how things are displayed at the UI level to users. Those are two different things.

If I build accounting software and store everything in cents, that doesn't mean I'm going to display all amounts to users in cents.

As far as I know, all projects handle things this way, eg. the smallest unit of Bitcoin is called a Satoshi, and that's what is actually represented in the ledger. However, users are used to operating in Bitcoin, so you'll see amounts like `0.184134 BTC`.

I agree though that for the FFNet this isn't important, and we can just show the uRAD amount, as at least that is correct.

NunoAlexandre:
> If I build accounting software and store everything in cents, that doesn't mean I'm going to display all amounts to users in cents.

yeah, that's absolutely clear to me :+1: 


Will rebase and merge given that we seem to be all into trying the current version using the unicode symbol.

PR #366, opened 2020-04-16
Document how to build from source

### LINKED ISSUE #318: Add instructions to build the project from source
We currently only provide prebuilt binaries for x86_64 linux targets. Users should also be able to run the node and use the tools on other targets. For this we want to provide instructions to build the binaries from source in the project readme. The instructions should not require any familiarity with the rust toolchain. We should also warn the users that it takes a long time to build.

We need to also consider that the built binaries should work with the ffnet by default. We need to ensure that they are properly configured.

NunoAlexandre:
Closes #318 

It wasn't straight forward to figure where to place this new info.
I decided to have it in the README since it's more user-friendly and satisfies
the developer and participant use cases fairly.

Ran the instructions in a vanilla macOS.

I find the README and DEVELOPING files rather convoluted, I will give it some thought on how to improve them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,15 @@ Radicle Registry
 
 Experimental Radicle Registry implementation with Substrate.
 
-See [`DEVELOPING.md`][dev-manual] for developer information.
+Click [here](https://radicle.run/docs/getting-started) to learn how to get
+started participating in the Radicle Registry Network.
+
+See [`DEVELOPING.md`](./DEVELOPING.md) for developer information.
 
 <!-- toc -->
 
+- [Prerequisites](#prerequisites)
+- [Build from source](#build-from-source)
 - [Getting the Node](#getting-the-node)
 - [Running the node](#running-the-node)
 - [Using the Client](#using-the-client)
@@ -18,6 +23,51 @@ See [`DEVELOPING.md`][dev-manual] for developer information.
 - [License](#license)
 
 <!-- tocstop -->
+
+Prerequisites
+---------------
+
+ⓘ Follow this guide to get started [developing](./DEVELOPING.md)
+   or to [build from source](#build-from-source).
+
+1. [Install Rust](https://www.rust-lang.org/tools/install)
+
+   It will run `rustup`, The Rust toolchain installer, which installs
+   `rustc`, `cargo`, `rustup`, amongst other standard tools that compose
+   the default Rust toolchain.
+
+2. Run `rustc --version` in a new shell.
+
+   To verify that the installion is sound.
+
+3. Run `./scripts/rustup-setup`
+
+   To install all `rustup` components and targets required by the
+   Radicle Registry.
+
+
+Build from source
+-----------------
+
+⚠ Make sure you have followed the [prerequisites](#prerequisites) guide
+  before proceeding.
+
+We currently only provide prebuilt binaries for `x86_64` linux targets in
+the [Radicle Registry Releases][releases-page] page.
+
+To build the Radicle Registry binaries to run on other targers, run:
+
+``` bash
+./scripts/build-release
+```
+
+The resulting binaries can be found at:
+
+* `./target/release/radicle-registry-cli`
+* `./target/release/radicle-registry-node`
+
+[releases-page]: https://github.com/radicle-dev/radicle-registry/releases
+
 
 Getting the Node
 ----------------
@@ -148,10 +198,6 @@ Using the CLI
 
 We provide a CLI to talk read and update the ledger in the `cli` directory. To
 learn more run `cargo run -p radicle-registry-cli -- --help`.
-
-
-[dev-manual]: ./DEVELOPING.md
-[rustup-install]: https://github.com/rust-lang/rustup.rs#installation
 
 
 License
--- a/ci/run
+++ b/ci/run
@@ -55,8 +55,8 @@ echo "--- cargo doc"
 RUSTDOCFLAGS="-D intra-doc-link-resolution-failure" \
   cargo doc --workspace --no-deps --document-private-items
 
-echo "--- scripts/build-release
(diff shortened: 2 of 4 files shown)


ci/run
```diff
@@ -55,8 +55,8 @@ echo "--- cargo doc"
 RUSTDOCFLAGS="-D intra-doc-link-resolution-failure" \
   cargo doc --workspace --no-deps --document-private-items
 
-echo "--- cargo build"
-DEFAULT_CHAIN=ffnet cargo build --workspace --all-targets --release
+echo "--- scripts/build-release"
+time ./scripts/build-release
```
CodeSandwich, line 59: I think that the CI already measures time

README.md
```diff
@@ -19,6 +21,54 @@ See [`DEVELOPING.md`][dev-manual] for developer information.
 
 <!-- tocstop -->
 
+Getting started
```
CodeSandwich, line 24: This is more like `Prerequisites` and it's necessary only for building. Maybe it should be the first sub-section of `Build from source`?

README.md
```diff
@@ -5,10 +5,12 @@ Radicle Registry
 
 Experimental Radicle Registry implementation with Substrate.
 
-See [`DEVELOPING.md`][dev-manual] for developer information.
+See [`DEVELOPING.md`](./DEVELOPING.md) for developer information.
 
 <!-- toc -->
 
+- [Getting started](#getting-started)
+- [Build from source](#build-from-source)
```
CodeSandwich, line 13: I think that build instructions should be below `Getting the Node`. Downloading a pre-built binary should be the default and the first thing to try out for a random passerby.

README.md
```diff
@@ -19,6 +21,54 @@ See [`DEVELOPING.md`][dev-manual] for developer information.
...
+``` bash
+./scripts/build-release
+```
+
+The resulting binaries can be found at:
+
+* `./target/release/radicle-registry-cli`
+* `./target/release/radicle-registry-node`
+
+Click [here](https://radicle.run/docs/getting-started) to learn how to get
```
CodeSandwich, line 66: Maybe it should be the very first thing in `README.md`? We want to push there as many people as possible not to scare them away with compilation and stuff.

NunoAlexandre: That's a good point.

README.md
NunoAlexandre: As explained, it's necessary for 1) building and for 2) developing, that's why I didn't nest it. We can rename it to Prerequisites but, in the end, it's the same, a list of steps to get started developing or building the binaries from source.

ci/run
NunoAlexandre: Doesn't hurt and other commands are `time`d as well

README.md
NunoAlexandre: Build instructions are general to all binaries (node and cli at the moment), not specific to 'getting a node'. Having a link to the releases page where the default binaries can be obtained and the section explaining how to build them from source, we should consider the 'Getting the node' section. What use case does it solve? None for the ffnet, none for me (personally) as a dev. How is it for you?

ci/run
CodeSandwich: Only `check-license-headers` and `cargo fmt`. All the real time sinks like `cargo clippy`s, `cargo doc`, `cargo build` and `cargo test` are not. We probably should make it consistent this way or another. I'd vote for complete removal, it only adds noise and brings no value in a context of the CI.

README.md
CodeSandwich: You're right it's necessary for everything below. Renaming alone will make it perfect.

NunoAlexandre: Done

ci/run
NunoAlexandre: Removed it from the new line added in this PR, feel free to follow up with making it consistent overall :+1:

NunoAlexandre:
@CodeSandwich thanks for your review. All discussions are resolved, can you have another look?

README.md
```diff
@@ -5,10 +5,15 @@ Radicle Registry
...
 
-See [`DEVELOPING.md`][dev-manual] for developer information.
+Click [here](https://radicle.run/docs/getting-started) to learn how to get
+started participating in the Radicle Registry Network.
+
+See [`DEVELOPING.md`](./DEVELOPING.md) for developer information.
 
 <!-- toc -->
 
+- [Getting started](#getting-started)
```
CodeSandwich, line 15: The ToC didn't update itself

scripts/build-release
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Build the CLI and node binaries for release. The resulting binaries can be found at:
+# * `./target/release/radicle-registry-cli`
+# * `./target/release/radicle-registry-node`
+
+DEFAULT_CHAIN=ffnet cargo build --workspace --all-targets --release
```
geigerzaehler, line 7: Is there a reason we build all targets in all packages? Since we only want to build the CLI and the node the following should be enough and quicker
```
cargo build --release -p radicle-registry-node -p radicle-registry-cli
```

NunoAlexandre: Note that that was the command previously executed already. As for your point, I think Cargo find finds which projects are binaries/executables, right? So I don't imagine a difference in performance but I will have a look.

geigerzaehler: With the current command we build all targets. This includes tests, examples, and all other binaries. This requires cargo to build more and I expect it to take more time.

NunoAlexandre: Was there a reason to have had those options in the past?

geigerzaehler: Ah, I just noticed that `./ci/run` also uses `./scripts/build-release`. For CI it makes sense to build everything since we want to type check everything and run the tests. Just for users who just want to build the node and CLI from source it’s overkill.

PR #368, opened 2020-04-17
cli: Refactor lookup_account

NunoAlexandre:
This was a thorn in my side, where I wanted to have the logic getting a specific account encapsulated within `account_storage` but my Rust had not figure how to overcome the compiler.

Thanks for the tip, @geigerzaehler :smile:

### DIFF
--- a/cli/src/account_storage.rs
+++ b/cli/src/account_storage.rs
@@ -28,7 +28,7 @@ use std::path::PathBuf;
 /// The data that is stored in the filesystem relative
 /// to an account. The account name is used as the key
 /// to this value, therefore not included here.
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone)]
 pub struct AccountData {
     pub seed: Seed,
 }
@@ -50,6 +50,10 @@ pub enum Error {
     /// Failed to read the accounts file
     #[error("Failed to read the accounts file: {0}")]
     FailedRead(#[from] ReadingError),
+
+    /// Could not find an account with the given name
+    #[error("Could not find an account with the given name")]
+    NotFound(),
 }
 
 /// Possible errors when writing to the accounts file.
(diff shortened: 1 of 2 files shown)


cli/src/account_storage.rs
```diff
@@ -97,6 +101,17 @@ pub fn list() -> Result<HashMap<String, AccountData>, Error> {
...
 
+/// Get an account by name.
+///
+/// It can fail from IO and Serde Json errors, or if no such
+/// account is found.
+pub fn get(name: &str) -> Result<AccountData, Error> {
+    list()?
+        .get(name)
+        .map(Clone::clone)
+        .ok_or(Error::NotFound())
```
geigerzaehler, line 112: It would be more idiomatic to return an `Option` instead of making the case an error.

NunoAlexandre: Yeah I agree, the thing is that getting an account might fail at getting all accounts, so the `None` case would cover both the io/deserialization error case as well as the case where there's no account with the given name. Imagining the accounts file getting corrupted, we would return "account not found" when the problem is not quite that.

geigerzaehler: Sorry for the confusion: I wasn’t suggesting using only `Option` but wrapping the option. We should indeed retain all the errors.
```rust
pub fn (get (name: &str) -> Result<Option<AccountData>, Error> {
  list()?.get(name).map(Clone::clone)
}
```

NunoAlexandre: Ah! Is it really more idiomatic? I understand the suggestion but it seems unnecessarily complex to have two lanes of error handling.

geigerzaehler: In this case it does not make too much of a difference. But in general the wrapped option is preferable. Often missing values are handled differently. For example the caller could provide a default value, whereas other errors would be propagated.

NunoAlexandre: Yeah, that's a good example. Although it would still be possible to that matching on `CommandError::NotFound`. If that's fine with you I will leave it as is.

Issue #289, opened 2020-03-26
Monitor difficulty


geigerzaehler:
We would like to monitor difficulty of the blockchain to analyze its behavior. The difficulty metric should be reported based on the best block in the chain.

PR #369, opened 2020-04-17
Add best block difficulty metric

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/289. I'm dissatisfied with exposing `Difficulty` types in pow algorithms and importing `scale_codec`. Both of them were avoidable (I still have a working branch with it), but the generic boundaries would get insane and borderline unmaintainable. It's a trade-off that had to be made.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -21,12 +21,12 @@ use futures::StreamExt;
 use std::convert::TryFrom;
 use std::sync::Arc;
 
-use sc_client::{BlockchainEvents as _, LongestChain};
+use sc_client::{light::blockchain::AuxStore, BlockchainEvents as _, LongestChain};
 use sc_executor::native_executor_instance;
-use sc_service::{AbstractService, Configuration, Error as ServiceError, ServiceBuilder};
+use sc_service::{AbstractService, Configuration, Error, ServiceBuilder};
 use sp_inherents::InherentDataProviders;
 
-use crate::pow::config::Config as PowAlgConfig;
+use crate::pow::{blake3_pow::Blake3Pow, config::Config, dummy_pow::DummyPow, Difficulty};
 use radicle_registry_runtime::{registry::AuthoringInherentData, AccountId, Block, RuntimeApi};
 
 // Our native executor instance.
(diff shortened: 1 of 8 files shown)


node/src/service.rs
```diff
@@ -18,15 +18,19 @@
...
 use sc_executor::native_executor_instance;
-use sc_service::{AbstractService, Configuration, Error as ServiceError, ServiceBuilder};
+use sc_service::{AbstractService, Configuration, Error, ServiceBuilder};
 use sp_inherents::InherentDataProviders;
 
-use crate::pow::config::Config as PowAlgConfig;
+use crate::pow::{
+    blake3_pow::{Blake3Pow, Difficulty as Blake3PowDifficulty},
+    config::Config as PowAlgConfig,
+    dummy_pow::{Difficulty as DummyPowDifficulty, DummyPow},
```
geigerzaehler, line 32: How about declaring only _one_ difficulty in `node/src/pow/mod.rs`? This would also allow us to make `register_difficulty_metric_impl` monomorphic in the difficulty. This would also address the concerns you raised in the description.

node/src/service.rs
```diff
@@ -214,13 +220,66 @@ pub fn new_light(config: Configuration) -> Result<impl AbstractService, ServiceE
...
+            ready(
+                sc_consensus_pow::PowAux::<Difficulty>::read::<_, S::Block>(&*client, &info.hash)
+                    .ok(),
+            )
+        })
+        .map(|difficulty| difficulty.difficulty.try_into().unwrap_or(u64::MAX))
+        .for_each(move |difficulty| {
+            gauge.set(difficulty);
+            ready(())
+        });
```
geigerzaehler, line 274: We’re using a lot of stream methods. I think it would be more readable, less complicated, and more performant to just call `for_each` on and have everything in the handler.
```suggestion
        .map(|difficulty| difficulty.difficulty.try_into().unwrap_or(u64::MAX))
        .for_each(move |info| {
            if info.is_new_best {
                let pow_aux_result = sc_consensus_pow::PowAux::<Difficulty>::read::<_, S::Block>(&*client, &info.hash);
                match pow_aux_result {
                    Ok(pow_aux) =>  {
                        let difficulty = pow_aux.difficulty.try_into().unwrap_or(u64::MAX);
                        gauge.set(difficulty);
                    },
                    Error(err) => todo!("log error"),
                }
            }
            ready(())
        });
```
Maybe it also makes sense to extract the `PowAux::read` and `.difficulty.try_into()` code into a separate function.

node/src/service.rs
CodeSandwich: Yeah, that's a wise solution, removes all the unnecessary complexity.

CodeSandwich: Done

node/src/service.rs
CodeSandwich: I've turned it into a single closure. It is a bit more readable and easier to understand. I wouldn't mind performance, the compiler is really good in inlining closures and removing unnecessary steps.

I haven't extracted `PowAux::read` and `.difficulty.try_into()` to a separate function, because its signature contained some very complex generics and it would only replace 2 rather clear oneliners.

PR #371, opened 2020-04-17
Add boot nodes to FFnet chain spec

geigerzaehler:
See https://github.com/radicle-dev/infra/pull/82

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -90,7 +90,13 @@ fn ffnet() -> ChainSpec {
         "ffnet",
         "ffnet",
         dev_genesis_config,
-        vec![],        // boot nodes
+        // Addresses are defined here: https://github.com/radicle-dev/infra/tree/master/registry/ffnet
+        vec![
+            "/dns4/boot-0.ff.radicle.network./tcp/30333/p2p/QmdEvLkAS8mxETQy1RCbdmcPPzxSs9RbExFcWvwJZDXxjG"
+                .parse().unwrap(),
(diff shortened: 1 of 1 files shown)


node/src/chain_spec.rs
```diff
@@ -90,7 +90,13 @@ fn ffnet() -> ChainSpec {
...
         "ffnet",
         dev_genesis_config,
-        vec![],        // boot nodes
+        // Addresses are defined here: https://github.com/radicle-dev/infra/tree/master/registry/ffnet
+        vec![
+            "/dns4/boot-0.ff.radicle.network./tcp/30333/p2p/QmdEvLkAS8mxETQy1RCbdmcPPzxSs9RbExFcWvwJZDXxjG"
+                .parse().unwrap(),
+            "/dns4/boot-1.ff.radicle.network./tcp/30333/p2p/QmceS5WYfDyKNtnzrxCw4TEL9nokvJkRi941oUzBvErsuD"
+                .parse().unwrap()
+        ],
```
NunoAlexandre, line 99:
```suggestion
        // Addresses are defined here: https://github.com/radicle-dev/infra/tree/master/registry/ffnet
        vec![
            "/dns4/boot-0.ff.radicle.network./tcp/30333/p2p/QmdEvLkAS8mxETQy1RCbdmcPPzxSs9RbExFcWvwJZDXxjG",
            "/dns4/boot-1.ff.radicle.network./tcp/30333/p2p/QmceS5WYfDyKNtnzrxCw4TEL9nokvJkRi941oUzBvErsuD",        
        ].iter().map(|x| x.parse().unwrap()).collect()
```

geigerzaehler: I was considering the same thing but I landed with the proposed. I believe it is just more readable since we don’t have the additional method calls `.iter()`, `.map()`, and `.collect()`.

Issue #294, opened 2020-03-26
Expose ledger error information


geigerzaehler:
_See also #293_

Ledger code might error, for example if a given project ID already exists. We want assign codes to these errors and expose this error information through the client API.

As a result of this it should be possible for the CLI to show an appropriate error message when a transfer transaction was sent but the user does not have enough funds

### Implementation

Currently all ledger errors are defined in `core::RegistryError`. These errors are returned from the the registry message handlers in the runtime when an error occurs and have an associated error code.

What is missing is exposing these errors properly through the client API. We have a mechanism to retrieve the result of a transaction via `TransactionApplied::result` which returns the associated `Message::Result` type. However this associated type is a result with `DispatchError` for all existing messages.

It remains to change the error type to a specific error enum for each message and ensure that this error is properly extracted from `DispatchError`

geigerzaehler:
I’m removing the “project-health” label since this delivers value to the client consumers

NunoAlexandre:
> It remains to change the error type to a specific error enum for each message and ensure that this error is properly extracted from DispatchError

We already have `RegistryError`, which, at the moment, is encoded in a DecodeError. In #372, I've introduced a way to attempt the conversion DispatchError -> RegistryError, which allows us to unwrap it if possible, better handle it, and provide better error information.

The quote above sounds like we'd want each message to have its own error type? We can consider that but focusing on first having the client providing better errors than the obscure/encoded DispatchError's should be the first step.

geigerzaehler:
I’m reopening this because we don’t yet expose ledger error information properly in the client API. We also need to decide whether we want to make the error type dependent on the message.

NunoAlexandre:
This was closed by mistake, probably wrongly automated through closing a PR referencing it.

> We also need to decide whether we want to make the error type dependent on the message.

Yes. I don't think that's beneficial overall. The upstream already suffers from ergonomic issues from having Message as a trait instead of an enum, if the error type also becomes a dependant type I only see those issues growing.

I'm started with this btw. Will address it in small chunks and invite discussion where fit.

geigerzaehler:
> I'm started with this btw. Will address it in small chunks and invite discussion where fit.

Could you also describe the design beforehand? With incremental changes one sometimes forgets the bigger picture.

NunoAlexandre:
> Could you also describe the design beforehand? With incremental changes one sometimes forgets the bigger picture.

[...]

PR #374, opened 2020-04-20
Thiserror/registry error

NunoAlexandre:
Step 1. of  #294

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -273,7 +273,7 @@ decl_module! {
                         Ok(())
                     }
                     else {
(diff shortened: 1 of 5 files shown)


runtime/src/registry.rs
```diff
@@ -273,7 +273,7 @@ decl_module! {
                         Ok(())
                     }
                     else {
-                        Err(RegistryError::NonUnregisterableUser.into())
+                        Err(RegistryError::UnregisterableUser.into())
```
NunoAlexandre, line 276: This variant was renamed to be sound and consistent with the older one, `UnregisterableOrg`.

PR #375, opened 2020-04-20
Client: expose registry errors

NunoAlexandre:
Step 2. of #294 

>  2. Move the CLI TransactionError to the core and have that as the error type of every Message
>     2.1 Adjust the affected tests


core/src/error.rs
```diff
@@ -19,6 +19,31 @@ use derive_try_from_primitive::TryFromPrimitive;
 use std::convert::{TryFrom, TryInto};
 use thiserror::Error as ThisError;
 
+/// The subset of possible errors having led a transaction to failure.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, ThisError)]
+pub enum TransactionError {
+    #[error(transparent)]
+    RegistryError(RegistryError),
```
CodeSandwich, line 26:
```suggestion
    RegistryError(#[from] RegistryError),
```

cli/src/command/org.rs
```diff
@@ -121,8 +121,7 @@ impl CommandT for Register {
             .await?;
         announce_tx("Registering org...");
 
-        let org_registered = register_org_fut.await?;
-        transaction_applied_ok(&org_registered)?;
+        register_org_fut.await?;
```
CodeSandwich, line 124: Are we checking the transaction `result`?

core/src/error.rs
```diff
@@ -19,6 +19,31 @@ use derive_try_from_primitive::TryFromPrimitive;
...
+            .unwrap_or(TransactionError::OtherDispatchError(dispatch_error))
+    }
+}
+
+impl From<RegistryError> for TransactionError {
+    fn from(registry_error: RegistryError) -> Self {
+        TransactionError::RegistryError(registry_error)
+    }
+}
+
```
CodeSandwich, line 46:
```suggestion
```

core/src/error.rs
NunoAlexandre: Thanks! RegistryError wasn't a `thiserror` before and forgot to do this here when it became
one :v:

cli/src/command/org.rs
NunoAlexandre: great catch! this is just checking the higher level result at the client-level, not at the transaction-level. Will fix. Thanks!

NunoAlexandre: :heavy_check_mark:

PR #377, opened 2020-04-21
cli: Mutually support address and account name

### LINKED ISSUE #376: cli: Receive account name in `account balance`
Now we need to pass the ss58 address to this command, requiring us to first look it up, when we could just specify the local account name. This was pointed out during the internal testing last week.

:exclamation: Now, if we do this, we lose the capacity to check the balance of **any** account.

@CodeSandwich @geigerzaehler do you see any use case where it would be important or even necessary to do that? I don't, I think we should be good with this change.

NunoAlexandre:
Closes #376 (+ it does that for all params parsing ss58 addresses)

All CLI params currently receiving an SS58 address now also support a
local account, in a polymorphic fashion by falling back.

### DIFF
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -47,7 +47,7 @@ impl CommandT for Command {
 /// Show the balance of an account
 #[derive(StructOpt, Clone)]
 pub struct ShowBalance {
-    /// SS58 address
+    /// SS58 address or name of a local account.
     #[structopt(
         value_name = "account",
         parse(try_from_str = parse_account_id),
@@ -109,7 +109,8 @@ pub struct Transfer {
     // The amount to transfer.
     amount: Balance,
 
-    /// Recipient Account in SS58 address format.
+    /// The recipient account.
+    /// SS58 address or name of a local account.
     #[structopt(parse(try_from_str = parse_account_id))]
     recipient: AccountId,
 
--- a/cli/src/command/mod.rs
+++ b/cli/src/command/mod.rs
@@ -15,7 +15,7 @@
 
 //! Define the commands supported by the CLI.
 
-use crate::{CommandError, CommandT, NetworkOptions, TxOptions};
+use crate::{lookup_account, CommandError, CommandT, NetworkOptions, TxOptions};
 use radicle_registry_client::*;
 
 use sp_core::crypto::Ss58Codec;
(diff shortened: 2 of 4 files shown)


cli/src/command/account.rs
```diff
@@ -109,7 +110,9 @@ pub struct Transfer {
     // The amount to transfer.
     amount: Balance,
 
-    /// Recipient Account in SS58 address format.
+    /// The recipient account.
+    /// A SS58 address can be given directly or
+    /// obtained given a local account name.
```
geigerzaehler, line 115: I’m not sure it makes sense to also change this for transfers.The benefits don’t seem to outweigh the drawbacks. For example we don’t get any information if checksum verification of the SS58 addresses failed.

cli/src/command/mod.rs
```diff
@@ -43,8 +43,12 @@ where
         .map_err(|dispatch_error| dispatch_error.into())
 }
 
+/// Retrieve the AccountId associated with the given input,
+/// that can be either a SS58 address or a local account name.
 fn parse_account_id(data: &str) -> Result<AccountId, String> {
-    Ss58Codec::from_ss58check(data).map_err(|err| format!("{:?}", err))
+    Ss58Codec::from_ss58check(data)
+        .map_err(|err| format!("{:?}", err))
+        .or_else(|_| lookup_account(data).map(|account| account.public()))
```
geigerzaehler, line 51: If `parse_account_id` fails we only show the error that the account was not found. We should tell the user that there are _two_ possible problems. The address is invalid or the account was not found.

cli/src/command/account.rs
```diff
@@ -47,7 +47,8 @@ impl CommandT for Command {
 /// Show the balance of an account
 #[derive(StructOpt, Clone)]
 pub struct ShowBalance {
-    /// SS58 address
+    /// SS58 address that can be given directly or
+    /// obtained given a local account name.
```
geigerzaehler, line 51: This is displayed to the user so we should tell them what to input.
```suggestion
    /// SS58 address or name of a local account.
```
Please also adjust the `value_name`. It should be `address_or_account`.

NunoAlexandre: > we should tell them what to input.

That's what it does now: it tells users that the param represents an SS58 address that can be given directly or obtained given a local account name.

I am fine with your edition though.

cli/src/command/mod.rs
NunoAlexandre: My bad, good catch. Fixed by not only telling the users about the two possible errors but also indicate what exactly went wrong for each case:

``` 
error: Invalid value for '<account>':
    ! Could not parse an ss58 address nor find a local account with the given name.
    ⓘ Error parsing SS58 address: BadLength
    ⓘ Error looking up account: Could not find an account with the given name
```

cli/src/command/account.rs
NunoAlexandre: I'm not sure if I get the issue. Was your point related to the way we handle a failed parsing of the ss58 address? If so, that's now covered. Otherwise, could you expand?

Issue #379, opened 2020-04-21
Released binaries are not executable


CodeSandwich:
The user is forced to `chmod +x` the node and CLI binaries after downloading them from the releases page. It's a bad user experience when one follows http://radicle.run instructions.

geigerzaehler:
This is not something we can resolve on our side. Since the users are simply downloading the binaries from the web the downloader controls the permissions of the files that land on the user’s machine. #380 won’t solve that.

CodeSandwich:
I think that we can. Even if we can't find anything elegant, we can zip the binaries to preserve their execution flag. Unzipping would be still more user friendly than `chmod`ding.

NunoAlexandre:
It's more effort to unzip both binaries than to make them executable :man_shrugging: 

The only trade-off I see that might justify it is reducing download stress for the node binary.

NunoAlexandre:
> Unzipping would be still more user friendly

why and how, given who our target users are?

CodeSandwich:
> why and how, given who our target users are?

To be honest, I wouldn't remember about the hidden execution flag, which is cleared while downloading over HTTP. I would get that papercut right away and I would have to troubleshoot. On the other hand with the `.tar.gz` it's obvious what to do.

The archive is also much more friendly for the GUIs, I can unpack it by right-clicking in Dolphin or open it in Arc right from Firefox, `chmod`ding is more complex. In the terminal both ways are single commands.

NunoAlexandre:
I don't really agree, the only benefit I see is the reduce download stress.

Fwiw, in my run with Massi he immediately made the binaries executable right after downloading them, without me saying anything at all.

Also, I'm quite sure the standard practice when providing compressed binaries is to provide both.tar.gz and.zip.

I find it much simpler to just provide the binaries and have the chmod +x commands in the docs, which means adding 4 lines of text to the docs as opposite to all the CI changes this involves. My 2 cents

PR #380, opened 2020-04-21
Make artifacts executable

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/379 by tarballing the artifacts

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -106,8 +106,8 @@ Make a release
     - Expand the `Test ci/run` section
     - Open the `Artifacts` tab
     - Download the following binaries:
-        - `radicle-registry-cli`
-        - `radicle-registry-node`
+        - `radicle-registry-cli.tar.gz`
+        - `radicle-registry-node.tar.gz`
 2. Create a new release
     - Open the [Radicle Registry Github Releases page](https://github.com/radicle-dev/radicle-registry/releases)
     - Define the tag version (see [Tags](#tags))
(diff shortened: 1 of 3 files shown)


DEVELOPING.md
```diff
@@ -106,8 +106,8 @@ Make a release
     - Expand the `Test ci/run` section
     - Open the `Artifacts` tab
     - Download the following binaries:
-        - `radicle-registry-cli`
-        - `radicle-registry-node`
+        - `radicle-registry-cli.tar.gz`
+        - `radicle-registry-node.tar.gz`
```
geigerzaehler, line 110: Does it make sense to have two tarballs or should we combine them into one?

NunoAlexandre: It makes sense to have two, and every time I saw binaries provided this way a `zip` counterpart is also made available.

CodeSandwich: I was thinking about it and to be honest I don't know. It probably doesn't make sense to download only one. I'll merge them.

NunoAlexandre: No. we simply shouldn't force users to download two things if they only need one.

CodeSandwich: Sorry, I didn't see your comment, GitHub is sometimes a bit asynchronouse. Either way is good for me.

NunoAlexandre: We really shouldn't merge it, especially given that the node binary is like 10x bigger than the cli binary. That is a real UX issue. Running a tar command vs chmod +x command seems way more subjective to me.

geigerzaehler: Either way is fine with me.

CodeSandwich: For the sake of conversation: it's 8 and 38.5 MB or 2.5 and 14 MB if compressed.

PR #381, opened 2020-04-21
cli: set default node host to ffnet 

### LINKED ISSUE #330: CLI connects to FFnet hosts by default
We want to make it easier for FFnet users to use the CLI without running a local node. Currently the user needs to set the `--node-host` option to a publicly accessible node. We’ll provide such nodes as part of the FFnet infrastructure. To have the CLI work out of the box we want to replace the `localhost` default value of `--node-host` with a node from our ffnet infrastructure.

To take advantage of load balancing and to give us the most flexibility we’ll use the `rpc.ffnet.radicle.network` domain as described in #290.

If we change this we need to ensure that the “Getting Started” guide is updated accordingly

NunoAlexandre:
Closes #330 

Now that the default will no longer be `127.0.0.1`, I also created a safe parser that translates `localhost` to `127.0.0.1`, avoiding the issue with the translation to its ipv6 version.

### DIFF
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -47,7 +47,7 @@ pub struct NetworkOptions {
     /// IP address or domain name that hosts the RPC API
     #[structopt(
         long,
-        default_value = "127.0.0.1",
+        default_value = "rpc.ff.radicle.network",
         env = "RAD_NODE_HOST",
         parse(try_from_str = url::Host::parse),
     )]
@@ -55,6 +55,7 @@ pub struct NetworkOptions {
 }
 
 impl NetworkOptions {
+
     pub async fn client(&self) -> Result<Client, Error> {
         Client::create_with_executor(self.node_host.clone()).await
     }
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -49,12 +49,22 @@ pub struct NetworkOptions {
         long,
         default_value = "rpc.ff.radicle.network",
         env = "RAD_NODE_HOST",
-        parse(try_from_str = url::Host::parse),
+        parse(try_from_str = Self::parse_node_host),
     )]
     pub node_host: url::Host,
 }
 
 impl NetworkOptions {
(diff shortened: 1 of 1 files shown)


cli/src/lib.rs
```diff
@@ -47,14 +47,23 @@ pub struct NetworkOptions {
...
+        parse(try_from_str = Self::parse_node_host),
     )]
     pub node_host: url::Host,
 }
 
 impl NetworkOptions {
+
+    pub fn parse_node_host(value: &str) -> Result<url::Host, url::ParseError> {
+        let node_host = match value {
+            "localhost" => "127.0.0.1",
```
geigerzaehler, line 61: To the uninitiated it is unclear why we do this. Could you add a comment and link to the relevant issue?

cli/src/lib.rs
```diff
@@ -47,14 +47,23 @@ pub struct NetworkOptions {
...
         env = "RAD_NODE_HOST",
-        parse(try_from_str = url::Host::parse),
+        parse(try_from_str = Self::parse_node_host),
     )]
     pub node_host: url::Host,
 }
 
 impl NetworkOptions {
+
+    pub fn parse_node_host(value: &str) -> Result<url::Host, url::ParseError> {
```
geigerzaehler, line 59: No reason for this to by publich.

geigerzaehler:
Could you also open a companion PR on `radicle-run` that updates the getting started guide?

NunoAlexandre:
> Could you also open a companion PR on radicle-run that updates the getting started guide?

Yes, working on it

cli/src/lib.rs
NunoAlexandre: Indeed, comment added.

cli/src/lib.rs
NunoAlexandre: Dropped

NunoAlexandre:
The radicle-run companion PR: https://github.com/radicle-dev/radicle-run/pull/21

PR #383, opened 2020-04-21
Update rust toolchain to 2020-04-20

geigerzaehler:
See https://github.com/rust-lang/rust/issues/69933


client/src/lib.rs
```diff
@@ -234,7 +234,7 @@ impl ClientT for Client {
 
     async fn free_balance(&self, account_id: &AccountId) -> Result<state::AccountBalance, Error> {
         let account_info = self
-            .fetch_map_value::<system::Account<Runtime>, _, _>(account_id.clone())
+            .fetch_map_value::<system::Account<Runtime>, _, _>(*account_id)
```
NunoAlexandre, line 237: Did you happen to see this or was this reported by the updated toolchain?

geigerzaehler: The latter. Clippy caught it now

PR #385, opened 2020-04-21
Document release process with small scripts

geigerzaehler:
We update the documentation for the release process with scripts instead of explaining the manual steps in prose. This should improve repeatability and make it easier to use.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -101,20 +101,7 @@ using Kubernetes for this. You can find the code in `ci/deploy`.
 Make a release
 --------------
 
-1. Download the latest binaries for the targeted master build [in Builkite](https://buildkite.com/monadic/radicle-registry/builds?branch=master)
-    - Select the targeted build
-    - Expand the `Test ci/run` section
-    - Open the `Artifacts` tab
-    - Download the following binaries:
-        - `radicle-registry-cli.tar.gz`
-        - `radicle-registry-node.tar.gz`
-2. Create a new release
-    - Open the [Radicle Registry Github Releases page](https://github.com/radicle-dev/radicle-registry/releases)
-    - Define the tag version (see [Tags](#tags))
-    - Select the right target
-    - Add a release title
-    - Describe the changes included in this release to users
-    - Attach the binaries downloaded in step 1.
+To create a release from the current master branch run `./scripts/create-release`.
 
 ### Tags
 
--- a/scripts/create-release
+++ b/scripts/create-release
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+#
+# Create a Github draft release from the current master branch
+
+set -euo pipefail
+
+if ! ( command -v hub 1>/dev/null ); then
+  echo "hub was not found. Please install it from https://github.com/github/hub"
+  exit 1
+fi
+
+echo "Running git --fetch tags"
+git fetch --tags
+commit=$(git rev-parse origin/master)
+
+echo "Download build artifacts"
+base_url="https://builds.radicle.xyz/radicle-registry/master/$commit/artifacts"
+artifacts_dir=$(mktemp -d)
+(
+  cd "$artifacts_dir"
+  curl -sSLO "$base_url/radicle-registry-cli.tar.gz"
+  curl -sSLO "$base_url/radicle-registry-node.tar.gz"
+)
+
+declare -i release_counter=0
+while true; do
+  release_name="$(date +%Y-%m-%d)-$release_counter"
+  if [[ $(git tag -l "$release_name" | wc -l) -eq 0 ]]; then
+    break
+  fi
+  echo "Tag $release_name already exists"
+  release_counter+=1
+done
+
+echo "Creating and pushing tag"
+git tag --sign --message "$release_name" "$release_name" "$commit"
+git push origin "refs/tags/$release_name"
+
+echo "Creating Github draft release \"$release_name\""
+hub release create \
+  --draft \
+  --prerelease \
+  --attach "$artifacts_dir/radicle-registry-cli.tar.gz" \
+  --attach "$artifacts_dir/radicle-registry-node.tar.gz" \
+  "$release_name"
+rm -rf "$artifacts_dir"
+echo "Created draft release. Publish it on Github."


DEVELOPING.md
```diff
@@ -101,20 +101,29 @@ using Kubernetes for this. You can find the code in `ci/deploy`.
...
-    - Add a release title
-    - Describe the changes included in this release to users
-    - Attach the binaries downloaded in step 1.
+1. Create and push a new release tag. See [Tags](#tags) for details on tag
+   names.
+   ```bash
+   git fetch --tags
+   release="$(date +%Y-%m-%d)-0"
+   git tag "$release" origin/master --sign --message "$release"
+   git push --tags
```
CodeSandwich, line 116: I think that it's done with `hub release create`, from its docs:
```
create

    Create a GitHub release for the specified TAG name. If git tag TAG does not exist, it will be created at TARGET (default: current branch).
```

geigerzaehler:
> I don't find this any easier to follow...

Are there any concerns beyond switching points 1 and 2?

DEVELOPING.md
CodeSandwich: Unless the `--sign` flag is the only way to sign the tag and make it untemperable :thinking:

geigerzaehler: Good point. I’ll switch it to just use `hub`.

NunoAlexandre:
> Are there any concerns beyond switching points 1 and 2?

No. I don't understand why we are spending time on this now but no further concerns besides the structure as pointed above.

geigerzaehler:
> I think that it's a good idea to automate this process! I think that at this point it doesn't have to be a runbook, it can be easily turned into a script with comments.

Are you suggesting to move all the code to `./scripts/publish-release` and just reference and explain the script in the developer’s guide. I’d be up for that. But one thing to consider is that the script may require manual intervention if there was already a release on a given day. This should be rare though and inline docs should suffice to make it possible for devs to resolve this.

geigerzaehler:
> No. I don't understand why we are spending time on this now

Because it reduces the time (at least for me) spent on creating a releases and reduces the chance of mistakes.

CodeSandwich:
@geigerzaehler Yes, just create another script. I think that we can print a suggestion about release date if `git tag` returns a non-zero status. We should also check if `hub` is installed before the script runs.

DEVELOPING.md
CodeSandwich: I couldn't find anything about signing such tag, I think that it's an inferior solution.

NunoAlexandre:
Then we should just automate it fully. Now we first spent time creating the original version to not spend time automating and now we are changing it to an in-between solution. I sure see the improvement, I'm just not fond of the process.

NunoAlexandre:
> Yes, just create another script.

@geigerzaehler if we do this could you receive the release suffix counter as an arg, defaulting to 0 to avoid having to touch the script or copy-paste to release a second version in the same day?

geigerzaehler:
I’ve updated the PR. Please take another look @CodeSandwich and @NunoAlexandre.

scripts/create-release
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+echo "Running git --fetch tags"
+git fetch --tags
+commit=$(git rev-parse origin/master)
+
+echo "Download build artifacts"
+base_url="https://builds.radicle.xyz/radicle-registry/master/$commit/artifacts"
+curl -sSLO "$base_url/radicle-registry-cli.tar.gz"
+curl -sSLO "$base_url/radicle-registry-node.tar.gz"
```
CodeSandwich, line 19: We leave some junk in the current directory. How about this:
```suggestion
artifacts=$(mktemp -d)
(
  cd "$artifacts"
  curl -sSLO "$base_url/radicle-registry-cli.tar.gz"
  curl -sSLO "$base_url/radicle-registry-node.tar.gz"
)
```
and below:
```bash
hub release create \
  --draft \
  --prerelease \
  --attach "$artifacts/radicle-registry-cli.tar.gz" \
  --attach "$artifacts/radicle-registry-node.tar.gz" \
  -t "$commit"
  "$release_name"
```

scripts/create-release
```diff
@@ -0,0 +1,37 @@
...
+done
+
+echo "Creating github release $release_name"
+hub release create \
+  --draft \
+  --prerelease \
+  --attach radicle-registry-cli.tar.gz \
+  --attach radicle-registry-node.tar.gz \
+  -t "$commit"
+  "$release_name"
```
CodeSandwich, line 37: I think that we're not signing the tag

scripts/create-release
```diff
@@ -0,0 +1,37 @@
...
+while [[ $(git tag -l "$release_name" | wc -l) -ne 0 ]]; do
+  release_counter+=1
+  echo "Tag $release_name already exists"
+  release_name="$(date +%Y-%m-%d)-$release_counter"
+done
+
+echo "Creating github release $release_name"
+hub release create \
+  --draft \
+  --prerelease \
```
CodeSandwich, line 33: `--prerelease` will probably stay with us until the mainnet, but isn't `--draft` too cautious?

scripts/create-release
```diff
@@ -0,0 +1,37 @@
...
+curl -sSLO "$base_url/radicle-registry-cli.tar.gz"
+curl -sSLO "$base_url/radicle-registry-node.tar.gz"
+
+declare -i release_counter=0
+release_name="$(date +%Y-%m-%d)-$release_counter"
+
+while [[ $(git tag -l "$release_name" | wc -l) -ne 0 ]]; do
+  release_counter+=1
+  echo "Tag $release_name already exists"
+  release_name="$(date +%Y-%m-%d)-$release_counter"
```
CodeSandwich, line 27: This complex pattern is duplicated, maybe we could use `if` and `break` to build it DRY-ly?

scripts/create-release
geigerzaehler: You’re right It does not seem necessary to sign the tags as the commits are signed already. I can revert this to signing again though.

scripts/create-release
geigerzaehler: It gives us the opportunity to check this on Github and publish it there. Will document this though.

scripts/create-release
CodeSandwich: The commits are signed, but the choice of which one the users are going to use is not, I think that it's important after all.

scripts/create-release
CodeSandwich: Is it documented somewhere already or is it a WIP?

scripts/create-release
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Creating Github draft release \"$release_name\""
+hub release create \
+  --draft \
+  --prerelease \
+  --attach "$artifacts_dir/radicle-registry-cli.tar.gz" \
+  --attach "$artifacts_dir/radicle-registry-node.tar.gz" \
+  "$release_name"
+rm -rf "$artifacts_dir"
+echo "Created draft release. Publish it on Github."
```
NunoAlexandre, line 47: > Publish it on Github

We should make it more clear that this is a required follow up action. Maybe use some unicode symbol to call the deserved attention?

scripts/create-release
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
NunoAlexandre, line 1: This whole script is a beautiful piece.

scripts/create-release
CodeSandwich: It's documented below, resolving :+1:

scripts/create-release
CodeSandwich: A URL to releases page could help too, many terminals have good support for links.

NunoAlexandre: Yes, good idea!

geigerzaehler: `hub create release` prints the URL, albeit above this line.

PR #390, opened 2020-04-23
Improve cli show commands outputs

CodeSandwich:
The outputs are now more user friendly. The most important part is that the the org member ids were displayed as hexes and the first 8 characters of the SS58, now it's just the full SS58.

### DIFF
--- a/cli/src/command/org.rs
+++ b/cli/src/command/org.rs
@@ -86,9 +86,9 @@ impl CommandT for Show {
             })?;
 
         println!("id: {}", org.id);
-        println!("account_id: {}", org.account_id);
-        println!("members: {:?}", org.members);
-        println!("projects: {:?}", org.projects);
+        println!("account id: {}", org.account_id);
+        println!("member ids: [{}]", org.members.iter().format(", "));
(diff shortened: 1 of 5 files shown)


cli/src/command/org.rs
```diff
@@ -16,6 +16,7 @@
 //! Define the commands supported by the CLI related to Orgs.
 
 use super::*;
+use itertools::Itertools;
```
NunoAlexandre, line 19: `nit`

Something I sort of introduced when restructured the CLI was to have the imports in the `commands/mod.rs` file and then have each submodule `use  super::*;`.

If that sounds sound, could we move this import to the `mod.rs` and thus only import it once?

CodeSandwich: Personally I personally prefer having all the imports not further than the file where they are used, but if that's the convention here, I'm going to follow it.

CodeSandwich: Fixed

NunoAlexandre: Yeah, I also don't like it in general, just started it and liked not having to repeat the same imports for all sub modules. We can reconsider, of course. and thanks!

PR #393, opened 2020-04-23
cli: Show org balance on 'org show'

NunoAlexandre:
> An open question: would it make sense to show the amount held by the org in `radicle-registry-cli org show <org_name>`? Now this information is hidden deep in a separate command `radicle-registry-cli account balance <org_id>` by @CodeSandwich

### DIFF
--- a/cli/src/command/org.rs
+++ b/cli/src/command/org.rs
@@ -84,9 +84,11 @@ impl CommandT for Show {
             .ok_or(CommandError::OrgNotFound {
                 org_id: self.org_id.clone(),
             })?;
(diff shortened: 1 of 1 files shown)


CodeSandwich:
I wonder if adding an org balance in `Org` in `client` API would be beneficial to the users of this API :thinking:

NunoAlexandre:
> I wonder if adding an org balance to Org in client API would be beneficial to the users of this API thinking

I wondered the same. I would keep it separate since all the org fields are much less variable than its funds.

Issue #396, opened 2020-04-24
Return user ids instead of account ids in Org::members


MeBrei:
To display the list of org members in radicle-upstream, the list of members of [get_org](https://github.com/radicle-dev/radicle-registry/blob/1ecf004bb1e27f152e73716b1b316e0db878b0fd/client/src/interface.rs) should return a list of user ids instead of the public key. This way, `get_user` can be used to display the members in the ui. 

This blocks part of the radicle-upstream issue [Overview: Org I](https://github.com/radicle-dev/radicle-upstream/issues/277)

NunoAlexandre:
@MeBrei @xla 

Right now, every `User` has, always, one account associated. However, not all accounts are associated with a user.

The `Org::members` field, as of now, only contains the `AccountId` (pub key) of the author who has registered said org. The author is not required to have be a registered user.

### Possible solutions

#### A
- A requirement is added when registering an org where the author must be a registered user (i.e, a user must exist whose account id equals the pub key of the author). 
- That user id of said author would be stored as the only org member, instead of its account id.

#### B

- We add a `fn get_user(account_id) -> Option<User>` to the `Client`
- you iterate with on the current `Org::members`

 From a product perspective, it makes sense to require a `User` to register an org, so the alternative `#1` would make sense, be more convenient, and more efficient. @cloudhead thoughts?

xla:
> ####
> 
>     * A requirement is added when registering an org where the author must be a registered user (i.e, a user must exist whose account id equals the pub key of the author).

I was under the impression that this has been already established as an requirement and is part of the spec.

NunoAlexandre:
Nop:

![Screenshot from 2020-04-24 09-51-30](https://user-images.githubusercontent.com/7075260/80188348-53fee080-8611-11ea-95e2-89a233891faf.png)

xla:
@NunoAlexandre It says "a is associated with a registered user:. Is that not exactly the semantics we stated above?

NunoAlexandre:
@xla that's right. I _really_ need new glasses.

Will go ahead with updating the registry to follow the spec and adding storing the user id instead of the pub key in the org::members vec.

geigerzaehler:
The spec is ambiguous. In the `register-member` transaction it explicitly talks about registering an _account ID_ as a member. Can we raise an issue there?

cloudhead:
[...]

PR #400, opened 2020-04-24
Make Org members Vec<Id> instead of Vec<AccountId>

### LINKED ISSUE #397: Require RegisterOrg tx author to be a registered user
### Context

Found that we miss respecting the validation _"α is associated with a registered user u."_ when registering an org while discussing solutions for https://github.com/radicle-dev/radicle-registry/issues/396.

Screenshot from 2020-04-24 09-51-30

_Originally posted by @NunoAlexandre in https://github.com/radicle-dev/radicle-registry/issues/396#issuecomment-618862275_

### Changes

When registering an org:

1. we require there to be a registered user with an account id equal to the public key of the tx author and fail otherwise.
2. store the user id as the single org member. This requires changing Org::members to be a list of Ids instead of a list of account ids.

NunoAlexandre:
Closes #397 #396 

We check that authors registering an org are
registered users, which was a missing validation step (#397). 

That also means that to authorize actions on orgs, we now need to first lookup the user
associated with the tx author. 

When it comes to testing, now most tests need to have the tx author registering itself as a user.

:warning: The changes of this PR breaks the ffnet. Do NOT merge it to master. We have to think of a way to make these changes available to the upstream without affecting the chain network users.

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -42,6 +42,9 @@ pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
             let run_cmd = args.run_cmd();
             run_cmd.init(&version)?;
             run_cmd.update_config(&mut config, spec_factory, &version)?;
+            config.execution_strategies.block_construction =
+                sc_client::ExecutionStrategy::NativeElseWasm;
+
             if unsafe_rpc_external {
                 // Allow all hosts to connect
                 config.rpc_cors = None;
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -150,12 +150,8 @@ decl_module! {
         pub fn register_project(origin, message: message::RegisterProject) -> DispatchResult {
             let sender = ensure_signed(origin)?;
 
-            let org = match store::Orgs::get(message.org_id.clone()) {
-                None => return Err(RegistryError::InexistentOrg.into()),
-                Some(o) => o,
-            };
-
-            if !org.members.contains(&sender) {
+            let org = store::Orgs::get(message.org_id.clone()).ok_or(RegistryError::InexistentOrg)?;
+            if !org_has_member_with_account(&org, sender) {
                 return Err(RegistryError::InsufficientSenderPermissions.into());
             }
 
@@ -186,10 +182,11 @@ decl_module! {
         pub fn register_org(origin, message: message::RegisterOrg) -> DispatchResult {
             let sender = ensure_signed(origin)?;
 
-            match store::Orgs::get(message.org_id.clone()) {
-                None => {},
-                Some(_) => return Err(RegistryError::DuplicateOrgId.into()),
+            if store::Orgs::get(message.org_id.clone()).is_some() {
+                return Err(RegistryError::DuplicateOrgId.into());
             }
+
+            let user = get_user_with_account(sender).ok_or(RegistryError::AuthorHasNoAssociatedUser)?;
 
             let random_account_id = AccountId::unchecked_from(
                 pallet_randomness_collective_flip::Module::<T>::random(
@@ -199,7 +196,7 @@ decl_module! {
 
             let new_org = state::Org {
                 account_id: random_account_id,
-                members: vec![sender],
+                members: vec![user.id],
                 projects: Vec::new(),
             };
             store::Orgs::insert(message.org_id.clone(), new_org);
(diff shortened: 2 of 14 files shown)


test-utils/src/lib.rs
```diff
@@ -141,6 +141,27 @@ pub fn key_pair_from_string(value: impl AsRef<str>) -> ed25519::Pair {
     ed25519::Pair::from_string(format!("//{}", value.as_ref()).as_str(), None).unwrap()
 }
 
+/// Create a key pair derived from the given value and associate
+/// its account id with a User in the registry using the said value,
+/// lowecased, as its id. For accounts that are not 'Alice', a
+/// transfer is done to them to have enough funds to pay for the
+/// tx fee when registering the user.
+pub async fn key_pair_with_associated_user(client: &Client, value: &str) -> ed25519::Pair {
```
geigerzaehler, line 149: Instead of expecting a `value` we could randomly generate it in the function and return a `(ed25519::Pair, UserId)` pair. this would eliminate the need for the caller having to provide the argument.

test-utils/src/lib.rs
```diff
@@ -141,6 +141,27 @@ pub fn key_pair_from_string(value: impl AsRef<str>) -> ed25519::Pair {
     ed25519::Pair::from_string(format!("//{}", value.as_ref()).as_str(), None).unwrap()
 }
 
+/// Create a key pair derived from the given value and associate
+/// its account id with a User in the registry using the said value,
+/// lowecased, as its id. For accounts that are not 'Alice', a
+/// transfer is done to them to have enough funds to pay for the
+/// tx fee when registering the user.
```
geigerzaehler, line 148: The doc is quite complex. How about this?
```suggestion
/// Create a key pair derived from `value` and register a user with that key pair.
/// The user name is `value.to_lowercase()`. Ensures that the account for the key
/// pair is equipped with at least 100.000 RAD.
```

runtime/tests/project_registration.rs
```diff
@@ -228,17 +228,15 @@ async fn register_project_with_bad_checkpoint() {
...
-    transfer(&client, &god_actor, bad_actor.public(), 1000).await;
+    let good_actor = key_pair_with_associated_user(&client, "Alice").await;
+    let bad_actor = key_pair_with_associated_user(&client, "BadActor").await;
 
     // The good actor creates an org, of which becomes its single member.
     let org_id = random_id();
     let register_org = message::RegisterOrg {
         org_id: org_id.clone(),
     };
-    submit_ok(&client, &god_actor, register_org.clone()).await;
```
geigerzaehler, line 241: :joy_cat:

node/src/command.rs
```diff
@@ -42,6 +42,9 @@ pub fn run(version: VersionInfo) -> sc_cli::Result<()> {
             let run_cmd = args.run_cmd();
             run_cmd.init(&version)?;
             run_cmd.update_config(&mut config, spec_factory, &version)?;
+            config.execution_strategies.block_construction =
+                sc_client::ExecutionStrategy::NativeElseWasm;
+
```
geigerzaehler, line 47: This is only here to make the tests pass. Warrants further discussion when we’ve solved the upgrade dilemma.

runtime/src/registry.rs
```diff
@@ -399,6 +391,26 @@ decl_module! {
...
+pub fn get_user_with_account(account_id: AccountId) -> Option<User> {
+    store::Users::iter()
+        .find(|(_, user)| user.account_id == account_id)
+        .map(|(id, user)| User::new(id, user))
+}
+
+/// Check whether the user associated with the given account_id is a member of the given org.
+/// Return false if the account doesn't have an associated user or if said user is not a member
+/// of the org.
+pub fn org_has_member_with_account(org: state::Org, account_id: AccountId) -> bool {
```
geigerzaehler, line 407: We can make `org` a reference which would eliminate some `.clone()`s. (In general, we should use references if possible and the value is not `Copy`)

test-utils/src/lib.rs
NunoAlexandre: Better :ok_hand: thanks!

runtime/tests/project_registration.rs
NunoAlexandre: :angel:

node/src/command.rs
NunoAlexandre: It's not _just_ to make the tests pass, it's to have the system actually behaving accordingly to the latest features without having us regenerate the wasm code. But yeah, the warrant is important and this is not what we want to have in.

runtime/src/registry.rs
NunoAlexandre: You are right, will adjust!

NunoAlexandre: Done.

test-utils/src/lib.rs
NunoAlexandre: Done. I had considered doing that initially but since that involved more changes I didn't go that route, but I agree that it's worth the change. Done everywhere else key_pair_with_associated_user is used.

NunoAlexandre:
@geigerzaehler this branch is now pointing to `upstream` (see #404). Could you now approve this PR?

Issue #382, opened 2020-04-21
Collect insights on FFnet nodes that are not run by us


geigerzaehler:
To better judge the health of the ffnet and the software that is running it we need to collect data from nodes that are not run by us. Question we would like to answer are
* How many nodes are participating in the network?
* Which version of the software are they running?
* Are they achieving consensus?

These are all questions that [substrate-telemetry](https://github.com/paritytech/substrate-telemetry) can be answer. Unless we have a better solution we should enable it again.

NunoAlexandre:
This is useful and if telemetry offers that we should go ahead. I don't know of any other solution.

CodeSandwich:
I agree that the substrate telemetry checks all the boxes despite being almost zero effort.

I can't see anything wrong with using telemetry.polkadot.io server as our default monitoring tool.

My proposition is to restore the `--telemetry-endpoints` and `--name` flags and to add `--no-default-telemetry`. The old flags would work like before, but not passing `--no-default-telemetry` would push `wss://telemetry.polkadot.io/submit/` onto the list of endpoints.

geigerzaehler:
You proposal sounds sensible. But I don’t see a reason to have the `--telemetry-endpoints` option. I doubt that user want to run their on telemetry endpoints.

PR #401, opened 2020-04-24
Add polkadot telemetry

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/382. The registry.radicle.xyz update: https://github.com/radicle-dev/registry.radicle.xyz/pull/34.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -23,6 +23,7 @@ use radicle_registry_runtime::{
     AccountId, BalancesConfig, GenesisConfig, SudoConfig, SystemConfig,
 };
 use sc_service::GenericChainSpec;
+use sc_telemetry::TelemetryEndpoints;
 use sp_core::{crypto::CryptoType, Pair};
 use std::convert::TryFrom;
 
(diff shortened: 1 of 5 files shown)


node/src/cli.rs
```diff
@@ -86,6 +86,14 @@ pub struct Arguments {
     /// Bind the prometheus metrics endpoint to 0.0.0.0 on port 9615
     #[structopt(long)]
     prometheus_external: bool,
+
+    /// Human-readable name for this node to use for telemetry
+    #[structopt(long, value_name = "NAME")]
+    name: Option<String>,
+
+    /// Disable sending telemetry data to the polkadot server
```
geigerzaehler, line 94:
```suggestion
    /// Disable sending telemetry data to https://telemetry.polkadot.io/
```

node/src/chain_spec.rs
```diff
@@ -78,16 +79,16 @@ fn devnet() -> ChainSpec {
                 .parse()
                 .expect("Parsing a genesis peer address failed"),
         ],
-        None,           // telemetry endpoints
+        polkadot_telemetry(enable_telemetry),
```
geigerzaehler, line 82: I’d prefer to not send telemetry data for the devnet to avoid additional load on an external service. Since we don’t expect anyone else to participate in the devnet we also don’t need to gain any insight.

node/src/cli.rs
CodeSandwich: Fixed

node/src/chain_spec.rs
CodeSandwich: Fixed

PR #415, opened 2020-05-01
cli: Move from 'local account' to 'key-pair'

### LINKED ISSUE #414: cli: Differentiate between “local” accounts and on-chain accounts in terminology
At the moment the CLI uses the term “account” for two things.

First, for accounts on the block chain. A blockchain account is a balance and a nonce identified by a public key and the account can be controlled by the corresponding private key. The user interacts with these accounts through `account show` and `account transfer`

Second, we us “account” for key pairs stored locally on disk. These key pairs are identified by a name. The user interacts with them through `account generate` and `account list`.

Mixing the terminology leads to [user confusion](https://radicle.community/t/spinning-up-a-testnet-node-first-user-experience/196/3?u=geigerzaehler). 

@cloudhead is suggesting to use the term “key pair” for accounts stored on disk. This would mean we replace the  respective account commands with `key-pair generate` and `key-pair list`.

NunoAlexandre:
Closes #414 

- Disambiguate the terminology 'account' and 'local account' by now
calling local accounts key-pairs;
- Move key-pair related commands to a new command group 'key-pair'
- Update the README and the CHANGELOG
- Migrate the local accounts.json file to key-pairs.json

**Test required:**
 Pull this branch and run either of the `key-pair` sub-commands. Doing so will trigger a migration to rename the `accounts.json` file to `key-pairs.json`. I faced a permissions issue myself, where the `accounts.json` file permissions seem to have been changed somehow (recent system upgrade?).

**Note**: Docs companion PR: https://github.com/radicle-dev/registry.radicle.xyz/pull/37

### DIFF
--- a/cli/src/command/account.rs
+++ b/cli/src/command/account.rs
@@ -16,18 +16,12 @@
 //! Define the commands supported by the CLI related to Accounts.
 
 use super::*;
-use crate::account_storage;
 
 /// Account related commands
 #[derive(StructOpt, Clone)]
 pub enum Command {
-    /// Show information for a local, user's, or org's account
+    /// Show account information.
     Show(Show),
-    /// Generate a local account and store it on disk.
-    /// Fail if there is already an account with the given `name`
-    Generate(Generate),
-    /// List all the local accounts.
-    List(List),
     /// Transfer funds from the author to a recipient account.
     Transfer(Transfer),
 }
@@ -37,8 +31,6 @@ impl CommandT for Command {
     async fn run(self) -> Result<(), CommandError> {
         match self {
             Command::Show(cmd) => cmd.run().await,
-            Command::Generate(cmd) => cmd.run().await,
-            Command::List(cmd) => cmd.run().await,
             Command::Transfer(cmd) => cmd.run().await,
         }
     }
@@ -46,7 +38,7 @@ impl CommandT for Command {
 
 #[derive(StructOpt, Clone)]
 pub struct Show {
-    /// The account's SS58 address or the name of a local account.
+    /// The account's SS58 address or the name of a local key pair.
     #[structopt(
         value_name = "address_or_name",
         parse(try_from_str = parse_account_id),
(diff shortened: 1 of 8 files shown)


cli/src/command/account.rs
```diff
@@ -16,18 +16,12 @@
 //! Define the commands supported by the CLI related to Accounts.
 
 use super::*;
-use crate::account_storage;
 
 /// Account related commands
 #[derive(StructOpt, Clone)]
 pub enum Command {
-    /// Show information for a local, user's, or org's account
+    /// Show account information for a local key-pair, a user's, or org's account.
```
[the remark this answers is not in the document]
NunoAlexandre: Perfect :ok:

NunoAlexandre:
Thanks, @CodeSandwich. I have included your rewording in the first commit.

cli/src/command/account.rs
```diff
@@ -16,18 +16,12 @@
 //! Define the commands supported by the CLI related to Accounts.
 
 use super::*;
-use crate::account_storage;
 
 /// Account related commands
 #[derive(StructOpt, Clone)]
 pub enum Command {
-    /// Show information for a local, user's, or org's account
+    /// Show account information for a user, an org or a local key-pair.
```
geigerzaehler, line 23: Shouldn’t this just say “show a key pair”? This doesn’t show user or org info.

cli/src/command/account.rs
```diff
@@ -37,16 +31,14 @@ impl CommandT for Command {
...
-            Command::List(cmd) => cmd.run().await,
             Command::Transfer(cmd) => cmd.run().await,
         }
     }
 }
 
 #[derive(StructOpt, Clone)]
 pub struct Show {
-    /// The account's SS58 address or the name of a local account.
+    /// The account's SS58 address or the name of a local key-pair.
```
geigerzaehler, line 41: Shouldn’t it _not_ be hyphenated?

cli/src/command/account.rs
NunoAlexandre: It doesn't show a key-pair, it shows the balance of address of an account, i.e., account info. Now, the goal is to be explicit that this supports/works just the same for a user's account, an org's account or a local key pair's account.

cli/src/command/account.rs
NunoAlexandre: You are right: https://web.cs.ucdavis.edu/~rogaway/papers/hyphen.htm

cli/src/command/account.rs
geigerzaehler: Yes, you’re right. I mixed this up with `key-pair show`. However this still might be confusing. There’s no conceptual distinction between user, org and key pair accounts. In fact the relation ship between each of those entities and an account is very much different. Plus there are accounts not related to any of these entities or related to multiple.

To make this clearer let’s just say that it show account information. And there are two ways to identify the account: By SS58 address and key pair name.

NunoAlexandre: That's good! Fixed.

NunoAlexandre:
@geigerzaehler thanks for your review! Can you have another look?

cli/src/key_pair_storage.rs
```diff
@@ -76,65 +76,78 @@ pub enum ReadingError {
...
+    let path = path_buf.as_path();
     if !path.exists() {
         std::fs::write(path, b"{}").map_err(WritingError::IO)?;
     }
 
     Ok(path_buf)
 }
 
+fn build_path(filename: &str) -> Result<PathBuf, Error> {
+    let dir = dir()?;
```
geigerzaehler, line 146: Suggestion: `dir` is only used here. might make sense just to inline it.

PR #416, opened 2020-05-01
cli: Report key-pairs.json file location on error

### LINKED ISSUE #412: cli: Report location of accounts.json in errors
When the user gets any `account.json` processing error, he doesn't get its location and can't fix the problem.

E.g. our user got a `Permission denied` error, because the whole directory had restricted access. He had problems fixing it, because he didn't know where the file was and which directory should be `chmod`ed.

NunoAlexandre:
Closes #412

### DIFF
--- a/cli/src/key_pair_storage.rs
+++ b/cli/src/key_pair_storage.rs
@@ -44,16 +44,24 @@ pub enum Error {
     AlreadyExists(),
 
     /// Failed to write to the key-pairs file
-    #[error("Failed to write to the key-pairs file")]
+    #[error("{}", io_error_message("write"))]
     FailedWrite(#[from] WritingError),
 
     /// Failed to read the key-pairs file
-    #[error("Failed to read the key-pairs file")]
+    #[error("{}", io_error_message("read"))]
     FailedRead(#[from] ReadingError),
 
     /// Could not find a key pair with the given name
     #[error("Could not find a key pair with the given name")]
     NotFound(),
+}
+
+fn io_error_message(action: &str) -> String {
+    let path_info = match build_path(FILE) {
+        Ok(x) => format!("{:?}", x),
+        Err(e) => format!("{}", e),
+    };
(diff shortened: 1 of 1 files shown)


cli/src/key_pair_storage.rs
```diff
@@ -26,37 +26,45 @@ use std::io::Error as IOError;
...
     NotFound(),
 }
 
-/// Possible errors when writing to the accounts file.
+fn io_error_message(action: &str) -> String {
+    let path_info = match build_path(FILE) {
+        Ok(x) => format!("{:?}", x),
+        Err(e) => format!("{}", e),
+    };
+    format!("Failed to {} the key-pairs file: {}", action, path_info)
```
geigerzaehler, line 64: In the issue that was reported the problem was the `~/.local` was not writable so this message would not be helpful. We also don’t take into account read errors of the old path during migration. My suggestion would be to attach the path to the error.

NunoAlexandre: >  My suggestion would be to attach the path to the error.

Not sure if I get it. This chunk is doing that, right? When we fail to read/write the `key-pairs.json` file we try to print the path; if we can't build it, we print why that is. Remember that the file will be under `~/.local` on Linux, not on macOS nor Windows.

geigerzaehler: Sorry, this is not super clear indeed. If migrating the storage in `get_or_create_path` fails because `accounts.json` is not readable then we print “Failed to write .../key-pairs.json`. If creating the directory `~/.local/share` fails we print `Permission denied error` without any path information. This means that there are still cases where the user does not have enough information to fix the issue. (In particular it does not resolve the problem that was initially reported)

The only way I see to address this is to add the path information to the error struct when the error is handled from the `std::fs` functions.

NunoAlexandre: Thanks, I understand now. 

As for the scope of this PR itself, what do you think?

I will address the issue reported :+1:

PR #423, opened 2020-05-05
Fix runtime compilation for wasm

### LINKED ISSUE #418: Building runtime as Wasm fails
Building the runtime for the Wasm target (with `./scripts/build-genesis-runtime-wasm` for example) fails. This blocks us from updating the runtime. 

The compilation error comes from the `core` crate referencing `std` even if we compile it for the `no_std` environment and without default features.

To prevent this from happening in the future we should also ensure that CI catches this. The following line is supposed to check that but it does not. https://github.com/radicle-dev/radicle-registry/blob/b2a1019d7c7a3c4c166dbb8b1a67b023e1ee74d3/ci/run#L52

geigerzaehler:
Fixes #418

* We fix the compilation errors from `./scripts/build-genesis-runtime-wasm`. These errors were due to `radicle-registry-core` as a dependency of `runtime` using `std`. The offending code were `std::convert` and the `thiserror` annotations.

* We enforce building the runtime with `no_std` on CI by enabling the `no-std` feature and raising a compilation error if the `std` feature is also set.

* To make the check pass we ensure that the `std` feature is not implicitly set by the dev dependencies when building the runtime. To achieve this we remove all dev dependencies from `runtime` and move the tests that required these dependencies into the `runtime-tests` package.


runtime/src/lib.rs
```diff
@@ -23,6 +23,9 @@
 #![recursion_limit = "256"]
 #![feature(alloc_prelude)]
 
+#[cfg(all(feature = "std", feature = "no-std"))]
+std::compile_error!("Features \"std\" and \"no-std\" cannot be enabled simultaneously. Maybe a dependency implicitly enabled the \"std\" feature.");
```
CodeSandwich, line 27: Why do we need the `no-std` feature, isn't "lack of `std`" enough? For example substrate does that.

geigerzaehler: The problem is that the `std` feature can be enabled for the runtime even when it with `--no-default-features`. You can find a detailed explanation in [this comment on #418](https://github.com/radicle-dev/radicle-registry/issues/418#issuecomment-623506052). To prevent this we introduce the `no-std` feature which enforces that the `std` feature is not enabled inadvertently.

Issue #420, opened 2020-05-04
Fix error message when migrating accounts.json to key-pairs.json


NunoAlexandre:
- If migrating the storage in `get_or_create_path` fails because `accounts.json` is not readable then we print “Failed to read .../key-pairs.json”. 
- If creating the directory `~/.local/share/radicle-registry-cli` fails we print:
```
Error: Failed to write to the key-pairs file: "/home/nuno/.local/share/radicle-registry-cli/key-pairs.json"
  Caused by: Permission denied (os error 13)
``` 

This means that there are still cases where the user does not have enough information to fix the issue. (In particular it does not resolve the problem that was initially reported)

The only way I see to address this is to add the path information to the error struct when the error is handled from the `std::fs` functions.

_Originally posted by @geigerzaehler in https://github.com/radicle-dev/radicle-registry/pull/416/files_

NunoAlexandre:
Considering the `get_or_create_path` fn:

``` rust
fn get_or_create_path() -> Result<PathBuf, Error> {
    let path_buf = build_path(FILE).expect("build path failed");

    let old_path = build_path("accounts.json").expect("build old_path failed");
    if old_path.exists() {
        println!("=> Migrating the key-pair storage to the latest version...");
        std::fs::rename(old_path, &path_buf).map_err(WritingError::IO).expect("rename failed");
        println!("✓ Done")
    }

    let path = path_buf.as_path();
    if !path.exists() {
        std::fs::write(path, b"{}").map_err(WritingError::IO).expect("write {} failed");
    }

    Ok(path_buf)
}
```

Two cases I tested, running `key-pair list`:

1. Set no permissions at all on the `~/.local/share/radicle-registry-cli` dir

 `if old_path.exists()` doesn't check so it doesn't even attempt to migrate. It then fails on `std::fs::write`, which is executed because `path` doesn't exist (again, because of permissions).

It seems like error-handling wasn't a thing when these libraries were written :thinking: 

The error message is totally sensible for users, since the command is to `list` (read), and it fails to write (sensible internally):

```
Error: Failed to write the key-pairs file: "/home/nuno/.local/share/radicle-registry-cli/key-pairs.json"
  Caused by: Permission denied (os error 13)
```

2. set no permissions at all on the `accounts.json` file

[...]

PR #425, opened 2020-05-05
cli: Check key-pair storage dir permissions

NunoAlexandre:
Closes: https://github.com/radicle-dev/radicle-registry/issues/420

### The problem

We have received [feedback from our user Alex Good](https://radicle.community/t/spinning-up-a-testnet-node-first-user-experience/196/6)  that got the following error when generating a key pair. 

```
Failed to read accounts file: Permission denied
```

This error, at this latest revision, would be: 

```
Failed to write the key-pairs file: "/home/<user>/.local/share/radicle-registry-cli/key-pairs.json" 
    Caused by: Permission denied (os error 13)
```

Although now we point out the directory we are working on, the error is incorrect when the real issue
is that the dir, ~/.local/share/radicle-registry-cli (on linux), is not readable or writable (can not be created). 

### Solution

In this commit, add `Error::InsufficientDirPermissions` to subset those specific cases that
have to do with the directory alone. 

A failed attempt to write the key-pairs.json file due to lacking permissions at the dir level will
still fail with the error reported above, which is fine since it shows the real failed intent and the whole path that must be writable.

The motivating issue, #420, focused on capturing the proper failing file when renaming `accounts.json` to `key-pairs.json`. However, I have never tested that being an issue, both when the dir and the `accounts.json` have no permissions at all.

### DIFF
--- a/cli/src/key_pair_storage.rs
+++ b/cli/src/key_pair_storage.rs
@@ -65,11 +65,12 @@ pub enum Error {
 }
 
 fn io_error_message(action: &str) -> String {
-    let path_info = match build_path(FILE) {
-        Ok(x) => format!("{}", x.display()),
-        Err(e) => format!("{}", e),
-    };
-    format!("Failed to {} the key-pairs file: '{}'", action, path_info)
+    let path = build_path(FILE);
+    format!(
+        "Failed to {} the key-pairs file: '{}'",
+        action,
+        path.display()
+    )
 }
 
 /// Possible errors when writing to the key-pairs file.
@@ -141,9 +142,10 @@ const FILE: &str = "key-pairs.json";
 /// it with an empty object so that it can be deserialized
 /// as an empty HashMap<String, KeyPairData>.
 fn get_or_create_path() -> Result<PathBuf, Error> {
-    let path_buf = build_path(FILE)?;
+    let path_buf = build_path(FILE);
+    dir_ready(path_buf.parent().unwrap().to_path_buf())?;
 
-    let old_path = build_path("accounts.json")?;
+    let old_path = build_path("accounts.json");
     if old_path.exists() {
         println!("=> Migrating the key-pair storage to the latest version...");
         std::fs::rename(old_path, &path_buf).map_err(WritingError::IO)?;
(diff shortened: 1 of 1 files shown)


cli/src/key_pair_storage.rs
```diff
@@ -150,17 +154,26 @@ fn get_or_create_path() -> Result<PathBuf, Error> {
...
 
-fn dir() -> Result<PathBuf, Error> {
-    let dir = BaseDirs::new()
+/// Ensure that the key-pair storage [dir()] is ready to be used.
+/// Fails with [Error::InsufficientDirPermissions] when said dir
+/// can not be read or can not be created.
+fn dir_ready() -> Result<PathBuf, Error> {
+    let dir = dir();
+    File::open(dir.as_path()).map_err(Error::InsufficientDirPermissions)?;
+    std::fs::create_dir_all(&dir).map_err(WritingError::IO)?;
```
geigerzaehler, line 170: Running the following command fails even though  my user has read, write, and execute permissions on `/tmp`.
~~~
XDG_DATA_HOME=/tmp cargo run -p radicle-registry-cli -- key-pair list
~~~

The reason is that `File::open("tmp/radicle-registry-cli")` fails because the directory does not exist yet.

To address this I’d suggest to map the `create_dir_all` error.
```suggestion
    std::fs::create_dir_all(&dir).map_err(|err| Error::CreateDirectoryFailed(err, dir))?;
```

together with
```rust
    #[error("Cannot create directory {1}")]
    CreateDirectoryFailed(#[source] IOError, PathBuf),
```

NunoAlexandre: Good catch! We need to check that we can read the directory anyway, so we shouldn't delete the File::open line. If the dir can't be read, neither list neither add will work.

So we could just swap the File::open and the create_dir_all lines and it should fly. I don't feel strong about adding that specific variant, as reporting to the user that the storage directory needs to have read and write access is sufficient.

What do you think?

NunoAlexandre: I pushed a fixup adding the new `CannotReadDirectory` and `CannotCreateDirectory` error variants, and swap the lines above to first try to create the dir and then check that it can be read. if it doesn't exist and we don't have permissions to create it, it fails with `CannotCreateDirectory`; if it exists but we can't read it, it fails with `CannotReadDirectory`.

cli/src/key_pair_storage.rs
```diff
@@ -150,17 +158,29 @@ fn get_or_create_path() -> Result<PathBuf, Error> {
     Ok(path_buf)
 }
 
+/// Build the path to the given file name under [dir_ready()].
 fn build_path(filename: &str) -> Result<PathBuf, Error> {
-    let dir = dir()?;
+    let dir = dir_ready()?;
```
geigerzaehler, line 163: Suggestion: Splitting `build_path` in three functions hampers readability through a lot of indirections. Why not 
```rust
fn build_path(filename: &str) -> Result<PathBuf, Error> {
    let dir = BaseDirs::new()
        .unwrap()
        .data_dir()
        .join("radicle-registry-cli");
    std::fs::create_dir_all(&dir).map_err(|err| Error::CannotCreateDirectory(err, dir.clone()))?;
    File::open(&dir).map_err(|err| Error::CannotReadDirectory(err, dir.clone()))?;
    let path = dir.join(filename);
    Ok(path)
}
```

NunoAlexandre: To fix the other issue you reported by chat, where `io_error_message` was effectful through `build_path`, I have changed `build_path` to be pure, so your suggestion does not apply anymore.

In any case, for me it's a trade-off. I prefer to have one more indirection but having very self-contained functions than flattening a lot of different things within one same function.

NunoAlexandre:
@geigerzaehler everything is addressed, including the remark about the side effects on `io_error_message`. Have squashed the commits and rebased.

cli/src/key_pair_storage.rs
```diff
@@ -133,9 +142,10 @@ const FILE: &str = "key-pairs.json";
 /// it with an empty object so that it can be deserialized
 /// as an empty HashMap<String, KeyPairData>.
 fn get_or_create_path() -> Result<PathBuf, Error> {
-    let path_buf = build_path(FILE)?;
+    let path_buf = build_path(FILE);
+    dir_ready(path_buf.parent().unwrap().to_path_buf())?;
```
NunoAlexandre, line 146: Note that we know that build_path always has a parent, thus the wild unwrap here.

Issue #422, opened 2020-05-04
Add transactions in best block metric


CodeSandwich:
This metric tells how many transactions are in the best block. This allows us to monitor the network performance and if the blocks are full.

geigerzaehler:
I guess block size in bytes would also be interesting and a better metric for determining if a block is full. Do you intend to implement this metric as a gauge that exposes the number of transaction in the most recent block or as a counter that accumulates the number of transactions in every block.

CodeSandwich:
> I guess block size in bytes would also be interesting and a better metric for determining if a block is full. 

AFAIK there are 2 limits: size of the block and number of transactions. I agree, adding size will be valuable.

> Do you intend to implement this metric as a gauge that exposes the number of transaction in the most recent block or as a counter that accumulates the number of transactions in every imported block?

I intend to implement a gauge which shows the size of the most recent best block. Having a global counter would be tricky when reorgs start happening. Anyway while accumulating could give us an interesting insight into the total chain history, it probably wouldn't be a good indicator of ongoing issues.

Issue #424, opened 2020-05-05
Add size of the best block metric


CodeSandwich:
This metric tells how big in bytes is the best block. This allows us to monitor the network performance and if the blocks are full.

Issue #409, opened 2020-04-29
Add block timestamp offset metric


CodeSandwich:
This metric tells what's the difference between a successfully imported best block and the machine's wall time. This gives us insight into the propagation times and possible clock differences, which may affect consensus.

CodeSandwich:
When I've implemented it, I've noticed that this metric is useless: the only timestamp saved in a block is the time when its mining was started. We don't know when propagation of the block is started, so we can't tell how long does it take.

@geigerzaehler @NunoAlexandre I can't think of any reasonable solution, do you have any ideas? If not, I'll close this issue.

geigerzaehler:
I guess you’re right @CodeSandwich. I can’t think of any valuable insight this data would give us.

PR #430, opened 2020-05-06
Add best block size metrics

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/422 and https://github.com/radicle-dev/radicle-registry/issues/424. 

Unfortunately this PR is rather big, because some helpers had to be modified, they simply  didn't make sense anymore in the context of updating two gauges from a single worker task.

I've also made the gauge generator generic, it was needed for https://github.com/radicle-dev/radicle-registry/issues/409, but it seems unsolvable, so I've reverted it. I like this genericness and it may be useful soon, so I've kept it. The current master of substrate already reexports the whole `prometheus`.

### DIFF
--- a/node/src/service.rs
+++ b/node/src/service.rs
@@ -23,11 +23,16 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
+// TODO remove in favor of substrate_prometheus_endpoint::prometheus after substrate upgrade
+use prometheus::core::Atomic;
+use sc_client::BlockImportNotification;
 use sc_client::{light::blockchain::AuxStore, BlockchainEvents as _, LongestChain};
 use sc_executor::native_executor_instance;
 use sc_service::{AbstractService, Configuration, Error, ServiceBuilder};
 use sp_inherents::InherentDataProviders;
-use substrate_prometheus_endpoint::{Gauge, U64};
+use sp_runtime::generic::BlockId;
+use sp_runtime::traits::Block as _;
+use substrate_prometheus_endpoint::{Gauge, Registry, U64};
 
 use crate::pow::{blake3_pow::Blake3Pow, config::Config, dummy_pow::DummyPow, Difficulty};
 use radicle_registry_runtime::{registry::AuthoringInherentData, AccountId, Block, RuntimeApi};
@@ -142,8 +147,7 @@ pub fn new_full(
     let block_import = import_setup.expect("No import setup set for miner");
 
     let service = builder.build()?;
-    register_difficulty_metric(&service)?;
-    register_is_major_synced_metric(&service)?;
+    register_metrics(&service)?;
 
     if let Some(block_author) = opt_block_author {
         let client = service.client();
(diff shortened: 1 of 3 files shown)


node/src/service.rs
```diff
@@ -214,77 +218,139 @@ pub fn new_light(config: Configuration) -> Result<impl AbstractService, Error> {
...
         })?
         .build()?;
-    register_difficulty_metric(&service)?;
-    register_is_major_synced_metric(&service)?;
+    register_metrics(&service)?;
     Ok(service)
 }
 
-fn register_difficulty_metric<S>(service: &S) -> Result<(), Error>
+fn register_metrics<S>(service: &S) -> Result<(), Error>
```
geigerzaehler, line 225: Now that most of the code in this file is about metrics it might be time to split this into a separate module.

node/src/service.rs
```diff
@@ -214,77 +218,139 @@ pub fn new_light(config: Configuration) -> Result<impl AbstractService, Error> {
...
-    let gauge_help = "The difficulty of the best block in the chain";
-    let task_gen = |gauge: Gauge<U64>| {
-        let client = service.client();
-        client.import_notification_stream().for_each(move |info| {
-            let ready = futures::future::ready(());
-            if !info.is_new_best {
-                return ready;
+    let registry = match service.prometheus_registry() {
+        Some(registry) => registry,
+        None => return Ok(()),
```
geigerzaehler, line 232: Shouldn’t we at least log a warning here?

node/src/service.rs
```diff
@@ -214,77 +218,139 @@ pub fn new_light(config: Configuration) -> Result<impl AbstractService, Error> {
...
-    let gauge = substrate_prometheus_endpoint::register(new_gauge, &registry)
-        .map_err(|e| format!("failed to register metric gauge '{}': {}", gauge_name, e))?;
-    let task_name = format!("{}_metric_notifier", gauge_name);
-    service.spawn_task(task_name, task_gen(gauge));
-    Ok(())
+    substrate_prometheus_endpoint::register(gauge, &registry)
+        .map_err(|e| format!("failed to register metric gauge '{}': {}", gauge_name, e).into())
+}
+
+fn spawn_metric_task(
```
geigerzaehler, line 347: This function is only used once now. Since it’s tiny we should probably inline it.

node/src/service.rs
CodeSandwich: Yes, that's exactly the next step I wanted to take :D I want to create a separate PR for that, because it would be disruptive in this PR.

node/src/service.rs
NunoAlexandre: I like having small functions like this one. It's important to remember that functions were originally created not only to dry code but also to abstract blocks. Just my two cents :)

node/src/service.rs
CodeSandwich: Fixed

node/src/service.rs
CodeSandwich: I've missed one spot, where it should be used. Now it's in two places, but with additional metrics it will be more frequent.

PR #448, opened 2020-05-11
cli: Version the key_pair_storage file

### LINKED ISSUE #427: cli: Version the key-pairs.json file schema
Part of #435 

### Problem

Changes to the schema of the `key-pairs.json` file need to be migrated. For example, if we add a new field to `KeyPairData`, change the type of an existing field, etc.

Without knowing for a fact in which schema version the file is, all we can do is to try: Try to parse the file in some specific form, try to parse a specific field, etc,

### Proposal

Version the schema of `key-pairs.json` file so that we can use parse the specific variant for the stored version of the file.

For example, currently `KeyPairData` is 

``` rust
pub struct KeyPairData {
    pub seed: Seed,
}
```

Say we add a new field `foo: Foo` to it. When parsing the file, we know that if it's version 1 the field is not present, so some default or `None` is set, while if it's version 2 we can assume the field to be present.

To accomplish this, we can either:

- A. Change the type of the contents of `key-pairs.json` from `HashMap<String, KeyPairData>`, i.e, `{ "name: { "seed": _ } }` to ` pub struct KeyPairsFile { version: u8, key_pairs: HashMap<String, KeyPairData> }`, i.e., { "version: x, "key_pairs": { "name: { "seed": _ } } }`.

- B. Have a `key-pairs.version` file containing the schema version of key-pairs.json alone: `1`, `2`, etc. That would help to resolve which variant of the key-pairs.json contents type we should use to deserialize/parse the actual key-pairs.json file, or which file name we should be looking at, even, for example.
 
While option B would simplify the lookup of the version, it would also mean that to parse the key-pairs.json file we need to keep both files along. Option A has all truth in one single file, we just have to define that the version field will always be found in the exact same way, which is a fine requirement.

NunoAlexandre:
Closes #427 

### Changes

- Use json serde adjacent enums to easily serialize and deserialize a versioned `key-pairs.json` file.

- ~~Add a migrations pipeline, where every new migration will be piped in the right sequence. See `key_pair_storage::migrate()`.~~

- Other minor improvements

#### :thinking: :thought_balloon: to think about

~~As of now~~ Before https://github.com/radicle-dev/radicle-registry/pull/448/commits/ca98fdd0029a7377231046783159985006e63e5c, we preemptively attempted to migrate to a newer version of the schema in `list()`, as that is the method used by all other public methods of `key-pair-storage`. This need arises from the fact that only attempting to migrate if `parse_file` returns something besides the latest file schema is not good enough since `parse_file` calls `get_or_create_path()`, which - as it suggests - will create `key-pairs.json` if it doesn't exist yet. If the user has the very first version of the key pair storage, `accounts.json`, this will result in a new, empty `key-pairs.json` file being created, which will by default be in the latest version. This isn't a problem on master since we do this migration step within `get_or_create_path`, here moved to the migrations pipeline.

While this works and has some benefits, such as not requiring any work from the user, it is a bullet waiting to be shot at our feet, since it made correct by coding it so, not by design.

_How can we improve this_

Don't be smart. setting things up if not there yet but separate concerns by adding two new commands:

1. `key-pair init`
    - Setup the directory - create it, check permissions, etc
    - create the key-pairs file for the first time OR migrate it to the latest version
    Should only be run _once_, when the user has installed the CLI.

2. `key-pair migrate`
     Run when the user has updated the CLI to a newer version which requires the
     file schema to be migrated to a newer version alongside.

Doing so would allow us to separate the concerns and have the rest of the code handling them accordingly: If the file is not present, it would fail with an IO error saying the file doesn't exist or cant be read. If outdated, return an error stating precisely that. Note that doing this would also help us have less mixture between pure code and side-effects, and less mix between the path we are concerned with and setting things up along the way.

I would prefer to address this in a separate issue.

### DIFF
--- a/cli/src/key_pair_storage.rs
+++ b/cli/src/key_pair_storage.rs
@@ -28,7 +28,7 @@ use std::path::PathBuf;
 /// The data that is stored in the filesystem relative
 /// to a key pair. The name of the key pair is used as
 /// the key to this value, therefore not included here.
-#[derive(Serialize, Deserialize, Clone)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct KeyPairData {
     pub seed: Seed,
 }
@@ -109,13 +109,15 @@ pub fn add(name: String, data: KeyPairData) -> Result<(), Error> {
 
 /// List all the stored key-pairs.
 ///
-/// It can fail from IO and Serde Json errors.
+/// It can fail from IO errors or Serde Json errors.
+/// Attempts to migrate the key-pairs file if outdated.
 pub fn list() -> Result<HashMap<String, KeyPairData>, Error> {
-    let path_buf = get_or_create_path()?;
-    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
-    let key_pairs: HashMap<String, KeyPairData> =
-        serde_json::from_reader(&file).map_err(ReadingError::Deserialization)?;
-    Ok(key_pairs)
+    use {KeyStorageFile::*, VersionedFile::*};
+
+    match parse_file()? {
+        Unversioned(key_pairs) => Ok(key_pairs),
+        Versioned(V1 { key_pairs }) => Ok(key_pairs),
+    }
 }
 
 /// Get a key pair by name.
@@ -127,9 +129,9 @@ pub fn get(name: &str) -> Result<KeyPairData, Error> {
 }
 
 fn update(key_pairs: HashMap<String, KeyPairData>) -> Result<(), Error> {
+    let data = VersionedFile::V1 { key_pairs };
     let path_buf = get_or_create_path()?;
-    let new_content =
-        serde_json::to_string_pretty(&key_pairs).map_err(WritingError::Serialization)?;
+    let new_content = serde_json::to_string_pretty(&data).map_err(WritingError::Serialization)?;
     std::fs::write(path_buf.as_path(), new_content.as_bytes()).map_err(WritingError::IO)?;
     Ok(())
 }
@@ -147,15 +149,13 @@ fn get_or_create_path() -> Result<PathBuf, Error> {
     let path = path_buf.as_path();
     dir_ready(path.parent().unwrap().to_path_buf())?;
 
-    let old_path = build_path("accounts.json");
-    if !path.exists() && old_path.exists() {
-        println!("=> Migrating the key-pair storage to the latest version...");
-        std::fs::rename(old_path, path).map_err(WritingError::IO)?;
-        println!("✓ Done")
-    }
-
     if !path.exists() {
-        std::fs::write(path, b"{}").map_err(WritingError::IO)?;
+        let old_path = build_path("accounts.json");
+        if old_path.exists() {
+            std::fs::rename(old_path, path).map_err(WritingError::IO)?;
+        } else {
+            std::fs::write(path, b"{}").map_err(WritingError::IO)?;
+        }
     }
 
     Ok(path_buf)
@@ -183,4 +183,32 @@ fn dir() -> PathBuf {
         .unwrap()
         .data_dir()
         .join("radicle-registry-cli")
+}
+
+fn parse_file() -> Result<KeyStorageFile, Error> {
+    let path_buf = get_or_create_path()?;
+    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
+
+    serde_json::from_reader(&file).map_err(|e| ReadingError::Deserialization(e).into())
+}
+
+/// The possible file variants to be handled when deserialization [FILE].
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+enum KeyStorageFile {
+    /// The genesis, unversioned file variant.
+    Unversioned(HashMap<String, KeyPairData>),
+
+    /// A versioned file variant, to which we have moved to
+    /// in order to leverage backwards-compatibility.
+    Versioned(VersionedFile),
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "version")]
+enum VersionedFile {
+    #[serde(rename = "1")]
+    V1 {
+        key_pairs: HashMap<String, KeyPairData>,
+    },
 }
--- a/cli/src/command/key_pair.rs
+++ b/cli/src/command/key_pair.rs
@@ -55,6 +55,7 @@ impl CommandT for Generate {
         Ok(())
     }
 }
+
 #[derive(StructOpt, Clone)]
 pub struct List {}
 


cli/src/key_pair_storage.rs
```diff
@@ -109,13 +109,17 @@ pub fn add(name: String, data: KeyPairData) -> Result<(), Error> {
...
+/// It can fail from IO errors or Serde Json errors.
+/// Attempts to migrate the key-pairs file if outdated.
 pub fn list() -> Result<HashMap<String, KeyPairData>, Error> {
-    let path_buf = get_or_create_path()?;
-    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
-    let key_pairs: HashMap<String, KeyPairData> =
-        serde_json::from_reader(&file).map_err(ReadingError::Deserialization)?;
-    Ok(key_pairs)
+    use {FileSchema::*, VersionedSchema::*};
+    migrate()?;
```
CodeSandwich, line 116: Wouldn't it be cleaner to read whatever is possible, but do the migration during saving? We could avoid multiple rounds of loading and parsing. It would be also less surprising, `list` actually *overwrites* an existing file!

CodeSandwich: Right, you've addressed it in the PR description, I'm reading it now.

CodeSandwich: I think that the root issue may be that `get_or_create_path()` prepares storage. It adds a lot of hidden and tricky behavior, which is hard to wrap head around.

What if to load a file we did (pseudocode):
```rust
let file_schema = if "key-pair.json".exists() {
  parse_json::<FileSchema>("key-pair.json")?
}
else {
  if "accounts.json".exists() {
    parse_json::<FileSchema>("accounts.json")?
  else {
    FileSchema::default()
  }
};
keys: HashMap = file_schema.into_keys();
```
And to save a file:
```rust
let _ = fs::delete("accounts.json"); // Ignore errors!
let file_schema = FileSchema::new_in_latest_version(keys);
write_json(file_schema, "key-pair.json");
```

cli/src/key_pair_storage.rs
```diff
@@ -184,3 +181,64 @@ fn dir() -> PathBuf {
...
+    /// The genesis, unversioned file schema.
+    Unversioned(HashMap<String, KeyPairData>),
+
+    /// A versioned file schema, to which we have moved to
+    /// in order to leverage backwards-compatibility.
+    Versioned(VersionedSchema),
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "version", content = "value")]
```
CodeSandwich, line 240: The `content` field complicates the structure of the JSON, but it's not required, it can be dropped to flatten the file a bit.

cli/src/key_pair_storage.rs
NunoAlexandre: I don't dislike your proposal, but there are some properties I want to meet (some mentioned in the PR description).

I would just much rather prefer to have the renaming of the "accounts.json" file happening just like any other migration because that's what it is: a migration. In your proposal, when reading, we would need to handle all possible files we will ever have, and when writing we should (as in, to keep clear after our mess) delete all the legacy files IF we succeed to write to the new one. Note how this requires both the read and the write blocks to know about these legacy files and do more than read and write.

In that proposal, as said above, the code still has to handle more things that what it should concern about. I would like to have the code concerned with its specific purpose and not handling all old and new things: it should read the file; if it fails because it doesn't exist, it fails. It reads the file and it's outdated: fail, because the file should be migrated.

I like the idea of parsing any `FileSchema` and convert it to its latest variant. It's essentially the idea of `migrate` but without the renaming and the update bit. But then again, I would much rather have the renaming as part of the migration.

All in all, I see two alternatives:

1. We delegate control to the user

    We fail if `init` or `migration` are needed.

    This can be made user-friendly by having the _init_ bit in the docs and explain the migration also in the docs and in the release notes, which should only happen very occasionally at worse.

    This means those two new commands would be added, as well as new error variants covering and reporting these cases.

    This alternative doesn't feel, to me, any different than web apps asking me to manually reload the page to get the latest version, which I much rather prefer over having the tool reloading out of the blue for me.

2. We fail whenever we detect the storage to not be ready or the file to be outdated and attempt the migration ourselves.

   For example, in `list()` we would have:

   ``` rust
    pub fn list() -> Result<HashMap<String, KeyPairData>, Error> {
        use {FileSchema::*, VersionedSchema::*};

        migrate()?;
        match parse_file() {
            Err(NotSetUp) => {
                setup()?;
            ),
            Err(e) => {
                return Err(e);
            },
            Ok(Versioned(V0(key_pairs))) => { 
                return Ok(key_pairs)
            },
            Ok(outdated) => {
                migrate()?;
            }
        }
        // try again
        list()
    }
   ```

`get()`, `list`()`, `add(name: String, data: KeyPairData)`

The 2. alternative doesn't require the user to play a role but it goes back to the initial issue: list() shouldn't have to be doing this, even it knows how to.

So, yeah, I prefer the 1. option. It separates concerns, it's transparent, it organically gives context to errors, etc.

cli/src/key_pair_storage.rs
NunoAlexandre: I'd rather keep the nesting since it provides a cleaner json schema spec, where we don't have all but one entry being of some other type.

NunoAlexandre:
@CodeSandwich regarding the remark about having a variant in `enum VersionedSchema` to catch the case of a newer version than supported. As it is now, it would err with:

``` ruby
➜ radicle-registry-cli key-pair list 
Error: Failed to read the key-pairs file: '/home/nuno/.local/share/radicle-registry-cli/key-pairs.json'
  Caused by: data did not match any variant of untagged enum FileSchema
```

This error seems pretty good to me, I don't see the need to have a default variant to throw another, similar error.

cli/src/key_pair_storage.rs
CodeSandwich: > where we don't have all but one entry being of some other type.

Could you explain?

I'd rather keep the file format simple. If we drop the `content`, we'll end up with a flat file with an optional `version` field. It's as simple as it gets.

NunoAlexandre: >> where we don't have all but one entry being of some other type.
>
> Could you explain?

Flat is not always simpler, and I think this is a good example of that. Having it flat makes its appearance cleaner, but its JSON schema more complicated. 

Now, the file schema can be described as an object with two fields, `version` and `value`, an `integer` and an object of serialized `KeyPairData` values, aka, `Map<String, KeyPairData>` . If we flatten it, all the fields of the top-level object are `Map<String, KeyPairData>` besides one, named `version`, which is an integer.

> I'd rather keep the file format simple. If we drop the content, we'll end up with a flat file with an optional version field. It's as simple as it gets.

Why would we do that? The whole point of this work is to introduce it to better handle the backwards-compatibility requirement.

The only argument I see to flatten it is to flatten the file because it looks better. I think there are better arguments to keep it as is, and looking at the file is not necessarily one of the main worries (it is up to some point, for us devs, thus the PR were I pretty-printed it). Moreover, serde handles the nesting for us, so if any nesting introduces complexity, it is handled for free.

NunoAlexandre:
@CodeSandwich and I have discussed further regarding the discussion started here [0], which touches a part of the concern raised in the description of this PR. Please read both before proceeding.

I will explain the alternative each of us presents and re-iterate on the problem they address.

### The problem

Right now, be it on master and on this PR, we have a considerable amount of mixed concerns. For instance:

- on master, when building the path to the file on disk we attempt to migrate from the legacy `accounts.json` to the latest `key-pairs.json`

- both on master and here, when reading the file, we create the file it doesn't exist yet.

- here, we attempt to parse the file and, if it's not in the latest version of the schema, we try to migrate it, where migrating at this point means moving from `accounts.json` to the latest `key-pairs.json` if necessary, and move the file schema to the latest `VersionedSchema::V1` variant.
 This is handled in `fn list()`, _the_ function used by all other public functions of `key_pair_storage`.

### Alternative A
by @NunoAlexandre 

My goal is to separate concerns, not having functions building paths also migrating files, nor creating them, etc, but to separate those actions into their concerns: init and migrate.
Ideally, when the user installs the CLI for the first time, we would receive an event to which we would respond by setting up the directory for the storage as well as initializing the file. Since we don't have such a feature, what we started by doing was to have the code concerned with these bits every time it is executed, both when reading and writing to the file.

Here, I introduced this idea of a migration patch. Renaming `accounts.json` to the latest `key-pairs.json` was the first migration patch we introduced. The second one, here introduced, is to version the key-pairs.json file. Therefore, I see these patches as an ordered migration patches pipeline, where the migration as a whole can take the state from the very first version released to the one running on the user's machine.

Now, instead of having, say, `get_or_create_path()` concerned with the setting up the directory, concerned with the file existing or not, creating it or not, etc, we should (when reading (`list`) and writing (`add`) only be creating the path to the file (pure function) and parsing it. I.e, we should assume those predicates. We would do that by moving the concerns listed above to a `key-pair init` command that the user would run when they install the CLI for the first time. If something changes on disk after that (say, the file is deleted or the permissions are insufficient), the errors will report enough information as to what is happening. On our side, we just have to document what the user needs to do for the key-pair storage to operate correctly.

Within the same vibe, when we read the file and encounter an older file schema, instead of creating support to read and write all of them, I defend that we should just report that the storage is outdated and that it needs to be updated by running `key-pair migrate`. As I said above, a rather similar procedure to web apps asking the user to reload the page to update.

Note that we could run the migration ourselves when we parse an older file schema. However, doing so might lead to uncontextualized errors (for instance, writting errors when `list()`ing the key pairs, and it would mean, again, a problem of mixed concerns and a lack of transparency to the user.

In sum, my suggestion is:

- Offer `key-pair init`, that should be run when the user installs the CLI for the first time ever
- Have the code assuming that `key-pair init` has been executed, or fail with the appropriate, rich errors, consequence of skipping such mandatory step
- Have the code pass on older file schemas and notify the user that a migration is required
- Offer `key-pair migrate`, which will update the state on disk to the latest version. The user would run this command after updating or downgrading to a version that has an outdated file schema.
 

### Alternative B
by @CodeSandwich  

Instead of having this notion of migration, have
- on load/read, support all files ever used (accounts.json or key-pairs.json), as well as every schema. We would migrate the parsed data to the latest variant and operate on it.

- on writing, where we would be operating on the latest schema variant, we would write that to the disk, meaning we would lazily migrate the file until a write command is executed.


### My consideration

My concerns with alternative B are:

- it makes the notion of migration sparse: it would handle legacy files when reading and compute the data migration as separate things, where I prefer to see them as the same idea: migration patches.

- the file only gets migrated on the next write, which might lead to some contextualized error. Say the user has been reading just fine, and now when adding a new key pair, it fails with an error related to schemas. Probably not an issue in practice, though.

- we migrate the parsed data to the latest variant at every read (note the duplicated computation) where we could compute that once and update it on disk right away. 

- either we keep supporting all possible files on disk, or we rename them, or we converge to the latest one when writing, deleting the old ones if write succeeds. None of the alternatives seems as clean, hygienic, and encapsulated as my suggestion of having that done in a dedicated migration command.

Alternative A asks the user to do something for us, but that something contextualized what is going on, what might have gone wrong, and it takes barely any time from the user, while it helps us clean the code considerably.

@CodeSandwich can you please verify that I am explaining your suggestion fairly?

@geigerzaehler we appreciate your take on this :v:

CodeSandwich:
> can you please verify that I am explaining your suggestion fairly?

Let the volumes speak for themselves :D I'll try to balance that shortly.

### Addressing considerations towards B

> it would handle legacy files when reading and compute the data migration as separate things, where I prefer to see them as the same idea: migration patches.

The loading process can be still performed as a series of patches, nothing changes in this regard.

> we migrate the parsed data to the latest variant at every read (note the duplicated computation) where we could compute that once and update it on disk right away.

We're talking about microseconds in a CLI. The whole migration is moving a few fields and providing some hardcoded defaults. The current design is the  loads the file twice, which may be a bigger burden.

> either we keep supporting all possible files on disk, or we rename them, or we converge to the latest one when writing, deleting the old ones if write succeeds.

We still need logic to support for old files in the code, there's no difference. The encapsulation shouldn't be a problem either, we need a separate module for handling loading the files anyway.

### Considerations towards A

- We're bothering the user with our internal problems.

- We have 2 loading functions: a bulletproof one for migration and a picky one for actual loading

- At least in the current design loading files can modify them, which is a bad separation of concerns. The alternative A may fix that.

- The storage can be in an invalid state when not initialized or not migrated. This is an unnecessary complication for both us and the user if we have a file loader capable of migrating anyway.

cli/src/key_pair_storage.rs
CodeSandwich: :facepalm: now I get it. I think that we've made a mistake in the beginning by dumping all the keys in an unnamed key-value tuple. I think that we should fix that in `V1` by making it `V1 { keys: HashMap<String, KeyPairData> }`. Otherwise we won't be able to safely add any fields in the future! Even now we're risking that if the user has an account in legacy format named `version`, serde will fail.

NunoAlexandre: > I think that we've made a mistake in the beginning by dumping all the keys in an unnamed key-value tuple.

In the beginning, there were no use cases for more, so that was fine.

> I think that we should fix that in V1 by making it V1 { keys: HashMap<String, KeyPairData> }

That's basically what it is now, but it is `value` instead of `keys`. I am fine with renaming it to `keyPairs`.

> Otherwise we won't be able to safely add any fields in the future! Even now we're risking that if the user has an account in legacy format named version, serde will fail.

Exactly, that would happen if we don't nest using the `content` property.

CodeSandwich: > That's basically what it is now, but it is value instead of keys.

Not quite. If we want to e.g. add `checksum`, we'll end up with
```
{
  "version": "V1",
  "keys": {
    "keys": [<out_keys>],
    "checksum": "abcd1234"
  }
}
```

NunoAlexandre:
> ### Addressing considerations towards B
> > it would handle legacy files when reading and compute the data migration as separate things, where I prefer to see them as the same idea: migration patches.
> 
> The loading process can be still performed as a series of patches, nothing changes in this regard.
> 

Well, patches would be separated in nature, those related to the file system and those related to the file schema. As I said, it would make it a sparse concept.

> > we migrate the parsed data to the latest variant at every read (note the duplicated computation) where we could compute that once and update it on disk right away.
> 
> We're talking about microseconds in a CLI. The whole migration is moving a few fields and providing some hardcoded defaults. The current design is the loads the file twice, which may be a bigger burden.

You are right, I didn't mean to make the point being about performance but about the fact that we know something is outdated, we migrate it in-memory, and we leave it there. Think of a cutlery tray :fork_and_knife: :spoon: Say in that first we decide to have them organize as spoons, forks, knives. Then we agree to make it forks, spoons, knives. For me, it makes more sense to get in place right away, instead of waiting until a new washed piece of cutlery is dry and ready to be put in the right place.


> 
> > either we keep supporting all possible files on disk, or we rename them, or we converge to the latest one when writing, deleting the old ones if write succeeds.
> 
> We still need logic to support for old files in the code, there's no difference. The encapsulation shouldn't be a problem either, we need a separate module for handling loading the files anyway.

Of course, there is a difference. One has that logic in a dedicated, purposeful place to represent the migration we underwent. As in, in patch X we move this file from here to there. That is not the same as saying "we might have to read the file from here, there or somewhere else".

> 
> ### Considerations towards A
> * We're bothering the user with our internal problems.

Yes and no. Sure, I would prefer to not have to involve the user. But it's a trade-off: it makes the code simpler and better separated, it's a more transparent alternative, it provides context for free in case of errors, etc. Those are substantial perks that, compared to simply guide the user to run a command in under a second, shouldn't be dismissed so easily, imho.

> * We have 2 loading functions: a bulletproof one for migration and a picky one for actual loading

I don't understand what you mean by having 2 loading functions, neither by bulletproof and picky. Maybe because I am sure what you mean by loading. I get 'load' as 'reading & parsing', i.e, essentially calling `Result<FileSchema, Error> = serde_json::from_reader(&file)`. Now, when `fn list`ing we would return an error like `RequiresMigration` when the variant is not the latest version. When `fn migrate`ing, we would execute the pipeline, which would parse the file, apply the patches in memory, and finally update the file. So there is really only one loading function, the way of handling its output would differ depending on the purpose of the callers. 

> * At least in the current design loading files can modify them, which is a bad separation of concerns. The alternative A may fix that.

Exactly :+1: 

> * The storage can be in an invalid state when not initialized or not migrated. This is an unnecessary complication for both us and the user if we have a file loader capable of migrating anyway.

Note that I agreed before that we can indeed just run the migration ourselves if we detect the need, with the cost of either duplicating that logic in all call sites of the public functions of `key_pair_storage`, or by having it in `list()` as done in this branch now, since that function is used by all the others.

Another thing to consider is that reading that argument makes it sound like a daily chore, when, in fact, `init`ializing would only be done _once_ in the user's lifecycle and migrating only when a new version of the CLI introduces changes to the schema of the file, something that we can announce as needed on the release entry of said binary version.

cli/src/key_pair_storage.rs
NunoAlexandre: You didn't need to use "keys" twice to make it look worse, but with that example, I see value within flattening, so it's done: https://github.com/radicle-dev/radicle-registry/pull/448/commits/279568526e5862d1d95fcaad01a8ddb886135600

:+1:

NunoAlexandre:
@geigerzaehler @CodeSandwich I have dropped the part of the migrations from this branch to have a cleaner iteration and move that discussion aside. This way, here we only start supporting versioned schemas, updating the file on disk on the next key pair generated, like @CodeSandwich's proposal suggested.

Note that this approach does not address any of my concerns described in the description of this PR, which my `init` and `migrate` proposal would.

cli/src/key_pair_storage.rs
```diff
@@ -147,15 +149,13 @@ fn get_or_create_path() -> Result<PathBuf, Error> {
...
-        println!("✓ Done")
-    }
-
     if !path.exists() {
-        std::fs::write(path, b"{}").map_err(WritingError::IO)?;
+        let old_path = build_path("accounts.json");
+        if old_path.exists() {
+            std::fs::rename(old_path, path).map_err(WritingError::IO)?;
+        } else {
+            std::fs::write(path, b"{}").map_err(WritingError::IO)?;
```
geigerzaehler, line 157: With `b"{}"` we hard-code the serialization of an empty, unversioned key pair storage (i.e. `FileSchema::Unversioned(HashMap::new()))`. I think it would be less brittle to write the serialized version of the latest empty key pair storage.
```suggestion
            let empty = FileSchema::Versioned(VersionedSchema::V1 { key_pairs: HashMap::new() });
            std::fs::write(path, b"{}").map_err(WritingError::IO)?;
```
Of course that one line should be extracted.

cli/src/key_pair_storage.rs
```diff
@@ -184,3 +184,31 @@ fn dir() -> PathBuf {
...
+    let path_buf = get_or_create_path()?;
+    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
+
+    serde_json::from_reader(&file).map_err(|e| ReadingError::Deserialization(e).into())
+}
+
+/// The possible file schemas to be handled when deserialization [FILE].
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+enum FileSchema {
```
geigerzaehler, line 198: If find this name confusing because the struct does not represent the schema but the actual data (along side the schema information). I’d prefer to rename it to something like `KeyStorageFile`.

NunoAlexandre: Initially, I called it `File`, `FileContents`, and `FileFormat`, but found that using schema made writing the docs and looking at the problem more idiomatic. If we call it a `File`, I'd find it confusing that there are multiple variants of a file, while it sounds better to be me that a file might be in the form of any supported schema. In that light, `Unversioned` is a schema, whose contents if a flat HashMap; `V1` represents a `Versioned` file schema containing a single `key_pairs` field holding an HashMap.

Does that change your mind? I am happy to reconsider, just not convinced about that particular suggestion.

cli/src/key_pair_storage.rs
NunoAlexandre: Yeah... I wanted to simply call `update(HashMap::new())?`, but `update` calls `get_or_create_path`, wich would then call `update`, resulting in a stack overflow.

Going with your suggestion, we would also need to json serialize the empty V1:

```diff
-   std::fs::write(path, b"{}").map_err(WritingError::IO)?;
+   let empty = FileSchema::latest_empty();
+   let empty_json =
+      serde_json::to_string_pretty(&empty).map_err(WritingError::Serialization)?;
+   std::fs::write(path, empty_json).map_err(WritingError::IO)?;
```
I didn't find the added noise and complexity worth it. Do you disagree? I'd prefer to leave it as is and, when we tackle the follow-up issues reported in the description above, just call `update(HashMap::new())?`.

cli/src/key_pair_storage.rs
geigerzaehler: I find the term “schema” very confusing. Usually a schema refers to an abstract description of the shape of the data and not to a representation of the data itself. We could call it just `KeyStorage` or `VersionedKeyStorage`.

CodeSandwich: `DTO` is a commonly used term

NunoAlexandre: > I find the term “schema” very confusing. Usually a schema refers to an abstract description of the shape of the data and not to a representation of the data itself. We could call it just `KeyStorage` or `VersionedKeyStorage`.

Yeah, I get your point.

NunoAlexandre: Renamed it to `KeyStorageFile` and `VersionedFile` respectively.

PR #449, opened 2020-05-13
cli: extract key pair storage init

NunoAlexandre:
A humble follow up of the issue described in the description of #448

### Problem

We currently have `get_or_create_path()` that, as the name suggests, builds the path to the storage file but it also creates the storage file it doesn't exist, renames the legacy file if present, creates its parent directory, checks for permissions, etc. That means that every time we simply wan to get the path we have these side effects taking place, which makes the code less reusable and composable.

### Changes

Extract the 'init' code from `get_or_create_path()` to `init()`, which is run preemptively in `list()`, the function that all other pub fn use.

This grants us a better separation of concerns, where now the rest of the code can purely compute the file path and operate on it without the side effects involved.

I don't find this to be a definitive solution (See the linked [PR description](https://github.com/radicle-dev/radicle-registry/pull/448#issue-415912111)) but it's a small step forward in the right direction.

Note: The last commit improves the code organization of the `key_pair_storage_ file and fixes docs, which increases the diff. I suggest checking each commit individually.

### DIFF
--- a/cli/src/key_pair_storage.rs
+++ b/cli/src/key_pair_storage.rs
@@ -22,8 +22,35 @@ use std::collections::HashMap;
 use std::fs::File;
 use thiserror::Error as ThisError;
 
+use lazy_static::lazy_static;
 use std::io::Error as IOError;
 use std::path::{Path, PathBuf};
+
+lazy_static! {
+    /// The file where the key pairs are stored.
+    static ref FILE: PathBuf = build_path("key-pairs.json");
+}
+
+/// The possible file variants to be handled when deserializing FILE.
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+enum KeyStorageFile {
+    /// The genesis, unversioned file variant.
+    Unversioned(HashMap<String, KeyPairData>),
+
+    /// A versioned file variant, to which we have moved to
+    /// in order to leverage backwards-compatibility.
+    Versioned(VersionedFile),
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "version")]
+enum VersionedFile {
+    #[serde(rename = "1")]
+    V1 {
+        key_pairs: HashMap<String, KeyPairData>,
+    },
+}
 
 /// The data that is stored in the filesystem relative
 /// to a key pair. The name of the key pair is used as
@@ -65,11 +92,10 @@ pub enum Error {
 }
 
 fn io_error_message(action: &str) -> String {
-    let path = build_path(FILE);
     format!(
         "Failed to {} the key-pairs file: '{}'",
         action,
-        path.display()
+        FILE.display()
     )
 }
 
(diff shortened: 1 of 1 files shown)


cli/src/key_pair_storage.rs
```diff
@@ -185,30 +180,43 @@ fn dir() -> PathBuf {
...
+/// Initialize the given directory to be used.
+/// Fails with
+///   * [Error::CannotCreateDirectory] if the directory
+///     does not exist and fails to be created.
+///   * [Error::CannotReadDirectory] if the directory
+///     does exist but cannot be read.
+fn init_dir(dir: PathBuf) -> Result<PathBuf, Error> {
+    std::fs::create_dir_all(&dir).map_err(|err| Error::CannotCreateDirectory(err, dir.clone()))?;
+    File::open(&dir).map_err(|err| Error::CannotReadDirectory(err, dir.clone()))?;
+    Ok(dir)
```
CodeSandwich, line 204: Why are we returning the path?

cli/src/key_pair_storage.rs
```diff
@@ -185,30 +180,43 @@ fn dir() -> PathBuf {
...
 }
 
-fn parse_file() -> Result<KeyStorageFile, Error> {
-    let path_buf = get_or_create_path()?;
-    let file = File::open(path_buf.as_path()).map_err(ReadingError::IO)?;
+/// Initialize the storage on disk to be used correctly.
+///   * Create the directory structure and check permissions
+///   * Create and initialize the [FILE] where the key pairs will be stored.
+fn init() -> Result<(), Error> {
+    let path_buf = build_path(FILE);
```
CodeSandwich, line 187: I suspect that creating a function building a *parent directory* path would improve the code structure

cli/src/key_pair_storage.rs
NunoAlexandre: Legacy, dropped it, thanks!

cli/src/key_pair_storage.rs
NunoAlexandre: Not sure if I get it. Do you mean something like this?

``` rust
fn init() -> Result<(), Error> {
    let dir = dir();
    init_dir(dir)?;
    init_file(dir.join(FILE))?;
    Ok(())
}
```

CodeSandwich: Yes. The `dir.join(FILE)` could be then done inside `init_file` to encapsulate knowledge even further.

NunoAlexandre: That's the `build_path(filename)` function :)

The thing is that we need to get the directory path to pass it to `init_dir`. If we compute it aside from the file path, there is space for two sources of truth as for the dir behind the file path. As in, dir() and build_path().parent().unwrap() _can_ be different.

I am pushing a fixup improving this bit a little and making the file path lazy_static. Let me know what you think :+1:

CodeSandwich: It's good, it was a nitpick from my side anyway :+1:

PR #450, opened 2020-05-13
Add a member to an org

MeBrei:
Add the functionality to add a member to an existing org. 

Todos:
- [x] add more runtime tests
  - [x] bad author scenario, where the said author would pay the fee
  - [x] no org scenario, where the said author would pay the fee
  - [x] no user associated scenario, where the said author would pay the fee
- [x] add end-to-end tests
- [x] add cli functionality
- [x] error out if an existing member is added again

In a follow up, the functionality to remove a member should be added

### DIFF
--- a/client/src/message.rs
+++ b/client/src/message.rs
@@ -59,6 +59,25 @@ impl Message for message::RegisterProject {
     }
 }
 
+impl Message for message::RegisterMember {
+    type Result = Result<(), TransactionError>;
+
+    fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError> {
+        let dispatch_result = get_dispatch_result(&events)?;
+        match dispatch_result {
+            Ok(()) => find_event(&events, "MemberRegistered", |event| match event {
+                Event::registry(registry::Event::MemberRegistered(_, _)) => Some(Ok(())),
+                _ => None,
+            }),
+            Err(dispatch_error) => Ok(Err(dispatch_error)),
+        }
+    }
+
+    fn into_runtime_call(self) -> RuntimeCall {
+        registry::Call::register_member(self).into()
+    }
+}
+
 impl Message for message::RegisterOrg {
     type Result = Result<(), TransactionError>;
 
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -70,6 +70,9 @@ pub enum RegistryError {
     #[cfg_attr(feature = "std", error("a user with the same ID already exists."))]
     DuplicateUserId,
 
+    #[cfg_attr(feature = "std", error("the user is already a member of the org"))]
+    AlreadyAMember,
+
     #[cfg_attr(feature = "std", error("the provided fee is insufficient"))]
     InsufficientFee,
 
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -92,6 +92,31 @@ pub struct UnregisterUser {
     pub user_id: Id,
 }
 
+/// Register a new member for an org on the Registry with the given user ID.
+///
+/// # State changes
+///
+/// If successful, the `user_id` is added to [crate::state::Org::members] of `org_id` .
+///
+/// # State-dependent validations
+///
+/// The identified org must exit.
+///
+/// The user associated with the author must be a member of the identified org.
+///
+/// A user associated with the `user_id` must exist.
+///
+/// The `user_id` must not already be a member of the org.
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct RegisterMember {
+    // The member to register, unique in the org.
+    pub user_id: Id,
+
+    /// The org in which to register the member.
+    pub org_id: Id,
+}
+
 /// Register a project on the Radicle Registry with the given ID.
 ///
 /// # State changes
--- a/core/src/state.rs
+++ b/core/src/state.rs
@@ -154,6 +154,16 @@ impl Org {
         }
         self
     }
+
+    /// Add the given user to the list of [Org::members].
+    /// Return a new Org with the new member included or the
+    /// same org if the org already contains that member.
+    pub fn add_member(mut self, user_id: Id) -> Org {
+        if !self.members.contains(&user_id) {
+            self.members.push(user_id);
+        }
+        self
+    }
 }
 
 /// # Storage
--- a/runtime-tests/tests/member_registration.rs
+++ b/runtime-tests/tests/member_registration.rs
@@ -1,0 +1,245 @@
+// Radicle Registry
+// Copyright (C) 2019 Monadic GmbH <radicle@monadic.xyz>
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License version 3 as
+// published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+/// Runtime tests implemented with [MemoryClient].
+///
+/// High-level runtime tests that only use [MemoryClient] and treat the runtime as a black box.
+///
+/// The tests in this module concern project registration.
+use radicle_registry_client::*;
+use radicle_registry_test_utils::*;
+
+#[async_std::test]
+async fn register_member() {
+    let client = Client::new_emulator();
+    let (author, author_id) = key_pair_with_associated_user(&client).await;
+    let (_, member_user_id) = key_pair_with_associated_user(&client).await;
+
+    let register_org = random_register_org_message();
+    submit_ok(&client, &author, register_org.clone()).await;
+
+    // The org needs funds to submit transactions.
+    let org = client
+        .get_org(register_org.org_id.clone())
+        .await
+        .unwrap()
+        .unwrap();
+    let initial_balance = 1000;
+    transfer(&client, &author, org.account_id, initial_balance).await;
+
+    let random_fee = random_balance();
+    let message = message::RegisterMember {
+        org_id: register_org.org_id.clone(),
+        user_id: member_user_id,
+    };
+    let tx_applied = submit_ok_with_fee(&client, &author, message.clone(), random_fee).await;
+
+    assert_eq!(
+        tx_applied.events[0],
+        RegistryEvent::MemberRegistered(message.clone().user_id, message.clone().org_id).into()
+    );
+
+    // Fetch the org again
+    let re_org = client
+        .get_org(message.clone().org_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(re_org.members.len(), 2);
+    assert!(
+        re_org.members.contains(&author_id),
+        "Org does not contain the founding member."
+    );
+    assert!(
+        re_org.members.contains(&message.clone().user_id),
+        "Org does not contain the added member."
+    );
+
+    assert_eq!(
+        client.free_balance(&re_org.account_id).await.unwrap(),
+        initial_balance - random_fee,
+        "The tx fee was not charged properly."
+    );
+}
+
+#[async_std::test]
+async fn register_member_with_inexistent_org() {
+    let client = Client::new_emulator();
+    let (author, user_id) = key_pair_with_associated_user(&client).await;
+    let initial_balance = client.free_balance(&author.public()).await.unwrap();
+
+    let random_fee = random_balance();
+    let inexistent_org_id = random_id();
+    let message = message::RegisterMember {
+        org_id: inexistent_org_id,
+        user_id,
+    };
+    let tx_applied = submit_ok_with_fee(&client, &author, message.clone(), random_fee).await;
+
+    assert_eq!(tx_applied.result, Err(RegistryError::InexistentOrg.into()));
+
+    // Check that the author payed for the transaction anyway.
+    assert_eq!(
+        client.free_balance(&author.public()).await.unwrap(),
+        initial_balance - random_fee,
+        "The tx fee was not charged properly."
+    );
+}
+
+#[async_std::test]
+async fn register_member_with_bad_actor() {
+    let client = Client::new_emulator();
+    let (good_actor, good_actor_id) = key_pair_with_associated_user(&client).await;
+    let (bad_actor, bad_actor_id) = key_pair_with_associated_user(&client).await;
+
+    // The good actor creates an org, of which becomes its single member.
+    let org_id = random_id();
+    let register_org = message::RegisterOrg {
+        org_id: org_id.clone(),
+    };
+    submit_ok(&client, &good_actor, register_org.clone()).await;
+
+    // The bad actor attempts to register themselves as a member within that org.
+    let initial_balance = client.free_balance(&bad_actor.public()).await.unwrap();
+    let register_member = message::RegisterMember {
+        org_id,
+        user_id: bad_actor_id,
+    };
+    let random_fee = random_balance();
+    let tx_applied =
+        submit_ok_with_fee(&client, &bad_actor, register_member.clone(), random_fee).await;
+
+    assert_eq!(
+        tx_applied.result,
+        Err(RegistryError::InsufficientSenderPermissions.into())
+    );
+
+    // Check that the bad actor payed for the transaction anyway.
+    assert_eq!(
+        client.free_balance(&bad_actor.public()).await.unwrap(),
+        initial_balance - random_fee,
+        "The tx fee was not charged properly."
+    );
+
+    // Re-fetch the org and check that the bad actor was not added as a member
+    let re_org = client
+        .get_org(register_member.clone().org_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(re_org.members, vec![good_actor_id]);
+}
+
+#[async_std::test]
+async fn register_duplicate_member() {
+    let client = Client::new_emulator();
+    let (author, author_id) = key_pair_with_associated_user(&client).await;
+
+    // Register the org.
+    let org_id = random_id();
+    let register_org = message::RegisterOrg {
+        org_id: org_id.clone(),
+    };
+    submit_ok(&client, &author, register_org.clone()).await;
+
+    // The org needs funds to submit transactions.
+    let org = client
+        .get_org(register_org.org_id.clone())
+        .await
+        .unwrap()
+        .unwrap();
+    let initial_balance = 1000;
+    transfer(&client, &author, org.account_id, initial_balance).await;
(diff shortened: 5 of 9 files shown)


client/src/message.rs
```diff
@@ -59,6 +59,25 @@ impl Message for message::RegisterProject {
     }
 }
 
+impl Message for message::RegisterMember {
```
NunoAlexandre, line 62: :heavy_check_mark:

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
     pub user_id: Id,
 }
 
+/// Register a member of an org on the Radicle Registry with the given ID.
```
NunoAlexandre, line 95:
```suggestion
/// Register a member of an org with the given ID.
```

It is unclear which entity the ID is referring to though, in both versions.

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
...
+///
+/// # State-dependent validations
+///
+/// The involved org must exit.
```
NunoAlexandre, line 99: Alexis uses "identified" which I quite like:

```suggestion
/// The identified org must exit.
```

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
...
+///
+/// A user associated with the author must exist.
+///
+/// The user associated with the author must be a member of the involved org.
```
NunoAlexandre, line 103:
```suggestion
/// The user associated with the author must be a member of the identified org.
```

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
...
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct RegisterMember {
+    // The member to register, unique in the org.
```
NunoAlexandre, line 107: Is the "unique in the org" necessary? For a moment I wondered if that was actually a state-dependent validation. Is it? If it isn't, then I would drop this bit here since Ids are already unique in nature. If it is, then we should have that under the state-dependent validations above.

core/src/state.rs
```diff
@@ -154,6 +154,16 @@ impl Org {
         }
         self
     }
+
+    /// Add the given user to the list of [Org::members].
```
NunoAlexandre, line 158: :heavy_check_mark:

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    assert_eq!(
+        tx_applied.events[0],
+        RegistryEvent::MemberRegistered(message.clone().user_id, message.clone().org_id).into()
+    );
+    assert_eq!(re_org.members.len(), 2);
+    assert!(
+        re_org.members.contains(&message.clone().user_id),
+        "Org does not contain the added member."
+    );
```
NunoAlexandre, line 61: This could be simplified by `assert_eq` that the re_org.members is precisely an expected `vec![user_id_a, user_id_b]`. It also gives more guarantees to its correctness, as it could be adding the new members but swapping the one that was present before.

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,68 @@
...
+    let re_org = client
+        .get_org(message.clone().org_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(
+        tx_applied.events[0],
+        RegistryEvent::MemberRegistered(message.clone().user_id, message.clone().org_id).into()
+    );
```
NunoAlexandre, line 56: I would swap these two blocks: first, assert the expected event and then verify the correctness of the state.

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,68 @@
...
+/// Runtime tests implemented with [MemoryClient].
+///
+/// High-level runtime tests that only use [MemoryClient] and treat the runtime as a black box.
+///
+/// The tests in this module concern project registration.
+use radicle_registry_client::*;
+use radicle_registry_test_utils::*;
+
+#[async_std::test]
+async fn register_member() {
```
NunoAlexandre, line 25: Next to this test, we would also need to test the

* bad author scenario, where the said author would pay the fee
* no org scenario,  where the said author would pay the fee
* no user associated scenario,  where the said author would pay the fee

runtime/src/registry.rs
```diff
@@ -178,6 +178,21 @@ decl_module! {
...
+        pub fn register_member(origin, message: message::RegisterMember) -> DispatchResult {
+            let sender = ensure_signed(origin)?;
+
+            let org = store::Orgs::get(message.org_id.clone()).ok_or(RegistryError::InexistentOrg)?;
+            if !org_has_member_with_account(&org, sender) {
+                return Err(RegistryError::InsufficientSenderPermissions.into());
+            }
+
+            store::Orgs::insert(message.org_id.clone(), org.add_member(message.user_id.clone()));
+
```
NunoAlexandre, line 191: maybe drop this line?

test-utils/src/lib.rs
```diff
@@ -130,6 +130,11 @@ pub fn random_register_project_message(
     }
 }
 
+/// Create a [message::RegisterMember] with random parameters to register a member with.
+pub fn random_register_member_message(org_id: Id, user_id: Id) -> message::RegisterMember {
```
NunoAlexandre, line 134: It isn't random though

runtime/src/registry.rs
xla:
```suggestion
```

test-utils/src/lib.rs
```diff
@@ -130,6 +130,11 @@ pub fn random_register_project_message(
...
+    message::RegisterMember { org_id, user_id }
```
xla, line 135: You can use `random_id` if you want that, but it won't be of much use in most cases as the entities need to exist. I'd drop this function alltogether and just contruct the message where needed.

core/src/state.rs
```diff
@@ -154,6 +154,16 @@ impl Org {
...
+    /// Return a new Org with the new member included or the
+    /// same org if the org already contains that member.
+    pub fn add_member(mut self, user_id: Id) -> Org {
```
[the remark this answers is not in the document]
NunoAlexandre: I disagree. The term `handle` is not present in the spec and it creates unclearity as to what it refers to. Everywhere else, starting in the spec, we name it `user_id` and I see no value in changing it to `handle`.

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
...
+    pub user_id: Id,
```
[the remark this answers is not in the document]
NunoAlexandre: See https://github.com/radicle-dev/radicle-registry/pull/450#discussion_r424401035.

More so, there is no confusion right now since the field needs to be explicitly called out when constructing `RegisterMember`:

``` rust
let x = RegisterMember {
  user_id: foo,
  org_id: bar
};
```

changing it to `handle` doesn't make it any better:

``` rust
let x = RegisterMember {
  handle: foo,
  org_id: bar
};
```

xla:
> One remark: The new message is called `RegisterMember`, which I find confusing. It's not really registering a user, it's adding an existing user to the members of an org. I would, therefore, prefer to call it "AddOrgMember", which will also be less likely to collide with future messages. Note how the fn you added to `Org` is named `add_member` already.
> 
> Edit: @cloudhead I see that the spec calls it comes from the spec, `3.3 register-member`.

Semantically registering a member is correct, as it is with org, user and (org/user, projectname) a unique entry in the registry. We don't need to qualify with Org here as there is no other concept of member and the `add_member` fn is for adding to the actual list, not about the external API.

core/src/message.rs
xla: This is why I advised to name it handle to make it clear that it is a user.

core/src/message.rs
xla: Isn't this to express that the member can only be registered once per org? Making the tuple (org_id, user_handle) unique?

core/src/state.rs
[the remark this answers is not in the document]
MeBrei: I don't see `handle` being used for this in registry. `user_id` seems to be the common variable.

runtime-tests/tests/member_registration.rs
MeBrei: 👍 Added them to the TODO list in the description

test-utils/src/lib.rs
MeBrei: Removed in 10d1f17

test-utils/src/lib.rs
MeBrei: Removed in 10d1f17

core/src/message.rs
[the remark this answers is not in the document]
MeBrei: If the author is a member of the org is checked in https://github.com/radicle-dev/radicle-registry/pull/450/files#diff-bb77f0662a497780033f0237bb300323R186

I'll add the other check to the todo

runtime-tests/tests/member_registration.rs
MeBrei: Can we assume that `members` will always be ordered in the same way?

NunoAlexandre: That is a good question. I think so but not sure. As long as we check the exact two ids it contains I'm good with it :+1:

NunoAlexandre:
> > One remark: The new message is called `RegisterMember`, which I find confusing. It's not really registering a user, it's adding an existing user to the members of an org. I would, therefore, prefer to call it "AddOrgMember", which will also be less likely to collide with future messages. Note how the fn you added to `Org` is named `add_member` already.
> > Edit: @cloudhead I see that the spec calls it comes from the spec, `3.3 register-member`.
> 
> Semantically registering a member is correct, as it is with org, user and (org/user, projectname) a unique entry in the registry. We don't need to qualify with Org here as there is no other concept of member and the `add_member` fn is for adding to the actual list, not about the external API.

I don't disagree with you, but present "register member" to an org vs "add member" to an org to any number of users and I am pretty sure which one will be simpler to understand.

xla:
> I don't disagree with you, but present "register member" to an org vs "add member" to an org to any number of users and I am pretty sure which one will be simpler to understand.

In this context we are talking about the semantics of the chain and not what is presented to the user in the UI.

NunoAlexandre:
> > I don't disagree with you, but present "register member" to an org vs "add member" to an org to any number of users and I am pretty sure which one will be simpler to understand.
> 
> In this context we are talking about the semantics of the chain and not what is presented to the user in the UI.

I know :) But why have two terminologies when "add member" works for both audiences? While registering an org / project / user actually registers new entities into the chain, "register a member" simply links two entities that are already registered.

core/src/state.rs
```diff
@@ -154,6 +154,16 @@ impl Org {
         }
         self
     }
+
+    /// Add the given user to the list of [Org::members].
+    /// Return a new Org with the new member included or the
+    /// same org if the org already contains that member.
```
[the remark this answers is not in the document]
MeBrei: added a todo

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
...
+/// Register a member of an org on the Radicle Registry with the given ID.
+///
+/// # State-dependent validations
+///
+/// The involved org must exit.
+///
+/// A user associated with the author must exist.
+///
+/// The user associated with the author must be a member of the involved org.
+///
```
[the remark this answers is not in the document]
MeBrei: The additional `///` is in style with the rest of the documentation in the file

runtime-tests/tests/member_registration.rs
MeBrei: Swapped in 88bdd79

runtime-tests/tests/member_registration.rs
MeBrei: added a check for the founding member in 05a63f2

xla:
Part of radicle-dev/radicle-upstream#349

core/src/message.rs
MeBrei: it's referring to the member being unique within the org, ie a member can't be added twice

core/src/message.rs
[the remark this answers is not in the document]
MeBrei: I also would keep with `user_id` here, then it is immediately clear that it is referring to the user and not the org.

MeBrei:
> > > I don't disagree with you, but present "register member" to an org vs "add member" to an org to any number of users and I am pretty sure which one will be simpler to understand.
> > 
> > 
> > In this context we are talking about the semantics of the chain and not what is presented to the user in the UI.
> 
> I know :) But why have two terminologies when "add member" works for both audiences? While registering an org / project / user actually registers new entities into the chain, "register a member" simply links two entities that are already registered.

How can we come to a decision on this discussion? @NunoAlexandre @xla @cloudhead

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,239 @@
...
+    let (author, user_id) = key_pair_with_associated_user(&client).await;
+
+    let inexistent_org_id = random_id();
+    let message = message::RegisterMember {
+        org_id: inexistent_org_id,
+        user_id,
+    };
+    let tx_applied = submit_ok(&client, &author, message.clone()).await;
+
+    assert_eq!(tx_applied.result, Err(RegistryError::InexistentOrg.into()));
```
NunoAlexandre, line 90: We need to check that `author` has paid the tx fee

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,239 @@
...
+        .get_org(register_member.clone().org_id)
+        .await
+        .unwrap()
+        .unwrap();
+
+    assert_eq!(re_org.members, vec![good_actor_id]);
+}
+
+#[async_std::test]
+async fn register_duplicate_member() {
```
NunoAlexandre, line 139: nice :ok_hand:

runtime-tests/tests/member_registration.rs
```diff
@@ -0,0 +1,239 @@
...
+        initial_balance - random_fee,
+        "The tx fee was not charged properly."
+    );
+
+    // Check that the author was not added again
+    assert_eq!(re_org.members, vec![author_id]);
+}
+
+#[async_std::test]
+async fn register_nonexistent_member() {
```
NunoAlexandre, line 192: I got confused with the name for a sec, maybe:

```suggestion
async fn register_nonexistent_user() {
```

runtime/src/registry.rs
```diff
@@ -178,6 +178,28 @@ decl_module! {
...
+        #[weight = SimpleDispatchInfo::InsecureFreeNormal]
+        pub fn register_member(origin, message: message::RegisterMember) -> DispatchResult {
+            let sender = ensure_signed(origin)?;
+
+            let org = store::Orgs::get(message.org_id.clone()).ok_or(RegistryError::InexistentOrg)?;
+            if !org_has_member_with_account(&org, sender) {
+                return Err(RegistryError::InsufficientSenderPermissions.into());
+            }
+
+            if None == store::Users::get(message.user_id.clone()) {
```
NunoAlexandre, line 190:
```suggestion
            if store::Users::get(message.user_id.clone()).is_none() {
```

runtime/src/registry.rs
```diff
@@ -178,6 +178,28 @@ decl_module! {
...
+                return Err(RegistryError::InexistentUser.into());
+            }
+
+            if org.members.contains(&message.user_id) {
+                return Err(RegistryError::DuplicateMember.into());
+            }
+
+            store::Orgs::insert(message.org_id.clone(), org.add_member(message.user_id.clone()));
```
NunoAlexandre, line 198: We mostly try to separate such lines, like so:

```suggestion
            let org_with_member = org.add_member(message.user_id.clone());
            store::Orgs::insert(message.org_id.clone(), org_with_member);
```

runtime/src/registry.rs
```diff
@@ -415,6 +437,7 @@ decl_event!(
     pub enum Event {
         CheckpointCreated(CheckpointId),
         CheckpointSet(ProjectName, Id, CheckpointId),
+        MemberRegistered(Id, Id),
```
NunoAlexandre, line 440: Forgot this one! Since the types are ambiguous, I'd rather change this to:

```suggestion
        MemberRegistered { user_id: Id, org_id: Id },
```

MeBrei: seems like the macro does not support variants with named fields :/

geigerzaehler:
@MeBrei: Could you ping me once this is ready? I’d also like to have a brief look at this

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
     pub user_id: Id,
 }
 
+/// Register a new member for an org on the Registry with the given user ID.
+///
+/// # State-dependent validations
+///
+/// The identified org must exit.
+///
+/// A user associated with the author must exist.
```
geigerzaehler, line 101: Two validations are missing
* The user `user_id` must exist
* `user_id` must not already be a member of the org.

core/src/error.rs
```diff
@@ -70,6 +70,12 @@ pub enum RegistryError {
     #[cfg_attr(feature = "std", error("a user with the same ID already exists."))]
     DuplicateUserId,
 
+    #[cfg_attr(
+        feature = "std",
+        error("a member with the same ID already exists within the org.")
```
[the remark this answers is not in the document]
NunoAlexandre: to be clear, this was just a suggestion, since `DuplicateMember` could reflect that the vec Org::members has duplicates, when the error is to say "you are trying to add a user to the members of an org but they are already  a member".

That said, if we go ahead with this change, I'd suggest extending the error message to this, (sorry for the double comment):

error("the user is already a member of the org")

core/src/message.rs
```diff
@@ -92,6 +92,25 @@ pub struct UnregisterUser {
     pub user_id: Id,
 }
 
+/// Register a new member for an org on the Registry with the given user ID.
+///
+/// # State-dependent validations
```
geigerzaehler, line 97: Could you also [list the state changes](https://github.com/radicle-dev/radicle-registry/blob/master/core/README.md#transaction-messages)?
```
/// # State changes
/// If successful, `user_id` is added to [Org::members] of `org_id`.
```

core/src/message.rs
NunoAlexandre: > Two validations are missing
> 
> * The user `user_id` must exist

It says already:

> A user associated with the author must exist.

we could change that to:

> A user associated with the author must exist and be identified by `user_id`

geigerzaehler: Isn’t the user associated with the author different from `user_id`?

MeBrei: yes, will change it.

core/src/message.rs
MeBrei: added in 8322c12

core/src/message.rs
MeBrei: changed in 8322c12

PR #452, opened 2020-05-14
Add CLI command to submit new wasm runtime

### LINKED ISSUE #443: runtime: build tool to submit WASM update transaction
Part of #434

To [update the runtime](#434) running on existing nodes of the ffnet, we need to broadcast a transaction containing the updated WASM runtime that the nodes will receive and pick up on.

To achieve that, we will need to:

 - Add a function to the client submitting the right message with the new WASM blob
 - Add equivalent command to the CLI
- Do version check, not allowing to submit a WASM version older than the one present in the nodes
- CI: Verify that new WASM lives on-chain after submission

NunoAlexandre:
Closes #443 

### `TBD`

- [x] Improve error propagation, reporting the case where the submission fails because the submitted wasm `spec_version` is not greater than the one the runtime of the node we are connected to is running.

~~- [ ] https://github.com/radicle-dev/radicle-registry/pull/452#discussion_r427319736 "There is additional information that we can extract here: We can determine whether the author matches the sudo key. I think this would be valuable to have as an additional error case."~~

~~- [ ] https://github.com/radicle-dev/radicle-registry/pull/452#pullrequestreview-414456802 Add client-side validations~~

**Note**: We can't do this at the moment since we would need to either call `frame_system::Module<T>::can_set_code` or duplicate its logic. I tried the second approach but we can't do either because both call `sp_io::misc::runtime_version`, which fails with `called outside of an Externalities-provided environment.`

### Notes

* Initially I didn't have this transaction as a `message` but ended up switching to that approach to reduce code duplication and use the same infrastructure already in place for all other messages. Some parts are necessarily custom but this approach still feels better this way

* Regarding the docs for the new `UpdateRuntime` message, I kept the `State changes` and `State-dependent validations` blocks since they are kind of applicable, although not quite the same as the other messages. Let me know how you see it.

* Get a Result from the events obtained from this transactions is not as clean as I'd wish.
  I set it so that when we get a `RawEvent::CodeUpdated` event, it means success. Otherwise, it fails with the new error `FailedRuntimeWasmUpdate`, since when the update fails we don't get any useful dispatch error nor a useful event.

* See https://github.com/radicle-dev/radicle-registry/pull/452#issuecomment-630034545 to learn how to run this locally successfully.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -162,7 +162,11 @@ runtime.
 ```
 
 For semantic updates to take effect on an existing chain they need to be
-deployed to the chain. This process has not been established yet.
+deployed to the chain.
+
+```
+radicle-registry-cli update-runtime ./runtime/latest.wasm --author <sudo_key> --node-host <node_host>
+```
 
 Changes to the chain state must be backwards-compatibility, requiring particular
 attention. A policy and process for such updates have not been established yet.
--- a/cli/src/command/other.rs
+++ b/cli/src/command/other.rs
@@ -23,6 +23,13 @@ use super::*;
 pub enum Command {
     /// Show the genesis hash the node uses
     GenesisHash(ShowGenesisHash),
+
+    /// Submit a transaction to update the on-chain runtime.
+    /// Requirements:
+    ///   * the tx author must be the chain's sudo key
+    ///   * the `spec_version` of the given wasm runtime must be greater than the chain runtime's.
+    ///   * the `spec_name` must match between the wasm runtime and the chain runtime.
+    UpdateRuntime(UpdateRuntime),
 }
 
 #[async_trait::async_trait]
@@ -30,6 +37,7 @@ impl CommandT for Command {
     async fn run(self) -> Result<(), CommandError> {
         match self {
             Command::GenesisHash(cmd) => cmd.run().await,
+            Command::UpdateRuntime(cmd) => cmd.run().await,
         }
     }
 }
@@ -46,6 +54,42 @@ impl CommandT for ShowGenesisHash {
         let client = self.network_options.client().await?;
         let genesis_hash = client.genesis_hash();
         println!("Genesis block hash: 0x{}", hex::encode(genesis_hash));
+        Ok(())
+    }
+}
+
+#[derive(StructOpt, Clone)]
+pub struct UpdateRuntime {
+    /// The path to the (wasm) runtime code to submit
+    path: std::path::PathBuf,
+
+    #[structopt(flatten)]
+    network_options: NetworkOptions,
+
+    #[structopt(flatten)]
+    tx_options: TxOptions,
+}
+
+#[async_trait::async_trait]
+impl CommandT for UpdateRuntime {
+    async fn run(self) -> Result<(), CommandError> {
+        let client = self.network_options.client().await?;
+        let new_runtime_code =
+            std::fs::read(self.path).expect("Invalid path or couldn't read the wasm file");
+
+        let update_runtime_fut = client
+            .sign_and_submit_message(
+                &self.tx_options.author,
+                message::UpdateRuntime {
+                    code: new_runtime_code,
+                },
+                self.tx_options.fee,
+            )
+            .await?;
+        announce_tx("Submitting the new on-chain runtime...");
+
+        update_runtime_fut.await?.result?;
+        println!("✓ The new on-chain runtime is now published.");
         Ok(())
     }
 }
--- a/client/src/message.rs
+++ b/client/src/message.rs
@@ -204,6 +204,31 @@ impl Message for message::TransferFromOrg {
     }
 }
 
+impl Message for message::UpdateRuntime {
+    type Result = Result<(), TransactionError>;
+
+    /// The only unequivocal sign we get that a wasm update was successful is the
+    /// `RawEvent::CodeUpdated` event. Anything else is considered a failed update.
+    fn result_from_events(events: Vec<Event>) -> Result<Self::Result, EventParseError> {
+        let error: TransactionError = RegistryError::FailedChainRuntimeUpdate.into();
+        find_event(&events, "System", |event| match event {
+            Event::system(system_event) => match system_event {
+                frame_system::RawEvent::CodeUpdated => Some(Ok(())),
+                _ => Some(Err(error)),
+            },
+            _ => Some(Err(error)),
+        })
+    }
+
+    fn into_runtime_call(self) -> RuntimeCall {
+        let set_code_call: RuntimeCall = frame_system::Call::set_code(self.code).into();
+        let sudo_call: pallet_sudo::Call<Runtime> =
+            pallet_sudo::Call::sudo(Box::new(set_code_call));
+
+        sudo_call.into()
+    }
+}
+
 /// Extract the dispatch result of an extrinsic from the extrinsic events.
 ///
 /// Looks for the [frame_system::Event] in the list of events and returns the inner result based on
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -99,6 +99,18 @@ pub enum RegistryError {
         error("the account is already associated with a user")
     )]
     UserAccountAssociated,
+
+    #[cfg_attr(
+        feature = "std",
+        error(
+            "failed to update the chain runtime. Ensure that:
+    * the author is the chain's sudo key
+    * the 'spec_name' match
+    * the wasm 'spec_version' is greater
+        "
+        )
+    )]
+    FailedChainRuntimeUpdate,
 }
 
 // The index with which the registry runtime module is declared
--- a/core/src/message.rs
+++ b/core/src/message.rs
@@ -19,6 +19,7 @@
 extern crate alloc;
 
 use crate::{AccountId, Balance, Bytes128, CheckpointId, Id, ProjectName, H256};
+use alloc::prelude::v1::Vec;
 use parity_scale_codec::{Decode, Encode};
 
 /// Registers an org on the Radicle Registry with the given ID.
@@ -199,4 +200,23 @@ pub struct TransferFromOrg {
 pub struct Transfer {
     pub recipient: AccountId,
     pub balance: Balance,
+}
+
+/// Attempts to update the on-chain runtime with the new given one.
+/// The `code` must be a valid WASM module and adhere to the substrate runtime API.
+///
+/// # State changes
+///
+/// If successful, the given `code` will be the new one herein.
+///
+/// # State-dependent validations
+///
+/// The tx author must be the chain's sudo key
+///
+/// The `spec_version` of the given runtime code needs to be greater than
+/// the `spec_version` of the chain runtime.
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct UpdateRuntime {
+    pub code: Vec<u8>,
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ Upcoming
 * Polkadot telemetry was removed
 
 ### Addition
+
+* cli: Add `update-runtime` command to update the on-chain runtime
 * cli: Mutually support local account names where only SS58 address were
   supported as params.
 * Add `user list` and `user show` CLI commands
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3670,6 +3670,7 @@ dependencies = [
  "jsonrpc-core-client",
  "lazy_static",
  "log 0.4.8",
+ "pallet-sudo",
  "parity-scale-codec",
  "radicle-registry-core",
  "radicle-registry-runtime",
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -28,6 +28,12 @@ thiserror = "1.0.14"
 tokio = "0.1"
 url = "1.7"
 
+
+[dependencies.pallet-sudo]
+git = "https://github.com/paritytech/substrate"
+rev = "3f134a1af2462482b2868adf73a050a8da0d3934"
+default_features = false
+
 [dependencies.frame-system]
 git = "https://github.com/paritytech/substrate"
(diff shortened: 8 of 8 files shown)


client/src/lib.rs
```diff
@@ -174,6 +174,37 @@ impl ClientT for Client {
...
+    }
+
+    async fn update_runtime(
+        &self,
+        author: &ed25519::Pair,
+        new_wasm_runtime: Vec<u8>,
+        fee: Balance
+    ) -> Result<Response<backend::TransactionApplied, Error>, Error> {
+        println!("Runtime update...");
+        let call: RuntimeCall = frame_system::Call::set_code(new_wasm_runtime).into();
```
geigerzaehler, line 192: The call needs to be wrapped in a `Sudo` call. Otherwise the runtime will rightfully reject it because it is not valid.

NunoAlexandre: Yes. I had already added `pallet_sudo` to the client Cargo file but wanted to try without it since I don't see a reason to need to wrap it in a Sudo call besides making the job simpler for substrate to auth the xt. Now wrapped it seems to work, running some tests to see how the whole thing behaves when submitting a new runtime with impl changes.

NunoAlexandre:
@geigerzaehler as a follow up of https://github.com/radicle-dev/radicle-registry/pull/452#discussion_r426418262, I have:

1. Run `./scripts/build-dev-node`
2. Purge local chain
3. Run freshly built node
4. Build a new WASM runtime (`./scripts/build-runtime-wasm`) with a change to the `register_org` handler in the runtime / registry.rs module, making it always `Err(RegistryError::InexistentOrg.into())` for purely debug purposes
5. Register an org against the running localhost node -> :heavy_check_mark: 
6. Submit new wasm generated in step 4. using the new `update-runtime` command
    It seems to have worked, as in, it didn't fail nor err.
7. Register another org -> It still worked :x: It should have failed with the Err introduced in step 4, submitted in step 6.

Checking the execution strategy flags, but wouldn't expect the issue to be there.

geigerzaehler:
The update is not successful. If you run the node with `RUST_LOG=runtime=debug` you can see that it complains that the new runtime version does not bump the spec version. (This is enforced [here](https://github.com/paritytech/substrate/blob/5a7600912b7fd50091990899e8ad93ad618f40c4/frame/system/src/lib.rs#L1242-L1244)).

NunoAlexandre:
Thanks, @geigerzaehler.

As you pointed out, bumping the spec version did the trick. There are some caveats worth pointing out, which consumed some of my time.

1. bump `spec_version` in runtime lib.rs
2. generate the new wasm runtime: `./scripts/build-runtime-wasm <name>
3. :exclamation: Stash the changes bumping the `spec_version`
     Without doing this, the requests now made by the CLI will be using a client whose runtime `spec_version` is greater than the one the node it connects to runs, so all txs will be invalid.
4. Submit the new wasm `radicle-registry-cli update-runtime <name> --node-host localhost --author alice --fee 200`
5.  :exclamation: Unstash the changes bumping the `spec_version`
   The reverse problem in step 3; Without this, the runtime version of the client used by the CLI is smaller than the runtime of the node it is connecting to, to all transactions fail. We can now align the versions and have the latest changes introduced in the new wasm runtime running.

geigerzaehler:
>     3\. Stash the changes bumping the `spec_version`
>         Without doing this, the requests now made by the CLI will be using a client whose runtime `spec_version` is greater than the one the node it connects to runs, so all txs will be invalid.

Could you elaborate on this? Why are the transactions submitted with a client that is compiled against the updated spec invalid? Shouldn’t the new client be able to submit txs even against older versions of the runtime provided that the transaction serialization did not change?

NunoAlexandre:
> > Stash the changes bumping the `spec_version`
> > Without doing this, the requests now made by the CLI will be using a client whose runtime `spec_version` is greater than the one the node it connects to runs, so all txs will be invalid.
> 
> Could you elaborate on this? Why are the transactions submitted with a client that is compiled against the updated spec invalid? Shouldn’t the new client be able to submit txs even against older versions of the runtime provided that the transaction serialization did not change?

Note that this is me explaining how to successfully update the wasm runtime running on a node now, it's now me setting the way it should be done.

That said, regarding your question, yes, I would expect that to be the case. I suspect that this happens because of the `CheckVersion` signed extension: https://substrate.dev/rustdocs/master/src/frame_system/lib.rs.html#1532. Still exploring this.

geigerzaehler:
> I suspect that this happens because of the `CheckVersion` signed extension: https://substrate.dev/rustdocs/master/src/frame_system/lib.rs.html#1532. Still exploring this.

Great find! I think you’re right. This is something we need to consider for client/cli updates.

NunoAlexandre:
> > I suspect that this happens because of the `CheckVersion` signed extension: https://substrate.dev/rustdocs/master/src/frame_system/lib.rs.html#1532. Still exploring this.
> 
> Great find! I think you’re right. This is something we need to consider for client/cli updates.

Indeed, it goes literally against the workflow we wanted to have in place. Will create a separate issue to discuss this.

NunoAlexandre:
@geigerzaehler this is now ready for review.

core/src/message.rs
```diff
@@ -200,3 +201,20 @@ pub struct Transfer {
...
+///
+///
+/// # State-dependent validations
+///
+/// The `spec_version` of the given `wasm` needs to be greater than
+/// the `spec_version` of the remote runtime.
+///
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct UpdateRuntime {
+    pub wasm: Vec<u8>,
```
geigerzaehler, line 219: I’d prefer the field name `code`. It aligns better with the `set_code` terminology.

core/src/message.rs
```diff
@@ -200,3 +201,20 @@ pub struct Transfer {
...
 }
+
+/// Attempts to update the WASM runtime with the new given one.
+
+/// # State changes
+///
+/// If successful, the given `wasm` runtime will be used herein.
+///
+///
+/// # State-dependent validations
```
geigerzaehler, line 212: Could you also add that `wasm` needs to be a valid WASM module and adhere to the substrate runtime API.

core/src/error.rs
```diff
@@ -99,6 +99,17 @@ pub enum RegistryError {
         error("the account is already associated with a user")
     )]
     UserAccountAssociated,
+
+    #[cfg_attr(
+        feature = "std",
+        error(
+            "failed to update the wasm runtime. Ensure that:
+    * the 'spec_name' match
+    * the new wasm 'spec_version' is greater
```
geigerzaehler, line 108: Can we avoid multiline error messages? They are likely to mess up our output. I’m ok with omitting ways to potentially fix this from the error message.

client/src/transaction.rs
```diff
@@ -77,7 +77,7 @@ pub struct TransactionExtra {
 /// validation is performed.
 ///
 /// `genesis_hash` is the genesis hash of the block chain this intrinsic is valid for.
-fn signed_extrinsic(
+pub fn signed_extrinsic(
```
geigerzaehler, line 80: Does not need to be public.

client/src/message.rs
```diff
@@ -212,7 +237,9 @@ impl Message for message::TransferFromOrg {
 /// Returns an outer [EventParseError] if no [frame_system::Event] was present in `events`.
 ///
 /// Because of an issue with substrate the `message` field of [TransactionError] will always be `None`
-fn get_dispatch_result(events: &[Event]) -> Result<Result<(), TransactionError>, EventParseError> {
+pub fn get_dispatch_result(
```
geigerzaehler, line 240: Does not need to be public.

DEVELOPING.md
```diff
@@ -197,3 +198,18 @@ DOCKER_IMAGE: gcr.io/opensourcecoin/radicle-registry/ci-base:e8c699d4827ed893d8d
 ```
 
 **What happens on the build agent:** when any commit with this change is pushed, the build agent will find the image under the configured tag. It will reuse it instead of rebuilding and save time.
+
+### Updating the Runtime
```
geigerzaehler, line 202: Could we integrate this under the “Runtime updates” section I created?

cli/src/command/other.rs
```diff
@@ -23,13 +23,21 @@ use super::*;
 pub enum Command {
     /// Show the genesis hash the node uses
     GenesisHash(ShowGenesisHash),
+
+    /// Update the WASM runtime running on the node it
+    /// will connect to. Requires the `spec_version` of
```
geigerzaehler, line 28: I think it is important to point out that this is a transaction that modifies the chain state.
```suggestion
    /// Submit a transaction to update the on-chain runtime.
    /// Requires the `spec_version` of
```

We should also point out that the author must match the chains sudo key.

client/src/interface.rs
```diff
@@ -23,7 +23,9 @@ pub use radicle_registry_core::*;
 
 pub use radicle_registry_runtime::{BlockNumber, Hash, Header};
 
-pub use radicle_registry_runtime::{registry::Event as RegistryEvent, Balance, Event};
+pub use radicle_registry_runtime::{
+    registry::Event as RegistryEvent, Balance, Event, UncheckedExtrinsic,
```
geigerzaehler, line 27: Additional export of `UncheckedExtrinsic` is not needed.

client/src/message.rs
```diff
@@ -204,6 +204,31 @@ impl Message for message::TransferFromOrg {
     }
 }
 
+impl Message for message::UpdateRuntime {
+    type Result = Result<(), TransactionError>;
+
+    /// The only unequivocal sign we get that a wasm update was successful is the
+    /// `RawEvent::CodeUpdated` event. Anything else is considered a failed update.
```
geigerzaehler, line 211: There is additional information that we can extract here: We can determine whether the author matches the sudo key. I think this would be valuable to have as an additional error case.

core/src/error.rs
NunoAlexandre: There isn't much that can go wrong really, and I crafted this to look good when outputted in the terminal. I prefer a rich-enough error message over a perfectly looking one that doesn't offer much.

client/src/message.rs
NunoAlexandre: Good point!

core/src/message.rs
NunoAlexandre: yes!

cli/src/command/other.rs
```diff
@@ -24,11 +24,12 @@ pub enum Command {
...
-    /// will connect to. Requires the `spec_version` of
-    /// the given wasm runtime to be greater than the
-    /// remote and the client runtime `spec_version`
-    /// version to match the remote.
+    /// Submit a transaction to update the on-chain runtime.
+    /// Requirements:
+    ///   * the tx author must be the chain's sudo key
+    ///   * the `spec_version` of the given wasm runtime must be greater
+    ///     than the remote.
+    ///   * the client runtime `spec_version` version to match the remote
```
geigerzaehler, line 32: The last point is true for every transaction. Does it make sense to call it out here?

NunoAlexandre: Yes, it does since this is not referring to the client runtime vs the node runtime but referring to the wasm runtime vs the node runtime.

NunoAlexandre: there were two typos here:

- it should be "wasm runtime" instead of "client runtime"
- "spec_name" instead of "spec_version"

Both fixed.

NunoAlexandre:
@geigerzaehler how do you feel about the steps to make a successful update? Should I add them to the respective section or do you suggest another place? Also, I recall you suggesting something related to automatically bump the `spec_version` based on its current value but I can't quite remember?

client/src/message.rs
NunoAlexandre: Just realized this is not dependant on the whole `can_set_point` part, working on it.

NunoAlexandre: Hm, submitting the tx using an author other than the sudo fails but with another error, so I suspect they check for the sudo user in an earlier stage. So, I would expect the error to be the `FailedChainRuntimeUpdate` variant but it is:

``` bash
Error: client error
  Caused by: Error from the underlying RPC connection
  Caused by: Server returned rpc error Server error: Invalid Transaction
```

Either we implement a client-side check to verify the author is the chain's sudo key, which probably runs into the same issue reported in the PR description relative to checking the wasm version, or we leave it as is. I'd prefer the second since only us, devs, will be running this command and we know that we should use the sudo key to run this command.

DEVELOPING.md
```diff
@@ -167,6 +167,25 @@ deployed to the chain. This process has not been established yet.
...
+./scripts/build-runtime-wasm <output_file>
+```
+
+2. Submit the new WASM runtime
+
+```
+radicle-registry-cli update-runtime <file> --author alice --node-host <node_host>
+```
+
+3. [Optional] Overwrite the `runtime/latest.wasm` with the WASM file successfully submitted.
```
geigerzaehler, line 185: This contradicts the “Runtime updates” section. There we say that you need to update this as part of the commit that does the change. My idea for this was that the new runtime update lands on master _before_ it is submitted so that it can be announced and reviewed.

NunoAlexandre:
> > Should I add them to the respective section or do you suggest another place?
> 
> It would make sense to integrate this into the “Semantic updates” section since submitting the transaction is only part of this larger process. That section already refers to the the submitting an update with a placeholder.
> 
> 
> Something that is still missing is documentation for the workaround for the version.

What I was referring in the first quote above was precisely the workaround. What were you referring to in that reply then? To the section "Updating the on-chain runtime" here added? If so, I don't think that should live there. I guess you say that because you don't see the point of updating the runtime with implementation changes? That was something we discussed and agreed on the value of continuously integrating such changes. 

In any case, this section is and should be independent of the type of changes included in the new version, so it should not live under one of the specific update types.

I will integrate the workaround here.

DEVELOPING.md
NunoAlexandre: You are right, will fix this.

Issue #434, opened 2020-05-08
Updates to the Runtime


NunoAlexandre:
Part of the [Registry Updates milestone](https://github.com/radicle-dev/radicle-registry/milestone/4)

#### Legend

🔨 - Issue
📝 - Something we will need to do or just keep in mind. Maybe pending becoming an issue.
📌 - A requirement or criteria

## Updates to the runtime

Independently on the type of update to the runtime, we need to:

- [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/437 Define a policy for developers on how to not break stuff when updating
- 📝 Publicly announced changelog
    Useful when some change might break previous user actions/patterns that worked previously.

- [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/438 Ensure that the dev node always runs the latest version

- 📝 Test run against the devnet, that should be always running at least the latest version of the ffnet
- [ ] 🔨 https://github.com/radicle-dev/radicle-registry/issues/444 Define release pipeline

### 1. Implementation changes

- [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/439 Need for criteria to verify that they are really just implementation changes
- 📌 The CLI must remain compatible
- Testing
    - [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/440  End to end tests need to run against both native and the wasm runtimes
- [x]  🔨https://github.com/radicle-dev/radicle-registry/issues/441  new miners and full nodes of the ffnet should run the latest native runtime version
    - 📌 substrate only runs native code if the execution strategy is set properly and if the version of the native code is compatible with the version of the on-chain runtime
- [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/442 existing miners and full nodes on the ffnet should run the latest wasm version
- [x]  🔨 https://github.com/radicle-dev/radicle-registry/issues/443 Build tool to submit WASM update transaction
   
### 2. Logic changes

- 📌 The CLI must remain compatible

### 3. State changes

- [ ]  🔨 https://github.com/radicle-dev/radicle-registry/issues/445 Devise way to make state changes backwards-compatible

geigerzaehler:
We have tickets for all the tasks so I’m closing this.

PR #454, opened 2020-05-15
Add runtime update policy

geigerzaehler:
We add a policy and describe the process for runtime updates in the developer handbook.

With this policy the repo now includes to wasm runtime blobs. `./runtime/latest.wasm` is used by the `dev*` chains.

We also change the execution strategy for all nodes to run both the wasm and the native code if possible.

See #434.


DEVELOPING.md
```diff
@@ -129,27 +132,39 @@ To update the revision of substrate run
...
+
+```
+./scripts/build-runtime-wasm ./runtime/latest.wasm
+```
+
+For specification changes to take effect on an existing chain they need to be
+deployed to the chain. This process has not been established yet.
+
+Special care must be taken that updated runtimes can handle state written be
+previous versions. A policy and process for this has not been established yet.
```
CodeSandwich, line 167:
```suggestion
Special care must be taken that updated runtimes can handle state written by
the previous versions. A policy and process for updates, which break that
compatibility have not been established yet.
```

DEVELOPING.md
```diff
@@ -129,27 +132,39 @@ To update the revision of substrate run
...
+
+### Specification changes
+
+Specification changes must increment the `spec_version` field.
+
+In a commit with a specification change you must also update the lastest Wasm
+runtime.
+
+```
+./scripts/build-runtime-wasm ./runtime/latest.wasm
```
CodeSandwich, line 160: Don't we want to keep the older runtimes? It could be useful for debugging when things start crashing in the real world with dozens of versions coexisting.

node/src/chain_spec.rs
```diff
@@ -30,7 +30,8 @@ use std::convert::TryFrom;
 /// Specialized `ChainSpec`. This is a specialization of the general Substrate ChainSpec type.
 pub type ChainSpec = GenericChainSpec<GenesisConfig>;
 
-const WASM_BINARY: &[u8] = include_bytes!("../../runtime/genesis_runtime.wasm");
+const FFNET_GENESIS_RUNTIME_WASM: &[u8] = include_bytes!("../../runtime/ffnet_genesis.wasm");
```
CodeSandwich, line 33: The binary is inflating, but I guess that it's inevitable if we want to keep it universal across networks.

DEVELOPING.md
```diff
@@ -129,27 +132,39 @@ To update the revision of substrate run
...
-To build a WASM runtime binary meant for packing into a transaction and broadcasting on the network run
-~~~
-./scripts/build-runtime-wasm <output_file>
-~~~
+Commits with implementation updates must increment the `impl_version` field of
+`VERSION`. They may not recompile `./runtime/latest.wasm`.
+
+### Specification changes
+
+Specification changes must increment the `spec_version` field.
```
CodeSandwich, line 154: Should it reset the `impl_version` to `0`?

DEVELOPING.md
NunoAlexandre: That is a good point, although I don't think we could submit an older version once we bump the runtime version at each release, or can we? In any case, git always tracks each version and we are not rewriting to master so that should be given for free.

CodeSandwich: That's correct :+1:

DEVELOPING.md
```diff
@@ -129,27 +132,39 @@ To update the revision of substrate run
...
 
-To build a genesis WASM runtime binary run
-~~~
-./scripts/build-genesis-runtime-wasm
-~~~
-This will create or update the `runtime/genesis_runtime.wasm` file.
-It will be then used as the genesis WASM runtime in all consecutive compilations of the node.
+Updates to the runtime are tracked by the `VERSION` exported from the `runtime`
+crate. The updates fall into two categories: Implementation updates and
+specification updates.
```
NunoAlexandre, line 142: We have been calling the latter semantic changes. can we settle on one terminology?

DEVELOPING.md
```diff
@@ -129,27 +132,39 @@ To update the revision of substrate run
...
-### Building a runtime for an update
+Implementation updates only change the implementation of the runtime but do not
+affect the semantics.
 
-To build a WASM runtime binary meant for packing into a transaction and broadcasting on the network run
-~~~
-./scripts/build-runtime-wasm <output_file>
-~~~
+Commits with implementation updates must increment the `impl_version` field of
+`VERSION`. They may not recompile `./runtime/latest.wasm`.
```
NunoAlexandre, line 150: > They may not recompile `./runtime/latest.wasm`.

I think your communication goal with this bit is to say that such changes are not required to be submitted/released?

DEVELOPING.md
NunoAlexandre:
```suggestion
Changes to the chain state must be backwards-compatibility, requiring particular attention.
A policy and process for such updates have not been established yet.
```

README.md
```diff
@@ -114,19 +115,28 @@ the [`RUST_LOG` environment variable][rust-log-docs].
 
 [rust-log-docs]: https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging
 
-### Dev
+Chains
+------
```
NunoAlexandre, line 119:
```suggestion
## Chains
```

@CodeSandwich also pointed out his preference to use this syntax instead of the underlying one.

README.md
```diff
@@ -114,19 +115,28 @@ the [`RUST_LOG` environment variable][rust-log-docs].
...
+
+The chain a node follows is set by the `--chain` flag. The following chains are
+supported.
```
NunoAlexandre, line 122: I would personally prefer these to be an enumeration instead of h3 headings, like so:

* `ffnet`
  The `ffnet` is our first public network for friends & family to participate.

* `dev`
  The `dev` chain is intended for local development. The node runs an isolated
  network with a dummy proof-of-work. The `dev` chain uses `./runtime/latest.wasm`
  as the genesis runtime.

* `devnet`
  We host a devnet that you can connect to. To join you need to use the most
  recent pre-built binary (see “Getting the node”).

  (...)

README.md
```diff
@@ -114,19 +115,28 @@ the [`RUST_LOG` environment variable][rust-log-docs].
...
-### Devnet
+### `devnet`
 
 We host a devnet that you can connect to. To join you need to use the most
 recent pre-built binary (see “Getting the node”).
 
-~~~
-radicle-registry-node --chain devnet
-~~~
+Like the `dev` chain we use `./runtime/latest.wasm` as the genesis runtime.
```
NunoAlexandre, line 139:
```suggestion
Like the `dev` chain, we use `./runtime/latest.wasm` as the genesis runtime.
```

runtime/src/lib.rs
```diff
@@ -81,10 +81,11 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<AccountId, Call, Signa
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("radicle-registry"),
     impl_name: create_runtime_str!("radicle-registry"),
-    authoring_version: 3,
     spec_version: 4,
     impl_version: 4,
     apis: runtime_api::VERSIONS,
+    // Ignored by us. Only `spec_version` is relevant.
```
NunoAlexandre, line 87: This left me confused. When is it the case that "Only `spec_version` is relevant"? On the docs above it states:

> Commits with implementation updates must increment the `impl_version` field of
`VERSION`. They may not recompile `./runtime/latest.wasm`.

node/src/chain_spec.rs
geigerzaehler: Yes, there is no way around this. We’ll have to keep the genesis runtime forever and also need the latest.

DEVELOPING.md
geigerzaehler: There’s no need for this. But it probably is confusing keeping the same version for the implementation. I’ll change it and require a reset to zero.

DEVELOPING.md
geigerzaehler: That is only secondary. The primary goal is to test the native version against the original wasm runtime.

README.md
geigerzaehler: I’d rather not mix the styles or update them all at once in this PR. I can follow up with a separate PR that changes the style.

README.md
geigerzaehler: I’d like to keep it as h3 headings. Having paragraphs (like for `devnet`) in list items is tricky. It will become more complicated if we have code examples etc. in there

DEVELOPING.md
geigerzaehler: Fixed in https://github.com/radicle-dev/radicle-registry/pull/454/commits/3e4c111a9e1a8d6fc40ca34aa58cded75f3be1d4

DEVELOPING.md
geigerzaehler: https://github.com/radicle-dev/radicle-registry/pull/454/commits/a40649eabbe7548a4f707a47d9dcbe8bf345a431

runtime/src/lib.rs
geigerzaehler: Clarified in https://github.com/radicle-dev/radicle-registry/pull/454/commits/6b2a763a1ec18ce65a084debb09198c2de88a038

README.md
NunoAlexandre: You'd just need to ident, I do that very often. I am fine with keeping it as is of course, it's not important.

PR #456, opened 2020-05-18
Improve CI build times with incremental builds

### LINKED ISSUE #378: Improve `cargo doc` build times.
At the moment CI spends substantial time on verifying the documentation with `cargo doc`. In [this build](https://buildkite.com/monadic/radicle-registry/builds/2051#f6b5cf0e-b56d-4066-8e6a-6ec56e81548b) `cargo doc` was responsible for a third of the build time and was slower than `./scripts/build-release`. This is representative of builds with good cache usages which is our most common case.

We should investigate the cause of this and suggest improvements.

geigerzaehler:
We improve build times on CI by re-enabling incremental builds. This allows the different builds in the `./ci/run` script to better share artifacts. We remove the incremental artifacts at the end of the script because they cannot be used by subsequent builds from a fresh repo checkout.

We also change the build of docs to use the release profile so that it can reuse artifacts from the other builds.

Fixes #378

### DIFF
--- a/ci/run
+++ b/ci/run
@@ -17,9 +17,6 @@ TIMEFORMAT='elapsed time: %R (user: %U, system: %S)'
 declare -r target_cache="/cache/target"
 
 export CARGO_HOME=/cache/cargo
-# Incremental builds use timestamps of local code. Since we always
-# check it out fresh we can never use incremental builds.
-export CARGO_BUILD_INCREMENTAL=false
 
 export RUSTC_WRAPPER=sccache
 export SCCACHE_DIR=/cache/sccache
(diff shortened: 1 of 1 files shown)


ci/run
```diff
@@ -85,8 +82,17 @@ tar -cpzf artifacts/radicle-registry-node.tar.gz -C target/release radicle-regis
...
 
 echo "--- Cleanup cache"
+
 # Remove all executables that need to be rebuild.
 find ./target -maxdepth 2 -executable -type f -exec rm {} \;
+
+# Remove incremental artifacts for local packages. These artifacts are
+# usesless for subsequent builds because incremental builds are based
+# on timestamps and we always do a fresh checkout of the repository.
+find ./target -name incremental -type d -exec rm -r {} \;
```
CodeSandwich, line 92: What will happen if we **don't** remove the artifacts? Will they be preserved for the subsequent calls, possibly from other branches?

geigerzaehler: If we don’t remove the artifacts we’ll keep accumulating them for every build and they will fill up the cache.

geigerzaehler:
> I wonder if it would be possible to build everything in `--debug` instead of `--release` except for `cargo build` iff the `main` branch is being built thinking

I tried this initially but if the node is not build with `--release` WASM execution is too slow and the node complains. Since we need to run a node on every build we need to build it with the release profile.

PR #458, opened 2020-05-18
Drop CheckVersion signed extension

### LINKED ISSUE #457: Consider dropping CheckVersion
>  I suspect that this happens because of the `CheckVersion` signed extension: https://substrate.dev/rustdocs/master/src/frame_system/lib.rs.html#1532. Still exploring this.

_Originally posted by @NunoAlexandre in https://github.com/radicle-dev/radicle-registry/pull/452#issuecomment-630173577_

### Problem

The `CheckVersion` signed extension requires the `spec_version` of the client's runtime to match the `spec_version` of the runtime of the node it connects to. If they don't match, all transactions submitted by the client are rejected.

This behaviour goes against our goal to be as compatible as technically possible between the client (runtime) and the node (runtime).

### Solution

I would just drop the usage of `CheckVersion` and later implement a similar signed extension that checks whether the two versions are compatible, something we still need to figure out how to do.

NunoAlexandre:
Closes #457 

`CheckVersion` requires a client and a node interacting to use the same
`spec_version` of the runtime, which is not something we want. Rather,
we want those to be as compatible as technically possible, something we
will tackle in the near future.


NunoAlexandre:
@geigerzaehler so the e2e tests fail because the wasm runtime expects the check version signed extension while the latest native runtime doesn't provide it :upside_down_face: any idea of how we can work around this and with such changes to signed extensions?

geigerzaehler:
> @geigerzaehler so the e2e tests fail because the wasm runtime expects the check version signed extension while the latest native runtime doesn't provide it upside_down_face any idea of how we can work around this and with such changes to signed extensions?

As mentioned on the issue, this change is a semantic change to the runtime and requires us to go through the upgrade process.

NunoAlexandre:
I have my hands tied to get this going. I can't run `./scripts/build-runtime-wasm` two times in a row, been forced to `cargo clean` after each call for some unknown reason, as the second call fails with an error like `mv: cannot stat '/tmp/tmp.WNHMfa3I2U/*': No such file or directory`, a symptom of the `cargo build` command not having generated any output.

Even when I have managed to build a new wasm and attempted to `runtime update`, it always fails with the vague `Invalid Transaction` error. Running the local dev node with the debug log flag but nothing useful is outputed.

The step I wish to take to get this PR in are:

0. Run `./scripts/build-dev-node && ./scripts/run-dev-node`
1. Run `cargo run -p radicle-registry-cli -- runtime version --node-host localhost`
    Verify that `spec_version` and `impl_version` are `4`
1. Run `cargo test` to verify that the e2e tests fail (Optional)
2. Increment the `spec_version` of the runtime (`runtime/src/lib.rs`)
3. Run `./scripts/build-runtime-wasm ./runtime/latest.wasm`
4. Stash the changes made to `runtime/src/lib.rs`
5. Update the on-chain runtime: `cargo run -p radicle-registry-cli -- runtime update ./runtime/latest.wasm --node-host localhost --author alice`
6. Unstash the changes made to  `runtime/src/lib.rs`
7. Run the e2e tests, that should now pass

NunoAlexandre:
@geigerzaehler this is ready for review. In terms of how to deploy this change, see https://github.com/radicle-dev/radicle-registry/issues/464#issuecomment-632241060.

runtime/src/lib.rs
```diff
@@ -81,7 +80,7 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<AccountId, Call, Signa
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("radicle-registry"),
     impl_name: create_runtime_str!("radicle-registry"),
-    spec_version: 4,
+    spec_version: 5,
     impl_version: 4,
```
geigerzaehler, line 84: Please reset `impl_version` [to 0](https://github.com/radicle-dev/radicle-registry/blob/master/DEVELOPING.md#semantic-updates)

geigerzaehler:
Could you please create a follow-up issue to add this check again once we’ve figured out the client compatibility story?

runtime/src/lib.rs
NunoAlexandre: Yes, I noticed that. I also have to bump the `spec_version` to `6` after https://github.com/radicle-dev/radicle-registry/pull/468 but I can't re-generate the wasm file: #469

Issue #455, opened 2020-05-18
Rename “TransactionApplied” to “TransactionIncluded”


geigerzaehler:
To avoid confusion between a transaction being successful (i.e. not resulting in a `DispatchError`) and a transaction being included in a block I suggest renaming `TransactionApplied` to `TransactionIncluded`. This applies to the public struct in the client and the one in the `backend` module.

Any thoughts @xla?

xla:
Looks good, we started to use the term `Confirmed` for the state of a transaction that has been included, but has not yet settled. In the domain of the app that makes more sense and is easier to grasp when communicated to a user.

NunoAlexandre:
Interesting. Could you expand how why that terminology is easier for users? I'd like to find one terminology that works for both if possible.

xla:
Inclusion here makes sense if one understands the mental model of the underlying datastructure. For a user who wants to know what the stauts of their tx is, it is not awfully helpful. What is giving more signal is that its confirmed and making progress on its way to settle. If you look at the product surface/specs you won't find even the reference of a block that often, yet the transaction states are: Pending -> Confirmed -> Settled. There is also Failed, which we use for all kinds of transaction failures.

NunoAlexandre:
Thanks, @xla. Wouldn't, however, `Confirm` sound like `Applied` and also give the idea that a tx was successful while in fact, it might have not been? :thinking:

xla:
From a domain perspective they have clear semantics and seem rather correct, wrt `Confirmed`. For the end-user experience you can check the design specs to see how confirmation are visually communicated, from that representation it's clear that the tx isn't successfull yet.

PR #459, opened 2020-05-18
Rename TransactionApplied to IncludedTransaction

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/455

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,10 @@ Upcoming
 
 ### Breaking changes
 
-* Rename `TransactionApplied` to `IncludedTransaction`
+* Rename `TransactionApplied` to `TransactionIncluded`
 * cli: Rename the key-pair storage file from 'accounts.json' to 'key-pairs.json'
 * cli: Move key-pair related commands under the new `key-pair` command group
-* client: `IncludedTransaction` result is now `Result<(), TransactionError>`
+* client: `TransactionApplied` result is now `Result<(), TransactionError>`
   for all messages.
 * cli: Drop `_TX_` from the environment options
 * cli: group and clean commands by domain and rename options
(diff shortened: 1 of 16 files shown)


xla:
For consideration https://github.com/radicle-dev/radicle-registry/issues/455#issuecomment-630506917

CHANGELOG.md
```diff
@@ -6,9 +6,10 @@ Upcoming
 
 ### Breaking changes
 
+* Rename `TransactionApplied` to `IncludedTransaction`
 * cli: Rename the key-pair storage file from 'accounts.json' to 'key-pairs.json'
 * cli: Move key-pair related commands under the new `key-pair` command group
-* client: `TransactionApplied` result is now `Result<(), TransactionError>`
+* client: `IncludedTransaction` result is now `Result<(), TransactionError>`
```
geigerzaehler, line 12: Does it make sense to retroactively change the change log? I think this misrepresents the change

NunoAlexandre: No, not at all, my bad.

NunoAlexandre:
> I would strongly suggest to use `TransactionIncluded` instead of `IncludedTransaction`. With the latter I would expect an `IncludedTransaction` to also be usable as a transaction. The adjective says basically “an included transaction is also a transaction”. But the struct does not reflect that: An included transaction only holds the hash of the transaction it references. IT cannot be used as a transaction.
> 
> On the other hand `TransactionIncluded` makes it clearer that the data is a representation of the fact that a transaction has been included. This also matches the documentation of the item much better: “Indicator that a transaction has been included in a block...”

I see your point. To be honest, I always struggled with `TransactionApplied` but I see your argument. To be the fairest to your point, `IncludedTxData` would be the clearest, as that is _exactly_ what the struct stands for: data relative to a transaction that was included in a block.

Thoughts?

NunoAlexandre:
@geigerzaehler regarding my comment above, let's now spend time there, I pushed a commit renaming the struct and all variables accordingly.

Issue #453, opened 2020-05-15
Client emulator should produce blocks continuously and manually


geigerzaehler:
The client emulator is used to run an app locally without being connected to a node but simulating the blockchain behavior. To function properly the app requires blocks to be continuously produced so that previously submitted transactions can be considered confirmed after reaching a certain depth.

Relatedly, for tests of the app it is desirable to put the emulator in a state where previously submitted can be considered confirmed.

At the moment the emulator only produces blocks when a transaction is submitted. 

## Proposal

When the client is used with the emulator backend we provide the user with a control interface that let’s the user manually advance the blockchain or setup automatic advancement.

```rust
let (client, emulator) = Client::new_emulator();
emulator.add_block();
let mining_interval = Duration::new(5, 0);
emulator.start_mining(mining_interval)
```

xla:
This proposal looks very good. Two things to add:

* `block_header_best_chain` should not return the most recent one, but an offset (maybe configurable) - so we can simulate settlement
* the first point paired with the `add_block` method would be enough already to satisfy our needs.

geigerzaehler:
Can’t you already simulate settlement with just the `add_block` method?
```
submit_tx(client);
// client is waiting for settlement
emulator.add_blocks(6)
// tx is settled with 6 confirmations.
```

Maybe what you want is that submitting a tx to the emulator does not automatically produce a block so that there is a state where the tx is pending?

xla:
> Maybe what you want is that submitting a tx to the emulator does not automatically produce a block so that there is a state where the tx is pending?

Not really, we rely on `block_header_best_chain` to advance the progress of txs. If it would all times return chain length - 6, combined with a method to add a block as the caller would seal the deal.

geigerzaehler:
@xla and I discussed this and it seems that having `add_block()` should be sufficient and we don’t need an offset for `block_header_best_chain()`. We need to validate this with a PR for this change.

geigerzaehler:
We’ve decided to not add the functionality to continously produce blocks in the background. With `add_block()` this can be implemented by the consumer.

NunoAlexandre:
@geigerzaehler do we want to keep this issue or do we consider it done at https://github.com/radicle-dev/radicle-registry/pull/460?

PR #460, opened 2020-05-19
Add control interface for client emulator

geigerzaehler:
Adds `EmulatorControl` and `add_blocks()` for #453. This PR does not yet at the capability to produce blocks automatically in the background.

A conscious decision was made to expose a separate `EmulatorControl` struct instead of `Emulator` so that the internal APIs of `Emulator` (e.g. the `Backend` implementation) are not leaked to the user.

### DIFF
--- a/client/src/backend/emulator.rs
+++ b/client/src/backend/emulator.rs
@@ -19,12 +19,12 @@ use futures::future::BoxFuture;
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
-use sp_runtime::{traits::Hash as _, BuildStorage as _, Digest};
+use sp_runtime::{traits::Block as _, traits::Hash as _, BuildStorage as _, Digest};
 use sp_state_machine::backend::Backend as _;
 
 use radicle_registry_runtime::{
-    registry, runtime_api, AccountId, BalancesConfig, GenesisConfig, Hash, Hashing, Header,
-    Runtime, RuntimeVersion,
+    registry, runtime_api, AccountId, BalancesConfig, Block, EventRecord, GenesisConfig, Hash,
+    Hashing, Header, Runtime, RuntimeVersion,
 };
 
 use crate::backend;
@@ -32,6 +32,8 @@ use crate::interface::*;
 
 /// [backend::Backend] implementation using native runtime code and in memory state through
 /// [sp_io::TestExternalities] to emulate the ledger.
+///
+/// The clone of an [Emulator] will share the state with the original emulator using a [Mutex].
 ///
 /// # Differences with real backend
 ///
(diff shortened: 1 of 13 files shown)


client/src/backend/emulator.rs
```diff
@@ -47,15 +48,33 @@ pub struct Emulator {
...
 /// Mutable state of the emulator.
 struct EmulatorState {
     test_ext: sp_io::TestExternalities,
     tip_header: Header,
     headers: HashMap<BlockHash, Header>,
 }
 
-/// Block author account used when the emulator creates blocks.
-pub const BLOCK_AUTHOR: AccountId = ed25519::Public([0u8; 32]);
+impl Default for Emulator {
```
CodeSandwich, line 73: Do we need that?

geigerzaehler: Clippy was complaining. I’d rather remove it

CodeSandwich: Ah, it thinks that `::new()` is worse than `::default()`. I can't say that I agree :weary:

xla:
While trying to test it I ran into the issue, that this change doesn't seem to be based on the `upstream` branch, which we require for certain features. Could we work on top of that branch, then I can start integrating it in radicle-dev/radicle-upstream#407.

geigerzaehler:
@xla: I’ve pushed a separate branch that bases these changes on upstream: `client-emulator-upstream`.

geigerzaehler:
@CodeSandwich: Could you have another look?

client/src/backend/emulator.rs
```diff
@@ -47,6 +50,32 @@ pub struct Emulator {
     state: Arc<Mutex<EmulatorState>>,
 }
 
+/// Control handle to manipulate the state of [Emulator].
+///
+/// Contrstruct this with [Emulator::control].
+pub struct EmulatorControl(Emulator);
```
CodeSandwich, line 56: Would it make sense to make it `#[derive(Clone, Debug)]`?

client/src/backend/emulator.rs
```diff
@@ -114,37 +142,47 @@ impl backend::Backend for Emulator {
...
-        let (new_tip_header, events) = state.test_ext.execute_with(move || {
+        let (block, event_records) = state.test_ext.execute_with(move || {
             runtime_api::initialize_block(&new_tip_header_init);
 
             let inherent_data = self.inherent_data_providers.create_inherent_data().unwrap();
             let inherents = runtime_api::inherent_extrinsics(inherent_data);
-            for inherent in inherents {
-                let _apply_result = runtime_api::apply_extrinsic(inherent).unwrap();
-            }
+            let extrinsics = [inherents, extrinsics].concat();
```
CodeSandwich, line 150: `Vec::append` would be easier to wrap head around.

client/src/backend/emulator.rs
geigerzaehler: That always makes sense

client/src/backend/emulator.rs
CodeSandwich: But then again we would get a `mut`.

geigerzaehler: I tried this but found it more confusing because it requires renaming.
```rust
inherents.append(&mut extrinsics);
let extrinsics = inherents;
```
I’d be ok with this change but would prefer to keep it.

client/src/backend/emulator.rs
geigerzaehler: We can’t derive `Debug` since `InherentDataProviders` does not implement it.

PR #461, opened 2020-05-20
Use different author for each test

geigerzaehler:
Backports changes that don’t affect the runtime from #400 to master. This should make it easier to integrate `upstream` with `master`.

### DIFF
--- a/client/tests/end_to_end.rs
+++ b/client/tests/end_to_end.rs
@@ -28,12 +28,12 @@ async fn register_project() {
     let _ = env_logger::try_init();
     let node_host = url::Host::parse("127.0.0.1").unwrap();
     let client = Client::create_with_executor(node_host).await.unwrap();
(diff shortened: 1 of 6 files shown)


NunoAlexandre:
I don't mind the diff, but could you explain in what way it makes it easier to integrate `upstream` with `master`?

geigerzaehler:
> I don't mind the diff, but could you explain in what way it makes it easier to integrate `upstream` with `master`?

When I merged `master` into `upstream` there were roughly 60 conflicting lines due to the changes in this PR which where on `upstream` and the changes in #459. This took quite some time to manually resolve.

This PR ensures that if we merge a change to master that changes the lines here it will be easier to merge this change to `upstream` too. In general we should keep the difference between `master` and `upstream` minimal so to make conflicts less likely.

Issue #463, opened 2020-05-20
Add client backward compatibility (by updating Substrate)


CodeSandwich:
We're in the process of upgrading Substrate to the newest master. We've hit an obstacle while upgrading to commit [2051ecbf79ed6a7262e8da1bfd97be4a423b9bcd](https://github.com/paritytech/substrate/commit/2051ecbf79ed6a7262e8da1bfd97be4a423b9bcd). It changes the `Weight` representation from u32 to u64. This data is stored in the state and it's returned over the RPC to the client when submitting transactions. Because the data is encoded using Scale codec, the decoder must know what to expect in advance. This causes a problem in the node, which won't be able to read the state set before the update (it **probably** won't need to ever do that). But the most importantly it will affect all the clients, they won't be able to tell, how to parse the data received over RPC and will need to be updated by all the users and be careful not to connect to an old node.

The main problems will be:
- handling 2 types of transactions (CheckVersion will make them incompatible after spec update)
- handling 2 types of events (old with u32, new with u64)
- making these changes invisible to the client API consumers

CodeSandwich:
There's no way to keep u32 weights in substrate. I've prepared a PR updating master as much as possible without touching the weights: https://github.com/radicle-dev/radicle-registry/pull/465

Next I think that we'll have to create another `upstream` branch, this time with a completely fresh genesis WASM runtime. We'll be able to merge it when we decide to make a full network reset and a breaking node and client release.

CodeSandwich:
The branch with the newest substrate and rebuilt genesis runtime WASM: [update_substrate](https://github.com/radicle-dev/radicle-registry/tree/update_substrate).

geigerzaehler:
Why do we need to use a “fresh genesis WASM runtime”? (I assume you are referring to `runtime/ffnet_genesis.wasm`.) Can’t we go through a proper upgrade process? Meaning we only update `runtime/latest.wasm` and keep the genesis so we don’t need to reset the ffnet?

Having another long living branch that we need to maintain is going to be a lot of work. (We’re already struggling with maintaining `upstream`.) The breaking change is only to the runtime modules so we could patch those modules temporarily with an old version and keep updating the rest of substrate.

CodeSandwich:
> Why do we need to use a “fresh genesis WASM runtime”?

This is a breaking change in the state. We don't have `System:Events` versioned yet, so we can't assure backward compatibility of a newer runtime. In order to properly add verioning we need a full chain restart.

[...]

PR #465, opened 2020-05-21
Update substrate somewhat

CodeSandwich:
Updates substrate to the commit right before switch of weight from u32 to u64. This is the most we can do without introducing a breaking change. Part of https://github.com/radicle-dev/radicle-registry/issues/463


NunoAlexandre:
Lookgs good to me, though I'd prefer us to update the whole way. I understand it means starting over, but we will have to do it anyway at some point. The earlier the better, it seems to me. What do you recon?

CodeSandwich:
I think that we don't need to update in one huge leap. There are a lot of changes already, it will be easier to manage them if we don't postpone them indefinitely.

NunoAlexandre:
> I think that we don't need to update in one huge leap. There are a lot of changes already, it will be easier to manage them if we don't postpone them indefinitely.

That's assuming that by including the breaking change that we will take much longer to update. You might be right.

CodeSandwich:
@geigerzaehler The CLI had to be reorganized for the new substrate API. Could you take a look as well at least to be aware of the changes?

NunoAlexandre:
> @geigerzaehler The CLI had to be reorganized for the new substrate API. Could you take a look as well at least to be aware of the changes?

what do you mean by organized?

CodeSandwich:
> what do you mean by reorganized?

The substrate CLI API has been changed, most importantly the (SubstrateCli)[https://substrate.dev/rustdocs/v2.0.0-alpha.8/sc_cli/trait.SubstrateCli.html] trait has been added, which forced some code to be pushed to different places.

node/src/cli.rs
```diff
@@ -96,54 +95,74 @@ pub struct Arguments {
...
+    fn support_url() -> &'static str {
+        "http://github.com/radicle-dev/radicle-registry/issues"
+    }
+
+    fn copyright_start_year() -> i32 {
+        2019
+    }
+
+    fn executable_name() -> &'static str {
+        "radicle-registry"
```
geigerzaehler, line 124:
```suggestion
        "radicle-registry-node"
```
Ideally we would get this from cargo with `env!` but I couldn’t find the right environment variable.

CodeSandwich: Done: https://github.com/radicle-dev/radicle-registry/pull/465/commits/8d76d044507e56606313587ba9af5566bd0a48d2 :smiley:

node/src/cli.rs
```diff
@@ -105,11 +105,11 @@ impl SubstrateCli for Cli {
     }
 
     fn description() -> &'static str {
-        "Radicle Registry Node"
+        env!("CARGO_PKG_DESCRIPTION")
```
NunoAlexandre, line 108: can this collide with other rust projects on the same machine?

CodeSandwich: I don't think so, this is only a part of a help message. This variable isn't magical either, it's set to a `Cargo.toml`'s `description`.

CodeSandwich:
FYI, may come in handy in the future :wink:: https://doc.rust-lang.org/cargo/reference/environment-variables.html

PR #472, opened 2020-05-26
runtime: Unregister user if not an org member

### LINKED ISSUE #399: Require user to not be a member of any org when unregistering
### Context
 
We found that the `user-unregistration` validation _"U must not be registered as a member in any org" is not in place at the moment.

_Originally posted by @NunoAlexandre in https://github.com/radicle-dev/radicle-registry/issues/396#issuecomment-618883022_

### Changes

When unregistering a user, require there to be no orgs where said user is a member. Fail otherwise.

At the moment, we don't store a map user -> orgs meaning 'the orgs where a user is a member', so we would need to go through all orgs. That is fine as a first version but highly inefficient in the long run. On the other hand, storing such map is also expensive since it means storing more data in the chain.

NunoAlexandre:
Closes #399

### DIFF
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -274,25 +274,19 @@ decl_module! {
 
         #[weight = SimpleDispatchInfo::InsecureFreeNormal]
         pub fn unregister_user(origin, message: message::UnregisterUser) -> DispatchResult {
-            fn can_be_unregistered(user: state::User, sender: AccountId) -> bool {
-                user.account_id == sender && user.projects.is_empty()
-            }
-
             let sender = ensure_signed(origin)?;
+            let sender_user = get_user_with_account(sender).ok_or(RegistryError::InexistentUser)?;
 
-            match store::Users::get(message.user_id.clone()) {
-                None => Err(RegistryError::InexistentUser.into()),
-                Some(user) => {
-                    if can_be_unregistered(user, sender) {
-                        store::Users::remove(message.user_id.clone());
-                        Self::deposit_event(Event::UserUnregistered(message.user_id));
-                        Ok(())
-                    }
-                    else {
-                        Err(RegistryError::UnregisterableUser.into())
-                    }
-                }
+            if sender_user.id != message.user_id {
+                return Err(RegistryError::InsufficientSenderPermissions.into());
             }
+            if find_org(|org| org.members.contains(&sender_user.id)).is_some() {
+                return Err(RegistryError::UnregisterableUser.into());
+            }
+
+            store::Users::remove(message.user_id.clone());
+            Self::deposit_event(Event::UserUnregistered(message.user_id));
+            Ok(())
         }
 
         #[weight = SimpleDispatchInfo::InsecureFreeNormal]
@@ -422,6 +416,12 @@ pub fn get_user_with_account(account_id: AccountId) -> Option<User> {
     store::Users::iter()
         .find(|(_, user)| user.account_id == account_id)
         .map(|(id, user)| User::new(id, user))
+}
+
+pub fn find_org(predicate: impl Fn(&state::Org) -> bool) -> Option<state::Org> {
+    store::Orgs::iter()
+        .find(|(_, org)| predicate(org))
+        .map(|(_, org)| org)
 }
 
 /// Check whether the user associated with the given account_id is a member of the given org.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Upcoming
 
 ### Breaking changes
 
+* Only unregistered a user if not a member of any org
 * Tx author needs to have an associated registered user to operate on Orgs
 * `Org::members` is now `Vec<Id>`
 * client: `Client::new_emulator()` now returns a pair of a `Client` and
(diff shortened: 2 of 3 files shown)


runtime/src/registry.rs
```diff
@@ -274,24 +274,19 @@ decl_module! {
...
-                        Ok(())
-                    }
-                    else {
-                        Err(RegistryError::UnregisterableUser.into())
-                    }
-                }
+            if sender_user.id != message.user_id {
+                Err(RegistryError::InsufficientSenderPermissions.into())
+            }
+            else if find_org(&|org| org.members.contains(&sender_user.id)).is_some() {
```
xla, line 283:
```suggestion
            } else if find_org(&|org| org.members.contains(&sender_user.id)).is_some() {
```

This formatting looks off.

runtime/src/registry.rs
```diff
@@ -424,6 +419,12 @@ pub fn get_user_with_account(account_id: AccountId) -> Option<User> {
         .map(|(id, user)| User::new(id, user))
 }
 
+pub fn find_org(predicate: &dyn Fn(&state::Org) -> bool) -> Option<state::Org> {
+    store::Orgs::iter()
+        .find(|(_, org)| predicate(org))
+        .map(|(_, org)| org)
```
xla, line 425: Can't this be merged into a [`FilterMap`](https://doc.rust-lang.org/std/iter/struct.FilterMap.html).

Are we planning to optimise this at some point, it seems expensive to iterate over the entire org space every time we do this check.

runtime/src/registry.rs
xla: Seems strange to have this extra function just to call a couple of iterator methods.

runtime/src/registry.rs
```diff
@@ -274,24 +274,19 @@ decl_module! {
...
-                        Err(RegistryError::UnregisterableUser.into())
-                    }
-                }
+            if sender_user.id != message.user_id {
+                Err(RegistryError::InsufficientSenderPermissions.into())
+            }
+            else if find_org(&|org| org.members.contains(&sender_user.id)).is_some() {
+                Err(RegistryError::UnregisterableUser.into())
+            }
+            else {
```
xla, line 286:
```suggestion
            } else {
```

Strange formatting.

runtime/src/registry.rs
NunoAlexandre: > Seems strange to have this extra function just to call a couple of iterator methods.

yeah. I just wanted to make the general logic reusable while still making it declarative and readable. we can go yagni style but it doesn't hurt at all to have it ready for reuse.

runtime/src/registry.rs
NunoAlexandre: TIL, should do that

NunoAlexandre: actually, opting for `filter_map` here results in more cumbersome code:

```suggestion
    .find_map(|(_, org)| if predicate(&org) { Some(org) } else { None })
```

Given that in this case the `map` only once, I'd rather keep this cleaner, more declarative structure.

NunoAlexandre: > Are we planning to optimise this at some point, it seems expensive to iterate over the entire org space every time we do this check.

Indeed, this is an expensive operation of O(n*m) complexity, where `n` is the number of orgs and `m` the number of members per org. However, considering the cost of storing stuff on-chain, I'm not convinced we need to look after this right now nor in the near future.

runtime/src/registry.rs
NunoAlexandre: :man_shrugging: 

cargo fmt seems happy with it. Do you know if 9or how we can automate your preference?

runtime/src/registry.rs
geigerzaehler, line 283: The rest of the runtime message handlers use guard-style ifs for condition checks. Could we also use this here for consistency?
```rust
if sender_user.id != message.user_id {
    return Err(RegistryError::InsufficientSenderPermissions.into())
}
```

runtime/src/registry.rs
```diff
@@ -424,6 +419,12 @@ pub fn get_user_with_account(account_id: AccountId) -> Option<User> {
         .map(|(id, user)| User::new(id, user))
 }
 
+pub fn find_org(predicate: &dyn Fn(&state::Org) -> bool) -> Option<state::Org> {
+    store::Orgs::iter()
```
geigerzaehler, line 423: Just like the iteration in `register_user` this will become extremely inefficient. Although it is fine for now I’m wondering whether we should track this in a an issue.

runtime/src/registry.rs
NunoAlexandre: No longer applicable.

runtime/src/registry.rs
NunoAlexandre: I'd say no. Far from advocating for ill-performant code, in this case, there's a clear trade-off and I'd rather wait until the Radicle Registry is so substantially used that this piece becomes a performance bottleneck.

NunoAlexandre:
> You also need to upgrade `spec_version` and recompile the latest runtime.

We don't really _need_ to. The e2e tests are not affected and in this branch, we run against the native runtime. Is there any practical need to bump it? If not, I'd leave it until this is merged to master otherwise we are both bumping it on master and here, which is just difficult to linearize.

runtime/src/registry.rs
geigerzaehler: I think this is something we should tackle earlier than so I created [an issue](https://github.com/radicle-dev/radicle-registry/issues/474)

Issue #389, opened 2020-04-22
Support project registration under org and user


NunoAlexandre:
Part 2 of #250

PR #477, opened 2020-05-29
User project registration - Prelude

NunoAlexandre:
[`Prelude`](https://www.youtube.com/watch?v=WXbb75TR9PM) for #389 

Introduces `ProjectDomain = Org(id) | User(id)`, used everywhere. Where before we'd have `org_id`, it's now `ProjectDomain::Org(org_id)`.

It does *not* support user project registration. That will come in the following PR.


core/src/state.rs
```diff
@@ -200,3 +200,17 @@ impl User {
         self
     }
 }
+
+/// The domain of a [Project]
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub enum ProjectDomain {
```
geigerzaehler, line 206: Can we move this to a different module? The `state` module should only include things that are stored in the chain state.

test-utils/src/lib.rs
```diff
@@ -51,7 +51,7 @@ pub async fn submit_ok<Message_: Message>(
 }
 
 pub async fn create_project_with_checkpoint(
-    org_id: Id,
+    domain: ProjectDomain,
```
geigerzaehler, line 54: It is more common to use references in arguments to avoid clones at the call site.

test-utils/src/lib.rs
```diff
@@ -67,23 +67,33 @@ pub async fn create_project_with_checkpoint(
...
     .unwrap();
 
-    let register_org_message = message::RegisterOrg {
-        org_id: org_id.clone(),
+    let domain_account = match &domain {
+        ProjectDomain::Org(org_id) => {
+            let register_org_message = message::RegisterOrg {
+                org_id: org_id.clone(),
+            };
+            submit_ok(&client, &author, register_org_message.clone()).await;
```
geigerzaehler, line 75: The `.clone()` call seems redundant. Same below.

test-utils/src/lib.rs
```diff
@@ -119,12 +129,12 @@ pub fn random_register_org_message() -> message::RegisterOrg {
 
 /// Create a [message::RegisterProject] with random parameters to register a project with.
 pub fn random_register_project_message(
-    org_id: Id,
+    domain: ProjectDomain,
```
geigerzaehler, line 132: See the comment about references above.

core/src/state.rs
NunoAlexandre: `ProjectDomain` is stored in the chain state, as part of the id of the `Projects` store.

test-utils/src/lib.rs
NunoAlexandre: Is there a technical reason as to why, for instance, `get_project` doesn't follow that recommendation or why that is not consistent across the repo?

https://github.com/radicle-dev/radicle-registry/blob/master/client/src/interface.rs#L137-L141

core/src/state.rs
geigerzaehler: But it’s not stored as a top-level object. For example, we can’t address a project domain by an ID. At the moment all ID types, in particular `ProjectId` are stored in the root module. We may reconsider that but for the moment we should keep it consistent.

test-utils/src/lib.rs
NunoAlexandre: Done btw.

test-utils/src/lib.rs
NunoAlexandre: Done.

test-utils/src/lib.rs
NunoAlexandre: Fixed.

test-utils/src/lib.rs
geigerzaehler: No specific reason. Just on oversight on my side.

core/src/state.rs
NunoAlexandre: Fixed.

core/src/lib.rs
```diff
@@ -57,7 +58,21 @@ pub type AccountId = ed25519::Public;
...
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub enum ProjectDomain {
+    Org(Id),
+    User(Id),
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for ProjectDomain {
+    fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self)
```
geigerzaehler, line 73: I think this crashes because it calls `ProjectDomain::fmt` recursively. Why not conditionally derive? `#[cfg_attr(feature = "std", derive(Display))]`

NunoAlexandre: ah, that's nice. TIL, and apparently we have never used it. I think now we don't need to control in great detail how `ProjectDomain` is printed so this is good.

geigerzaehler: I was wrong: Deriving obviously doesn’t work for `Display`, only for `Debug`.

NunoAlexandre: wait, that seems like it requires some dependency to be added, such as [enum_display_derive](https://docs.rs/enum-display-derive/0.1.0/enum_display_derive/).

NunoAlexandre: @geigerzaehler as I said, for now the `Debug` printing style is enough for us so I will just fix it to `{}` -> `{:?}`.

geigerzaehler: Yes, this is fine for now.

PR #479, opened 2020-06-02
Fix dev node build command

geigerzaehler:


scripts/build-dev-node
```diff
@@ -6,5 +6,5 @@
 set -euo pipefail
 
 cargo build -p radicle-registry-node --release
-(set +o pipefail;  yes | ./target/release/radicle-registry-node --chain dev purge-chain)
+(set +o pipefail;  yes | ./target/release/radicle-registry-node purge-chain --chain dev)
```
NunoAlexandre, line 9:
```suggestion
(set +o pipefail;  yes | ./scripts/run-dev-node purge-chain)
```

geigerzaehler: What you are suggesting will run the same command like before which does not work.

NunoAlexandre: I just ran it and it worked:

```
 ➜ ./scripts/run-dev-node purge-chain                                                                                     ~/registry  master ●
11:45:05.592 INFO  sc_cli  💬 Not registering Substrate logger, as there is already a global logger registered!
Are you sure to remove "/home/nuno/.local/share/radicle-registry/chains/dev/db"? [y/N]: y
"/home/nuno/.local/share/radicle-registry/chains/dev/db" removed.
```

PR #481, opened 2020-06-02
Support user project registration

### LINKED ISSUE #389: Support project registration under org and user

NunoAlexandre:
Closes #389 and #250


test-utils/src/lib.rs
```diff
@@ -215,3 +215,28 @@ pub async fn transfer(
         "Failed to grant funds to the recipient account."
     );
 }
+
+pub async fn project_domains(
```
geigerzaehler, line 219: This function is always paired with
```rust
let (author, user_id) = key_pair_with_associated_user(&client).await;
```
I’d suggest combining this with the code here in to a `create_project_domains` function. This way we don’t spread the creation of the org and the user between the two functions.

runtime-tests/tests/project_registration.rs
```diff
@@ -21,127 +21,261 @@
 use radicle_registry_client::*;
 use radicle_registry_test_utils::*;
 
+// Verify that not only can a project be registered under a user and an org.
```
geigerzaehler, line 24:
```suggestion
// Verify that a project can be registered under a user and an org.
```

runtime-tests/tests/project_registration.rs
```diff
@@ -21,127 +21,261 @@
...
+                assert!(
+                    user.projects.contains(&project.name),
+                    "The user does not contain the added project."
+                );
+
+                assert_eq!(
+                    client.free_balance(&user.account_id).await.unwrap(),
+                    initial_balance - random_fee,
+                    "The tx fee was not charged properly."
+                );
```
geigerzaehler, line 122: To avoid repeating the assertions you could return `(projects, account_id)` like you did for other tests.

runtime-tests/tests/project_registration.rs
```diff
@@ -21,127 +21,261 @@
...
-    let inexistent_org_id = random_id();
-    let message =
-        random_register_project_message(&ProjectDomain::Org(inexistent_org_id), checkpoint_id);
-    let tx_included = submit_ok(&client, &author, message.clone()).await;
+    let message = random_register_project_message(&domain_org_1, checkpoint_id);
+    submit_ok(&client, &author, message.clone()).await;
 
-    assert_eq!(tx_included.result, Err(RegistryError::InexistentOrg.into()));
+    // Duplicate submission under another domain.
+    let domain_org_2 = ProjectDomain::Org(register_random_org(&client, &author).await.id);
```
geigerzaehler, line 255: I find it easier to follow if you split async side effects
```suggestion
    let org = register_random_org(&client, &author).await;
    let domain_org_2 = ProjectDomain::Org(org.id);
```

runtime-tests/tests/project_registration.rs
```diff
@@ -21,127 +21,261 @@
...
+                domain
+            )
+        );
+    }
 }
 
+// Verify that two different orgs can have a project identified by the same name.
 #[async_std::test]
-async fn register_project_under_inexistent_org() {
+async fn re_register_project_org_domain() {
```
geigerzaehler, line 234: This name does not reflect what the test does. How about
```suggestion
async fn register_same_project_name_under_different_orgs() {
```

runtime-tests/tests/project_registration.rs
```diff
@@ -21,127 +21,261 @@
...
+    )
+    .await;
+
+    assert!(registration_2.result.is_ok());
 }
 
+// Verify that two different users can have a project identified by the same name.
 #[async_std::test]
-async fn register_project_with_duplicate_id() {
+async fn re_register_project_user_domain() {
```
geigerzaehler, line 271: See comment above.

client/tests/end_to_end.rs
```diff
@@ -29,94 +29,102 @@ async fn register_project() {
...
+                assert!(
+                    user.projects.contains(&project.name),
+                    "The user does not contain the added project."
+                );
+
+                assert_eq!(
+                    client.free_balance(&user.account_id).await.unwrap(),
+                    initial_balance - random_fee,
+                    "The tx fee was not charged properly."
+                );
```
geigerzaehler, line 124: See my comment about returning `(projects, account_id)` from the matcher.

test-utils/src/lib.rs
NunoAlexandre: I actually did that initially. The problem with doing that is that it couples concerns unnecessarily, having the code that generates project domains also registering a user  associated with the key pair. While that is fine for the org lane, where we can generate as many orgs as we wish, we can't make that so for users, where a key pair can only be associated with one user.

More so, we often need to use the `user_id`, so `project_domains` would need to return something like `(Id, Vec<ProjectDomain>)`, which is odd imo.

geigerzaehler: > it couples concerns unnecessarily, having the code that generates project domains also registering a user associated with the key pair.

In contrast, I find it odd that one piece of code registers the user while another generates the org and just uses the user. 

> More so, we often need to use the `user_id`, so `project_domains` would need to return something like `(Id, Vec<ProjectDomain>)`, which is odd imo.

I took a look and nowhere do we need the user ID except if were extracting it from `ProjectDomain::User`.

If you would still like to go with this I’d suggest we add some documentation for this function and give it a better name that suggest that it creates an org on the chain. Could you also split out the org creation into a separate statement to make it more readable.

runtime-tests/tests/project_registration.rs
NunoAlexandre: hm maybe it's not a clear name, but it does reflect what it does: it registered the same project under the `org` domain. But yeah, it's not clear, I like your suggestion better.

test-utils/src/lib.rs
NunoAlexandre: > I took a look and nowhere do we need the user ID except if were extracting it from `ProjectDomain::User`.

Just had a look as well and you are right!

> If you would still like to go with this I’d suggest we add some documentation for this function and give it a better name that suggest that it creates an org on the chain. Could you also split out the org creation into a separate statement to make it more readable.

Will do.

NunoAlexandre: Done. The `fn` is now:

``` rust
/// Generate project domains owned by the given `author`. It associates the author
/// with a random user and registers a random org with the author account.
pub async fn generate_project_domains(
    client: &Client,
    author: &ed25519::Pair,
) -> Vec<ProjectDomain> {
    let user_id = associate_key_pair_with_random_user(client, author).await;
    let org = register_random_org(&client, &author).await;

    vec![ProjectDomain::User(user_id), ProjectDomain::Org(org.id)]
}
```

PR #482, opened 2020-06-03
Decouple telemetry config from chain spec

geigerzaehler:
We leverage `RunCmd::no_telemetry` to decouple the telemetry configuration from the chain spec.

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -125,7 +125,7 @@ impl SubstrateCli for Cli {
     }
 
     fn load_spec(&self, id: &str) -> Result<Box<dyn ChainSpec>, String> {
-        let chain_spec = parse_chain(id)?.spec(!self.no_telemetry);
+        let chain_spec = parse_chain(id)?.spec();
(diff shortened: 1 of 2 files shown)


node/src/cli.rs
```diff
@@ -148,6 +148,7 @@ impl Cli {
     fn create_run_cmd(&self) -> RunCmd {
         // This does not panic if there are no required arguments which we statically know.
         let mut run_cmd = RunCmd::from_iter_safe(vec![] as Vec<String>).unwrap();
+        run_cmd.no_telemetry = self.no_telemetry;
```
CodeSandwich, line 151: Does it add `wss://telemetry.polkadot.io/submit/` telemetry endpoint?

geigerzaehler: Yes, it takes it from the chain spec: https://github.com/paritytech/substrate/blob/799f5c148f2c17d982a0fc395eff61d1baf47c34/client/cli/src/commands/run_cmd.rs#L330-L344

CodeSandwich: And a node is indeed visible on the polkadot telemetry site, nice!

PR #483, opened 2020-06-03
Disable RPC CORS on node configuration

geigerzaehler:
The previous approach of disabling RPC CORS on `RunCmd` did not work anymore. We need to disable it on the node configuration.

Note that we are only adjusting the config for the full node which is ok for now. I’d like to use [`Runner::config_mut`][1] but in our substrate version it has the wrong type.

[1]: https://releases.parity.io/substrate-rustdoc/sc_cli/struct.Runner.html#method.config_mut

### DIFF
--- a/node/src/cli.rs
+++ b/node/src/cli.rs
@@ -139,7 +139,12 @@ impl Cli {
                 .run_subcommand(subcommand, service::new_for_command),
             None => self.create_runner(&self.create_run_cmd())?.run_node(
                 service::new_light,
-                move |config| service::new_full(config, self.mine),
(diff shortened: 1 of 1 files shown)


CodeSandwich:
> Note that we are only adjusting the config for the full node which is ok for now.

Is it necessary to leave it partially fixed? I've prepared https://github.com/radicle-dev/radicle-registry/pull/486 and it seems to work fine.

CodeSandwich:
Sorry about the confusion, but #486 updates, not supersedes this PR :D I'll switch its merge target to `master`.

PR #484, opened 2020-06-03
Fix CLI runtime command help

geigerzaehler:
* Fix the description of the runtime group
* Move details about the update from the command help (which was not properly formatted with regards to newlines) to the developer docs.
* Don’t print irrelevant version information.


geigerzaehler:
> Though I would like some docs on the requirements to have the CLI command run successfully, but since it's only us running it this is good to go.

This is documented in `DEVELOPING.md`. I think it’s more approriate there because there is more context.

PR #487, opened 2020-06-03
Add CI check to forbid fixup commits on pull requests

geigerzaehler:
Add CI check to forbid commits that start with "fixup!" in pull requests

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -1,18 +1,27 @@
+env:
+  DOCKER_IMAGE: gcr.io/opensourcecoin/radicle-registry/ci-base:c068afc4d2012cd70a9943ead3671e2e5b35508a
+  DOCKER_FILE: ci/base-image/Dockerfile
+
 .test: &test
   label: "Test"
   command: "ci/run"
   timeout_in_minutes: 60
   env:
-    DOCKER_IMAGE: gcr.io/opensourcecoin/radicle-registry/ci-base:c068afc4d2012cd70a9943ead3671e2e5b35508a
-    DOCKER_FILE: ci/base-image/Dockerfile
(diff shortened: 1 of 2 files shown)


CodeSandwich:
Our workflow usually involves pushing without force a few fixup commits to take into account review comments. This change will make the CI always failing and red for the whole review duration except before it starts (vanilla commits) and after it ends (squashed commits).

This check also helps only if a strict `fixup` commit naming convention is followed.

If this is merged, my incentive will be to never use `fixup` in commit names to prevent false positives in CI.

NunoAlexandre:
What motivates this, @geigerzaehler?

CodeSandwich:
> What motivates this, @geigerzaehler?

I think that it's a reminder to squash.

geigerzaehler:
This doesn’t work yet since our CI does not pass the base branch environment variable correctly to the build script.

> What motivates this, @geigerzaehler?

The fact that we sometimes forget to squash fixup commits: `git log origin/master --format=%s | grep fixup`.

> This change will make the CI always failing and red for the whole review duration except before it starts (vanilla commits) and after it ends (squashed commits).

It only marks the check `buildkite/radicle-registry/check-fixup-commits ` as failed. Does this change your view?

CodeSandwich:
> It only marks the check buildkite/radicle-registry/check-fixup-commits as failed. Does this change your view?

Only a little. It still requires a strict naming convention and produces noise in the CI results. I think that the list of commits should be reviewed before hitting the merge button anyway.

NunoAlexandre:
I agree with @CodeSandwich.

The thing I see is that if we have a PR with:

- Add X
- fixup! Y
- Add Z

this will still make its way into master, or not?

geigerzaehler:
Since there’s no strong support for this and the benefits are likely marginal I’ll close this.

PR #490, opened 2020-06-04
client: Provide naive deposit_costs method

### LINKED ISSUE #485: client: Provide API to get deposit costs for a given tx
Support work for https://github.com/radicle-dev/radicle-upstream/issues/391.

In a call with @rudolfs we agreed to have the registry client providing a stub API to obtain the deposit costs for a given tx. For now, we can just hardcode those values to quickly support the UI work.

@cloudhead do we have the deposit values defined somewhere for each tx?

NunoAlexandre:
Closes #485 

This allows the upstream to get the deposit costs associated with a
given Message. It is a naive implementation to allow the upstream to
move forward with a real API and it will be followed up by me in the
coming week, where tx deposits will be implemented.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -45,6 +45,8 @@ pub use pallet_balances as balances;
 pub use radicle_registry_core::*;
 pub use runtime_api::{api, RuntimeApi};
 
+pub type RegistryCall = registry::Call<Runtime>;
+
 /// An index to a block.
 pub type BlockNumber = u32;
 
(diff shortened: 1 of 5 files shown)


NunoAlexandre:
@rudolfs could you have a look? I am interested as to whether providing a `Message` works on your end and what more transactions/messages we want to have with an associated deposit? Maybe also `CreateCheckpoint`? Since this is just a stub it doesn't matter too much but we can always just get that in shape now.

runtime/src/lib.rs
```diff
@@ -202,3 +204,17 @@ construct_runtime!(
...
 );
+
+/// Get the deposit costs associated with a given call.
+pub fn deposit_costs(call: &Call) -> Balance {
+    match call {
+        Call::Registry(registry_call) => match registry_call {
+            RegistryCall::register_project(_)
+            | RegistryCall::register_member(_)
+            | RegistryCall::register_user(_)
+            | RegistryCall::register_org(_) => 10,
```
cloudhead, line 215: Is `Balance` denominated in uRAD or RAD? I would assume the former? If so, this should probably be something like `10 * COIN_VALUE` where `COIN_VALUE = 1_000_000`, otherwise this will show up as `0.000001` in the UI.

NunoAlexandre: `Balance` represents `uRAD`. This is really just a naive temporary implementation so that @rudolfs could work against a real API from the upstream. Your remark is useful anyway, of course :+1:

cloudhead: Yup, it'll have to be multiplied by 1 million then in the UI @rudolfs

Issue #491, opened 2020-06-05
cli: Support user project registration


NunoAlexandre:
At the moment, `radicle-registry-cli project register` only operates on the `org` domain, not supporting registering projects under users.

I see two ways of doing this:

#A - Add an argument of type `DomainType = Org | User` to the command, which would be specified as: `radicle-registry-cli project register <user|org> <id>`

#B - Offer optional flags `--user <id>` and `--org <id>`, where one should be specified. If more or less is specified, an error is shown. The command would be specified as `radicle-registry-cli project register --user <id>` or `radicle-registry-cli project register --org <id>`

Option #A seems much simpler.

PR #494, opened 2020-06-08
cli: Support user project registration

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/491

### DIFF
--- a/cli/src/command/project.rs
+++ b/cli/src/command/project.rs
@@ -16,6 +16,7 @@
 //! Define the commands supported by the CLI related to Projects.
 
 use super::*;
+use structopt::clap::arg_enum;
 
 /// Project related commands
 #[derive(StructOpt, Clone)]
@@ -93,8 +94,17 @@ impl CommandT for List {
 pub struct Register {
     /// Name of the project to register.
     project_name: ProjectName,
-    /// Org under which to register the project.
-    org_id: Id,
+
+    /// The type of domain under which to register the project.
+    #[structopt(
+        possible_values = &DomainType::variants(),
+        case_insensitive = true,
+    )]
+    domain_type: DomainType,
+
+    /// The id of the domain under which to register the project.
+    domain_id: Id,
+
     /// Project state hash. A hex-encoded 32 byte string. Defaults to all zeros.
     project_hash: Option<H256>,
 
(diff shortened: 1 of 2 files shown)


cli/src/command/project.rs
```diff
@@ -143,8 +156,56 @@ impl CommandT for Register {
...
             "✓ Project {}.{} registered in block {}",
-            self.project_name, self.org_id, project_registered.block,
+            self.project_name, project_domain, project_registered.block,
         );
         Ok(())
     }
 }
+
+#[derive(Clone, Eq, PartialEq, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "lowercase")]
```
geigerzaehler, line 166: Is there a reason why we implement `Deserialize` and `Serialize` for this type?

cli/src/command/project.rs
```diff
@@ -143,8 +156,56 @@ impl CommandT for Register {
...
+}
+
+impl FromStr for DomainType {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "org" => Ok(Self::Org),
+            "user" => Ok(Self::User),
+            _ => Err("Invalid DomainType".to_string()),
```
geigerzaehler, line 179: This is the message that is displayed to the user when they use a wrong domain. To make the message more helpful it should avoid code terminology (`DomainType`), repeat the input string, and list the possibilities.
```suggestion
            input => Err(format!("Invalid domain type \"{}\". Must be either org or user.", value)),
```

NunoAlexandre: It actually isn't given that we define the possible values in the struct opt construct. This case should therefore never happen and the user should never see that error message.

cli/src/command/project.rs
NunoAlexandre: My bad, forgot to remove it. I had copy pasted it from another location.

cli/src/command/project.rs
NunoAlexandre: To be more exact, the error users get is:

```
error: 'ORG' isn't a valid value for '<domain-type>'
        [possible values: org, user]
```

cli/src/command/project.rs
```diff
@@ -143,8 +156,55 @@ impl CommandT for Register {
...
             "✓ Project {}.{} registered in block {}",
-            self.project_name, self.org_id, project_registered.block,
+            self.project_name, project_domain, project_registered.block,
         );
         Ok(())
     }
 }
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+enum DomainType {
```
CodeSandwich, line 166: Have you considered using https://docs.rs/clap/2.33.1/clap/macro.arg_enum.html?

CodeSandwich: Cool!

NunoAlexandre: TIL. Introduced it now. It implements `FromStr` and `Display` out of the box and dry's the `possible_values` property :+1:

PR #500, opened 2020-06-09
Static ffnet chain spec

### LINKED ISSUE #493: Fresh nodes cannot sync to ffnet
Since f13c1d7d3fcccad484b2b580b25051f8ce3301dc (#458) nodes that are started fresh (i.e. without having any blocks in their database) fail to sync with the ffnet. The reason is that nodes after this commit initialize the database with a different genesis block and genesis state. New nodes print
~~~
Initializing Genesis block/state (state: 0xa0f9…17fd, header-hash: 0x4bd9…42fb)
~~~
while nodes before the offending commit print
~~~
Initializing Genesis block/state (state: 0x39e1…b18d, header-hash: 0x53f0…2d3f)
~~~

We need to further investigate what the cause in this difference is and how we can ensure that new nodes use the same genesis state and block.

geigerzaehler:
Fixes #493.

Before this change the genesis state was dynamically generated using the
runtime code. Since the runtime code has changed (specifically the
`spec_version`) we were now generating a different genesis state—and
thus genesis block—for the ffnet. As a consequence nodes compiled with
the newer runtime failed to join the ffnet.

With this change we fix the genesis state for the ffnet once and for all
in JSON chain spec and check it into the repo. We also document the
process and remove the unused `ffnet-genesis.wasm` runtime code.


CodeSandwich:
I don't understand this change. Why parsing the genesis spec from JSON is better than building it in code? Where's the genesis runtime, is it encoded in this weird structure in `genesis.raw.top`?

geigerzaehler:
> Why parsing the genesis spec from JSON is better than building it in code?

The genesis spec (specifically the genesis state) is built using the runtime code. This means that changes to the runtime code might result in changes to the genesis state. Indeed this is what happens if we bump the `spec_version`. The runtime code (specifically `GenesisConfig::build_storage`) generates different genesis state depending on `spec_version`.

> Where's the genesis runtime, is it encoded in this weird structure in `genesis.raw.top`?

Yes

CodeSandwich:
So when the genesis runtime WASM is loaded from code, it uses the native runtime to create the genesis state, but if the genesis runtime WASM is loaded from JSON, the genesis state is generated using the WASM runtime?

geigerzaehler:
> So when the genesis runtime WASM is loaded from code, it uses the native runtime to create the genesis state.

Yes. This is the case if we use `ChainSpec::from_genesis`.

> but if the genesis runtime WASM is loaded from JSON, the genesis state is generated using the WASM runtime?

Not quite. The genesis state is contained in the JSON representation of the chain spec and can just be loaded. That means it is generated when the chain spec is built and will never change.

CodeSandwich:
Aah, `genesis.raw.top` contains the state, which includes the runtime WASM. That's why there are so many blobish entries of which one is enormous.

Issue #495, opened 2020-06-08
Client validates runtime version


CodeSandwich:
The client should validate the runtime version of a node it connects to. See [this thread](https://radicle.community/t/cli-runtime-compatibiliy/221).

geigerzaehler:
@CodeSandwich: Could you provide more details? What does it mean to “validate”?

CodeSandwich:
The client library must check if the runtime spec version is an explicitly supported one. If not, it should refuse to connect.

PR #503, opened 2020-06-09
Client checks runtime version

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/495

### DIFF
--- a/client/src/error.rs
+++ b/client/src/error.rs
@@ -29,6 +29,9 @@ pub enum Error {
     /// Invalid transaction
     #[error("Invalid transaction")]
     InvalidTransaction,
+    /// Chain is running an incompatible runtime specification version
+    #[error("Chain is running an incompatible runtime specification version {0}")]
+    IncompatibleRuntimeVersion(u32),
     /// Other error
     #[error("Other error: {0}")]
     Other(String),
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -70,6 +70,7 @@ impl RemoteNode {
             chain: channel.clone().into(),
             author: channel.clone().into(),
         });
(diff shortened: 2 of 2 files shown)


client/src/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum Error {
     /// Invalid transaction
     #[error("Invalid transaction")]
     InvalidTransaction,
+    /// Connected to a node with runtime in incompatible specification version
```
NunoAlexandre, line 32:
```suggestion
    /// Connected to a node with an incompatible runtime specification version
```

client/src/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum Error {
...
+    #[error("Connected to a node with runtime in incompatible specification version {0}")]
+    IncompatibleSpecVersion(u32),
```
geigerzaehler, line 34: Mentioning the node might confuse people because it sounds like the runtime is something specific to the node. 
```suggestion
    /// Chain is running incompatible runtime spec version
    #[error("Chain is running incompatible runtime spec version {0}")]
    IncompatibleRuntimeVersion(u32),
```

CodeSandwich: Fixed

NunoAlexandre: I liked that the variant was explicit about the incompatibility being related to the spec version.

client/src/error.rs
CodeSandwich: Fixed

NunoAlexandre:
@CodeSandwich `nit` (!) but how do you feel about renaming the commit to `client: Check runtime spec_version compatibility`?

CodeSandwich:
@NunoAlexandre Renamed

Issue #445, opened 2020-05-08
runtime: Devise way to make state changes backwards-compatible


NunoAlexandre:
Part of #434 // State changes

### The problem

Here lives the hairiest of the beasts. Say, we move from having the members of an organization stored as account ids to user ids. Such change fundamentally impacts the way different versions parse and handle data from the chain.

We could decide, in this early stage, to hard-fork when faced with such type of change, meaning that we would run a new, parallel network where these breaking changes would start making their way in a new, fresh ground. However, that is not going to cut once we go main net, so we are focused on learning how to do this properly as early as possible.

We need to be backwards-compatible concerning the chain state. We need to devise a way to deserialize and support older forms of the data that might still be on-chain, as well as the new one. Given the example above, when deserializing Orgs, we need to think of the old and the new variants, the one where its members were account ids and the one where they now are user ids. To help users adjust to new realities, we must provide methods for them to “migrate their data” to its latest form, not only giving control to users but also keeping everything tracked in verifiable transactions.

This category might introduce breaking changes to the CLI. To cover for this, we need to have our [client ensure that the version of the runtime the node it is connected to is not greater than the one it supports](#26). That means that the state changes that impact the CLI require a new version of the CLI to be installed. 

### How

_First, the requirements:_
📌 Support multiple versions of stored types
📌 Converge old types to their new form when deserializing the storage
📌 Offer methods for users to update their entities to the latest version form or logic

_Proposal:_
Add a tag version to every type stored on-chain. Probably the best way to go is with [serde adjacently tagged enums](https://serde.rs/enum-representations.html#adjacently-tagged).
Each version of an entity on-chain will have a corresponding variant, which will be deserialized respectively. Since we haven't introduced this before making to the ffnet, we will need to have the set up so that if no version tag is present, a variant representing the version 0 is used.

To converge and uniformize those entities, we might have to then have a struct representing the latest variant, since we don't want the system to necessarily support all the variants for all code paths. Something to think further about.

NunoAlexandre:
See @kim's input on this at https://radicle.community/t/rambles-on-updating-the-radicle-registry/210/4

> Yaaa, one of the many issues with serde is that those tags are all strings — this means that a rename (or a typo) can break things in unexpected ways, and it’s hard to guard against this. Consider that there may exist third-party clients at some point.
>
[...]

PR #504, opened 2020-06-09
Document policy for versioning state

geigerzaehler:
See #445.

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -1,5 +1,4 @@
-radicle-registry-core
-=====================
+# radicle-registry-core
 
 This package provides the types that constitute the registry ledger and provides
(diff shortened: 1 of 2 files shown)


core/README.md
```diff
@@ -44,3 +41,27 @@ For each entity the documentation has the following sections
   <dt>Relevant messages</dt>
   <dd>Links to message types that effect or use the entity.</dd>
 </dd>
+
+### Versioning
+
+To make runtime state backwards compatible every state entity that is added item
+must be versioned using the following schema.
```
NunoAlexandre, line 48:
```suggestion
To make the runtime state backwards compatible, every state entity that is added
must be versioned using the following schema.
```

PR #507, opened 2020-06-10
Add opaque block so node can deal with old blocks

### LINKED ISSUE #505: Fresh nodes fail to mine for ffnet
When starting a fresh mining node (with an empty database) we see the following error message:
~~~
sc_consensus_pow  Mining block failed with BlockProposingError("Msg(\"Failed to decode result
of `inherent_extrinsics`: Error decoding field Call :: Registry.0\")")
~~~
And, although it discovers peers, the mining nodes fails to sync and import blocks.

geigerzaehler:
We make the runtime block type independent of the node block type so that runtime changes don’t break the node. Fixes #505.

Recently the `UncheckedExtrinsic` type has changed in the runtime. As a consequence the node was not able to deserialize old blocks containing the old `UncheckedExtrinsic` when synching.


geigerzaehler:
Incidentally this made the decoding fix redundant.

CodeSandwich:
> Incidentally this made the decoding fix redundant.

Is it because substrate fixed it? If so, can https://github.com/paritytech/substrate/issues/5359 be closed?

Issue #506, opened 2020-06-10
core: Provide necessary trait implementations 


NunoAlexandre:
## The problem

> > I see that we duplicate some of the registry types, such as Id and ProjectName. I followed the example for ProjectDomain, but I wonder if there's another motive to do this besides being able to derive certain traits that are not derived in the Registry?
>
> The main reason for the duplication is serialisability esp. types who want to use hex for that. Would love to avoid it by moving it upstream into the registry repo.

Quote from https://github.com/radicle-dev/radicle-upstream/pull/453#pullrequestreview-425900351

The `upstream` now need to duplicate the types below to get them to follow specific necessary traits.

We should move them to the registry since they are general-purpose.

## Types

This consideration affects the following types:

### `Id`
https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L23

This types implements:

* `Display`
      https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L25

* `TryFrom<String>`
    https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L31

* `TryFrom<&str>`
  https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L39

* Serde `Deserialize`
  https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L48

* Serde `Serialize`
  https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L61

### `ProjectName`
https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L23

This types implements:

* `Display`
      https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L74

* `TryFrom<String>`
    https://github.com/radicle-dev/radicle-upstream/blob/3d856233e2dfa74304349ec082db27245f3849bd/proxy/src/registry.rs#L104

* `TryFrom<&str>`
[...]

PR #511, opened 2020-06-10
core: Provide trait impls needed in the upstream

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/506

### DIFF
--- a/cli/src/command/project.rs
+++ b/cli/src/command/project.rs
@@ -65,7 +65,7 @@ impl CommandT for Show {
                 project_name: self.project_name.clone(),
                 project_domain,
             })?;
-        println!("Project: {}.{}", project.name, project.domain);
+        println!("Project: {}.{:?}", project.name, project.domain);
         println!("Checkpoint: {}", project.current_cp);
         Ok(())
     }
@@ -84,7 +84,7 @@ impl CommandT for List {
         let project_ids = client.list_projects().await?;
         println!("PROJECTS ({})", project_ids.len());
         for (name, org) in project_ids {
-            println!("{}.{}", name, org)
+            println!("{}.{:?}", name, org)
         }
         Ok(())
     }
@@ -156,7 +156,7 @@ impl CommandT for Register {
         let project_registered = register_project_fut.await?;
         project_registered.result?;
         println!(
-            "✓ Project {}.{} registered in block {}",
+            "✓ Project {}.{:?} registered in block {}",
             self.project_name, project_domain, project_registered.block,
         );
         Ok(())
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -14,6 +14,7 @@ default = ["std"]
 std = [
     "parity-scale-codec/std",
     "rand",
+    "serde",
     "sp-core/std",
     "sp-runtime/std",
     "thiserror",
@@ -22,6 +23,7 @@ std = [
 [dependencies]
 derive-try-from-primitive = "1.0.0"
 rand = { version = "0.7.2", optional = true }
+serde = { version = "1.0", features = ["derive"], optional = true }
 thiserror = { version = "1.0", optional = true }
 
 [dependencies.parity-scale-codec]
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -60,17 +60,19 @@ pub type Balance = u128;
 /// The id of a project. Used as storage key.
 pub type ProjectId = (ProjectName, ProjectDomain);
 
-/// The domain of a [Project]
+/// The domain under which a [Project] lives.
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
 pub enum ProjectDomain {
     Org(Id),
     User(Id),
 }
 
-#[cfg(feature = "std")]
-impl std::fmt::Display for ProjectDomain {
-    fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{:?}", self)
+impl ProjectDomain {
+    pub fn id(&self) -> Id {
+        match self {
+            Self::Org(id) | Self::User(id) => id.clone(),
+        }
     }
 }
 
--- a/core/src/project_name.rs
+++ b/core/src/project_name.rs
@@ -21,6 +21,8 @@ use core::convert::{From, Into, TryFrom};
 use parity_scale_codec as codec;
 
 #[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "std", serde(try_from = "String"))]
+#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
 pub struct ProjectName(String);
 
 impl ProjectName {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3530,6 +3530,7 @@ dependencies = [
  "derive-try-from-primitive",
  "parity-scale-codec",
  "rand 0.7.3",
+ "serde",
  "sp-core",
  "sp-runtime",
  "thiserror",
--- a/cli/src/lib.rs
+++ b/cli/src/lib.rs
@@ -155,7 +155,7 @@ pub enum CommandError {
     #[error("cannot find user {user_id}")]
     UserNotFound { user_id: Id },
 
-    #[error("cannot find project {project_name}.{project_domain}")]
+    #[error("cannot find project {project_name}.{project_domain:?}")]
     ProjectNotFound {
         project_name: ProjectName,
         project_domain: ProjectDomain,
--- a/core/src/id.rs
+++ b/core/src/id.rs
@@ -23,6 +23,8 @@ use core::convert::{From, Into, TryFrom};
 use parity_scale_codec as codec;
 
 #[derive(codec::Encode, Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "std", serde(try_from = "String"))]
+#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
 pub struct Id(String);
 
 impl Id {
--- a/runtime-tests/tests/project_registration.rs
+++ b/runtime-tests/tests/project_registration.rs
@@ -209,7 +209,7 @@ async fn re_register_project_same_domain_entity() {
         assert!(
             projects_list.contains(&project.name),
             format!(
-                "Registered project not found in the project list of {}",
+                "Registered project not found in the project list of {:?}",
                 domain
             )
         );


core/src/lib.rs
```diff
@@ -60,20 +62,47 @@ pub type Balance = u128;
...
     fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
         write!(f, "{:?}", self)
     }
 }
 
+arg_enum! {
+    /// The type of domains under which a project can live.
+    #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+    #[serde(rename_all = "lowercase")]
+    pub enum DomainType {
```
geigerzaehler, line 91: The domain type is only used by the CLI parser. I don’t see a reason to move it into core especially since it introduces a dependency on `clap`.

CodeSandwich:
This branch can't compile the WASM runtime :grimacing: It introduces serde, which relies on std. We need to switch it to the [no-std](https://serde.rs/no-std.html) mode.

I'm not sure if these types should care about serde encoding in the first place. Runtime doesn't need serde or its encoding, this is functionality used only in a dependee. This reverses the dependency flow and makes changes creep from the outside of our system to its core. Maybe [remote derive](https://serde.rs/remote-derive.html) could be used here?

geigerzaehler:
@CodeSandwich: We are discussing this in #506

cli/src/command/project.rs
```diff
@@ -97,7 +96,7 @@ pub struct Register {
 
     /// The type of domain under which to register the project.
     #[structopt(
-        possible_values = &DomainType::variants(),
+        possible_values = &["user", "org"],
```
CodeSandwich, line 99: Even if `arg_enum` can't be applied on `DomainType`, `variants` should be implemented for it manually. Otherwise we have the knowledge about possible variants and their `FromStr` representations distributed over the codebase with no checks.

core/src/lib.rs
```diff
@@ -60,20 +61,57 @@ pub type Balance = u128;
...
+        match self.clone() {
+            Self::Org(id) | Self::User(id) => id,
+        }
+    }
+}
+
 #[cfg(feature = "std")]
 impl std::fmt::Display for ProjectDomain {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
         write!(f, "{:?}", self)
```
CodeSandwich, line 82: This delegates `Display` to `Debug`, why is it needed in the first place?

core/src/lib.rs
```diff
@@ -60,20 +61,57 @@ pub type Balance = u128;
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "org" => Ok(Self::Org),
+            "user" => Ok(Self::User),
+            _ => Err("Invalid DomainType variant"),
+        }
+    }
+}
+
+impl From<(DomainType, Id)> for ProjectDomain {
```
CodeSandwich, line 106: Maybe this should be a separate function like `new`?

core/src/project_name.rs
```diff
@@ -106,6 +107,30 @@ impl core::fmt::Display for ProjectName {
     }
 }
 
+impl<'de> Deserialize<'de> for ProjectName {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s: &str = Deserialize::deserialize(deserializer)?;
```
CodeSandwich, line 115: AFAIR deserializing to `&str` requires the input text to be fully available in the memory. All it does is borrow a slice of it. If it's deserialized from a UTF-8 JSON in a `String`, it will work, but if it comes from a stream like an open file, it will fail in runtime.

core/src/project_name.rs
```diff
@@ -106,6 +107,30 @@ impl core::fmt::Display for ProjectName {
...
+        Self::try_from(s).map_err(|_err| {
```
CodeSandwich, line 116: Have you considered serde's [try_from](https://serde.rs/container-attrs.html#try_from)?

core/src/lib.rs
```diff
@@ -60,20 +61,57 @@ pub type Balance = u128;
...
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
 pub enum ProjectDomain {
     Org(Id),
     User(Id),
 }
 
+impl ProjectDomain {
+    pub fn id(&self) -> Id {
+        match self.clone() {
+            Self::Org(id) | Self::User(id) => id,
```
CodeSandwich, line 74: Nitpick: `clone()`ing an `id` would be cleaner than `clone()`ing `self`.

NunoAlexandre:
@CodeSandwich I appreciate your review but this PR is in draft

core/src/lib.rs
NunoAlexandre: That's right, thanks!

NunoAlexandre:
> This branch can't compile the WASM runtime  It introduces serde, which relies on std. We need to switch it to the [no-std](https://serde.rs/no-std.html) mode.

I tried switching to serde's no-std mode but I still can't get it going. Maybe tomorrow we can quickly look into it together :+1:

core/src/lib.rs
NunoAlexandre: The dependency on `clap` has been dropped. I'd like to have it on `core` to have these types together. It can also go to `client`, but I'd prefer to have it here.

core/src/lib.rs
NunoAlexandre: I'd prefer to keep it just to avoid bumping against the compiler saying that `ProjectDomain` can't be formatted using the default formatter.

core/src/lib.rs
NunoAlexandre: They are equivalent solutions, and this one is already in place in the upstream (`(domain_type, id).into()`), so i'd keep this as is.

core/src/project_name.rs
NunoAlexandre: How would it work? would it become:

```rust
#[cfg_attr(feature = "std", serde(try_from = "String"))]
pub struct ProjectName(String);
```

and this `impl<'de> serde::Deserialize<'de> for ProjectName` impl would be removed? I believe that'd be it but appreciate the confirmation.

NunoAlexandre: (I believe that's it, pushed in https://github.com/radicle-dev/radicle-registry/pull/511/commits/53639d58c8d6d12a6b8f5ac9d99c67da8b746310)

core/src/project_name.rs
NunoAlexandre: No longer applicable. thanks for the insight, though!

cli/src/command/project.rs
NunoAlexandre: Having a `fn variants() -> vec[str]` or something alike in `DomainType` doesn't really help dry that logic in the `FromStr` impl though. Unless if we guard the input against this array, but then we have to pattern match anyway to build the specific variant, so it seems kind of counter-productive. In the `node` cli we also have the `chain` arg `possible_values` detached from the enum, and we don't seem to see an issue with that. either.

core/src/lib.rs
```diff
@@ -60,20 +60,59 @@ pub type Balance = u128;
...
     fn fmt(&self, f: &mut core::fmt::Formatter) -> std::fmt::Result {
         write!(f, "{:?}", self)
     }
 }
 
+/// The type of domains under which a project can live.
+#[derive(Clone, Debug, Eq, PartialEq)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "std", serde(rename_all = "lowercase"))]
+pub enum DomainType {
```
geigerzaehler, line 99: This type is used only by the CLI and internally by upstream. I don’t think it is a good idea to share the code because the purpose is very different. Also `core` is not the right place for it since this is used by neither the client nor the runtime. If we need to change this type this effect would ripple through all dependents. I would also argue that upstream should not use this type and use `ProjectDomain` proper instead.

core/Cargo.toml
```diff
@@ -22,6 +22,7 @@ std = [
 [dependencies]
 derive-try-from-primitive = "1.0.0"
 rand = { version = "0.7.2", optional = true }
+serde = { version = "1.0", default-features = false, features = ["derive"] }
```
geigerzaehler, line 25: This should be an optional dependency. Otherwise it is compiled into the runtime.

NunoAlexandre: fixed.

core/src/lib.rs
NunoAlexandre: I don't want to now go on rethinking the usage of `DomainType` in the upstream, that seems appropriate as of now. We can do that in another issue. And the type is used to represent the same model both in the upstream and in the CLI, so I'd rather not duplicate them. Can we agree on moving it to the `client`?

cli/src/command/project.rs
CodeSandwich: At this point it's not about drying anymore, Rust can't do that without additional crates. It's just about keeping all the knowledge concentrated in one place.

`-> &'static[&'static str]`

core/src/lib.rs
CodeSandwich: Then we should switch from using it as `Display` to using it as `Debug`. It's a matter of adding `:?` in a few places. We don't have a proper implementation of `Display` and we shouldn't pretend that we do, there's no value in it.

core/src/lib.rs
geigerzaehler: I feel strongly about not exposing `DomainType` to the upstream. It creates unecessary coupling between the CLI, upstream and the client. Moreover the concerns served by `DomainType` for the CLI and upstream are othorgonal. The Serialize trait is only used by upstream and `FromStr` only used by the CLI where it is also strongly coupled with `possible_values`. Finally upstream is already working without using `DomainType` from the client with very little overhead.

cli/src/command/project.rs
NunoAlexandre: I will undo the moving of this type from the CLI to here so this will be obsolete.

core/src/lib.rs
NunoAlexandre: Why? I honestly don't understand how this is worth our time discussing. We just say here that the Display for ProjectDomain is the same as the Debug implementation. It's done, it's applied everywhere, why do we need to spend more time on this?

core/src/lib.rs
NunoAlexandre: Ok.

core/src/lib.rs
CodeSandwich: Because we don't need it and it introduces more code to maintain, we're also wasting a potentially useful `Display` API and misleading its users.

NunoAlexandre: It's gone.

NunoAlexandre:
@geigerzaehler @CodeSandwich everything is addressed, thanks for the reviews. Could you approve if there is nothing more?

core/Cargo.toml
```diff
@@ -22,6 +23,7 @@ std = [
 [dependencies]
 derive-try-from-primitive = "1.0.0"
 rand = { version = "0.7.2", optional = true }
+serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
```
geigerzaehler, line 26: Since `serde` is now optional we could omit `default-features = false`. But we can also keep it.

NunoAlexandre: Because serde is only used in a context where that flag doesn't add value, right?

PR #514, opened 2020-06-11
Require version in transaction signature

### LINKED ISSUE #471: Re-introduce version check
> Could you please create a follow-up issue to add this check again once we’ve figured out the client compatibility story?
>
> _Originally posted by @geigerzaehler in https://github.com/radicle-dev/radicle-registry/pull/458#issuecomment-633458725_

Part of #435 

In #458, we have dropped the `CheckVersion` signed extension, that guaranteed the client to be connected to a runtime with the same `spec_version` as the runtime it is linked against.

Ultimately, we want to be as backwards and forwards compatible as possible. However, certain changes might require users to update the CLI to keep interacting with the network. In such cases, we want to have a similar version check that validates the CLI client runtime to be compatible to the runtime the client is connected to.

geigerzaehler:
Fixes #471


client/src/transaction.rs
```diff
@@ -98,13 +98,17 @@ fn transaction_extra_to_runtime_extra(
...
     let check_genesis = frame_system::CheckGenesis::new();
     let check_era = frame_system::CheckEra::from(Era::Immortal);
     let check_nonce = frame_system::CheckNonce::from(extra.nonce);
     let check_weight = frame_system::CheckWeight::new();
     let pay_tx_fee = PayTxFee { fee: extra.fee };
 
     let additional_signed = (
+        check_version
+            .additional_signed()
+            .expect("statically returns ok"),
```
CodeSandwich, line 111: What does it mean?

geigerzaehler: I mean that `.additional_signed()` [always returns `Ok()`](https://github.com/paritytech/substrate/blob/0a169d4a6974a457412d8f48f89b56a23e6dcfb2/frame/system/src/lib.rs#L1878-L1880)

CodeSandwich: Then how about
```suggestion
            .expect("statically guaranteed to always return ok"),
```

Issue #496, opened 2020-06-08
Version the state


CodeSandwich:
Every piece of data in the state must be versioned, preferably as an enum. See [this thread](https://radicle.community/t/cli-runtime-compatibiliy/221/2).

geigerzaehler:
An open question is how we want to introduce the versioning without breaking the ffnet. We cannot simply replace the existing state types with the versioned types since the new runtime will be unable to read old unversioned state. We should also clearly document this policy in the developer manual.

CodeSandwich:
I wouldn't mind to make a full chain restart for this. Without versioning we can't introduce versioning without breaking the state. The only way would be to create a new storage and support both of them with a fallback mechanism. In my opinion it's not worth the effort, it's just the ffnet.

geigerzaehler:
We could also just introduce new storage locations for everything we start versioning. For example
~~~diff
diff --git a/runtime/src/registry.rs b/runtime/src/registry.rs
index 3ca9a46..f5f6583 100644
--- a/runtime/src/registry.rs
+++ b/runtime/src/registry.rs
@@ -86 +86 @@ pub mod store {
-            pub Orgs: map hasher(blake2_128_concat) Id => Option<state::Org>;
+            pub OrgsVersioned: map hasher(blake2_128_concat) Id => Option<state::Org>;
~~~
This would make the old storage inaccessible and but we could keep the ffnet running and don’t affect the operation of the blockchain. I’d prefer this.

CodeSandwich:
All the old state would be effectively lost, but we wouldn't need a restart, fine by me :+1:

NunoAlexandre:
> We could also just introduce new storage locations for everything we start versioning. For example
> 
> ```diff
> diff --git a/runtime/src/registry.rs b/runtime/src/registry.rs
> index 3ca9a46..f5f6583 100644
> --- a/runtime/src/registry.rs
> +++ b/runtime/src/registry.rs
> @@ -86 +86 @@ pub mod store {
> -            pub Orgs: map hasher(blake2_128_concat) Id => Option<state::Org>;
> +            pub OrgsVersioned: map hasher(blake2_128_concat) Id => Option<state::Org>;
> ```
> 
> This would make the old storage inaccessible and but we could keep the ffnet running and don’t affect the operation of the blockchain. I’d prefer this.

Can't we have this instead?

``` rust
pub mod VersionedStore {
[...]

PR #515, opened 2020-06-11
Version Orgs state

CodeSandwich:
Part of https://github.com/radicle-dev/radicle-registry/issues/496. ~Blocked by https://github.com/radicle-dev/radicle-registry/pull/514.~

This PR introduces few changes:
- Update state versioning guide
- Drop `Orgs` storage in favor of `OrgsV1`. It allows further versioning of the storage itself by introducing `OrgsV2` (e.g. to change the key type).
- Rename `state::Org` to `state::OrgV1` and introduce versioning enum `state::Org`. `Org`'s API reflects `OrgV1`'s for ease of use. This API is expected to change in the future, e.g. addition of `OrgV2` may or may not alter it, we don't know that yet.
- Move `core::Org` to `client::interface`. This struct is used only in the client's API, so that's a better place for it. It lets the client make decision about what it exposes without touching the runtime. This change is invisible to the client consumers, neither the structure nor its import path changes.

### DIFF
--- a/core/src/state.rs
+++ b/core/src/state.rs
@@ -108,23 +108,83 @@ pub type AccountTransactionIndex = u32;
 
 /// # Storage
 ///
-/// This type is only used for storage. See [crate::Org] for the
-/// complete Org type to be used everywhere else.
-///
-/// Orgs are stored as a map with the key derived from [crate::Org::id].
+/// Orgs are stored as a map with the key derived from [crate::Id].
 /// The org ID can be extracted from the storage key.
-///
-/// # Invariants
-///
-/// * `account_id` is immutable
-/// * `projects` is a set of all the projects owned by the Org.
 ///
 /// # Relevant messages
 ///
 /// * [crate::message::RegisterOrg]
 /// * [crate::message::UnregisterOrg]
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
-pub struct Org {
+pub enum Orgs1Data {
+    V1(OrgV1),
+}
+
+impl Orgs1Data {
+    /// Creates new instance in the most up to date version
+    pub fn new(account_id: AccountId, members: Vec<Id>, projects: Vec<ProjectName>) -> Self {
+        Self::V1(OrgV1 {
+            account_id,
+            members,
+            projects,
+        })
+    }
+
+    /// Account ID that holds the org funds.
+    ///
+    /// It is randomly generated and, unlike for other accounts,
+    /// there is no private key that controls this account.
+    pub fn account_id(&self) -> AccountId {
+        match self {
+            Self::V1(org) => org.account_id,
+        }
+    }
+
+    /// Set of members of the org. Members are allowed to manage
+    /// the org, its projects, and transfer funds.
+    ///
+    /// It is initialized with the user id associated with the author
+    /// of the [crate::message::RegisterOrg] transaction.
+    /// It cannot be changed at the moment.
+    pub fn members(&self) -> &Vec<Id> {
+        match self {
+            Self::V1(org) => &org.members,
+        }
+    }
+
+    /// Set of all projects owned by the org. Members are allowed to register
+    /// a project by sending a [crate::message::RegisterProject] transaction.
+    pub fn projects(&self) -> &Vec<ProjectName> {
+        match self {
+            Self::V1(org) => &org.projects,
+        }
+    }
+
+    /// Add the given project to the list of [Orgs1Data::projects].
+    /// Return a new Org with the new project included or the
+    /// same org if the org already contains that project.
+    pub fn add_project(self, project_name: ProjectName) -> Self {
+        match self {
+            Self::V1(org) => Self::V1(org.add_project(project_name)),
+        }
+    }
+
+    /// Add the given user to the list of [Orgs1Data::members].
+    /// Return a new Org with the new member included or the
+    /// same org if the org already contains that member.
+    pub fn add_member(self, user_id: Id) -> Self {
+        match self {
+            Self::V1(org) => Self::V1(org.add_member(user_id)),
+        }
+    }
+}
+
+/// # Invariants
+///
+/// * `account_id` is immutable
+/// * `projects` is a set of all the projects owned by the Org.
+#[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
+pub struct OrgV1 {
     /// Account ID that holds the org funds.
     ///
     /// It is randomly generated and, unlike for other accounts,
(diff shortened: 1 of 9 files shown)


core/src/state.rs
```diff
@@ -108,23 +108,103 @@ pub type AccountTransactionIndex = u32;
 
 /// # Storage
 ///
-/// This type is only used for storage. See [crate::Org] for the
-/// complete Org type to be used everywhere else.
+/// This type is only used for storage.
```
geigerzaehler, line 111: I think we can delete this line.

core/src/state.rs
```diff
@@ -108,23 +108,103 @@ pub type AccountTransactionIndex = u32;
...
-pub struct Org {
+pub enum Org {
+    V1(OrgV1),
+}
+
+impl Org {
+    /// Creates new instance in the most up to date version
+    ///
+    /// # Stability warning
+    /// This function will most likely change when new versions of this type are added
```
geigerzaehler, line 129: Do we need this? Who is this for?

runtime/src/registry.rs
```diff
@@ -94,7 +94,7 @@ pub mod store {
             // The storage for Orgs, indexed by Id.
             // We use the blake2_128_concat hasher so that the Id
             // can be extracted from the key.
-            pub Orgs: map hasher(blake2_128_concat) Id => Option<state::Org>;
+            pub OrgsV1: map hasher(blake2_128_concat) Id => Option<state::Org>;
```
geigerzaehler, line 97: This is a bit confusing. We also have `state::OrgsV1` and do we intend to have `store::OrgsV2` at some point? I’d opt for `OrgsVersioned`.

core/src/state.rs
CodeSandwich: This is for the future developers. It's a reminder not to rely on this API too heavily with code structure and not to be afraid to break it. I think that it has a little bit of value, I'm not very attached to it.

runtime/src/registry.rs
CodeSandwich: I've explained it in the PR description. I want to leave an open door to version the storage itself, e.g. to be able to change the key type or the data structure altogether.

core/src/state.rs
NunoAlexandre: I had the same thought as Thomas here but I think it's good to keep tbh.

NunoAlexandre:
> Drop Orgs storage in favor of OrgsV1. It allows further versioning of the storage itself by introducing OrgsV2 (e.g. to change the key type).

This is smart. However, it confuses me a little since now we have `OrgsV1` that regards the `storage` and `OrgV1` that regards a variant of data it stores. Maybe it's a matter of getting used to it but I see it getting in the way.

runtime/src/registry.rs
NunoAlexandre: https://github.com/radicle-dev/radicle-registry/pull/515#issuecomment-642722303

core/src/state.rs
CodeSandwich: Fixed

runtime/src/registry.rs
geigerzaehler: If we want to go down that route we should be consistent and use this approach everywhere. Could you please update the relevant section in the developer docs?

core/src/state.rs
geigerzaehler: This raises a couple of questions: Are you talking about registry devs or client users?  What does it mean to no rely on this API too heavily? Do we put this warning on all method of state objects? I’m worried about this not having any well defined meaning. This would lead to inconsistencies in applying this warning which would lead to more confusion for people reading this.

For now I think we should omit this. We can re-introduce this once we’ve figured all this out.

CodeSandwich: I wrote it thinking about client library developers. They should not base their APIs on this type, they should use either specific version of `Org` or their custom wrapper to improve chance of being able to hide the introduction of a new `Org` variant.

geigerzaehler: I wouldn’t prescribe how we expose things on the client just yet. We’re breaking APIs left and right and will do for some time while the runtime keeps evolving. (I’m also leaning more towards the client exposing the state transparently. This way the client does not have to decide how to handle old state and that decision can be made by the consumers giving them full control)

CodeSandwich: > I’m also leaning more towards the client exposing the state transparently. 

That's a good strategy, it could be practical to expose `enum Org` directly. The only worry I have is that somewhere deep in upstream somebody will trustingly start using these unstable methods only to find out that the whole project needs deep changes whenever newer version of client lib is released. Maybe we can wrap these unstable methods in a new trait `OrgUnstable`? This way anybody can use or expose raw `Org` with relative future proofness, but when somebody needs the unstable helper tooling, he can explicitly import it.

geigerzaehler: >  The only worry I have is that somewhere deep in upstream somebody will trustingly start using these unstable methods only to find out that the whole project needs deep changes whenever newer version of client lib is released.

I’m not really concerned about this. We are breaking the API constantly in all places and the state is not something special in that regard. Upstream is also in a state where they are still very flexible. It does seem strange to have this stability warning on one piece of the API.

CodeSandwich: Ok, let's leave ensuring stability for the future. I'll remove this comment.

CodeSandwich: Fixed

runtime/src/registry.rs
CodeSandwich: Fixed

CodeSandwich:
I've renamed `state::Org` to `state::OrgsV1Data`. I've figured out that if we want to have `store::Orgs` versioned, its data needs to be versioned as well.

geigerzaehler:
> Now we have `OrgsV1 -> OrgsV1Data::V1 -> OrgV1`

I find this even more confusing to be honest. `OrgsVX` and `OrgVX` are hard to distinguish and the relationship between is counterintuitive. Plus `OrgsV1Data` now gets its name from its location as opposed to what it contains.

I’d suggest we punt on a decision about a versioning scheme for the storage keys. We don’t need to solve this now because we don’t need to prepare for backwards compatibility and we can always use an ad-hoc name. So for now I’d suggest to use `OrgsVersioned`.

@CodeSandwich If you believe we should tackle this now then @radicle-dev/registry should have a chat and explore the options a bit.

CodeSandwich:
Just for the record: we've settled on:
```rust
store::Orgs1 -> store::Orgs1Data::V1 -> state::OrgV1
store::Orgs1 -> store::Orgs1Data::V2 -> state::OrgV2
store::Orgs2 -> store::Orgs2Data::V3 -> state::OrgV3
```

PR #517, opened 2020-06-12
client: Define struct Hash

### LINKED ISSUE #512: Consider defining a Hash struct
For https://github.com/radicle-dev/radicle-upstream/issues/464, we'd need to move the upstream definition of `Hash` to the registry. Doing that would result in a clash with the `Hash` type alias definition. 

In order to provide the type and traits needed by the upstream, we would not only need to move the type definition and its trait implementations to the registry but to actually use it in here instead of the alias, which is a deeper change that affects the runtime.

I'm all in for further type-safety and control, so I'm in for it.

NunoAlexandre:
Closes #512 

`Draft`

For now, it only offers `struct Hash` at the client level. The struct definition should be moved to client :: interface.


NunoAlexandre:
We concluded that introducing this change adds substantial overhead with minimal added value, so we are better off dropping it.

Issue #492, opened 2020-06-08
Build runtime Wasm artifacts on CI


geigerzaehler:
We want to build and upload the Wasm artifact of the runtime as part of the CI. This would allow us to retrieve a specific version of the runtime for runtime compatibility tests (discussed [here](https://radicle.community/t/runtime-updates-workflow-and-testing/224).

In addition, this would replace the need for checking the latest runtime into `runtime/latest.wasm`. Having the artifact in code requires developers to build it locally and to follow a specific policy. The policy cannot be enforced and might be violated. 

If we have the Wasm runtime as an artifact we should also update the chain update policy and specify that a build artifact is used instead of `runtime/latest.wasm`.

### Open questions
Figure out how to address the artifacts. What we currently can do is retrieve the Wasm runtime for a specific commit hash. We may need to get the runtime for a specific runtime spec and implementation version or a git tag.

At the moment `runtime/latest.wasm` is used as the genesis runtime for the dev and devnet chains. We need to figure out what genesis runtime we use for the dev and devnet chain and how the node can be build.

`runtime/latest.wasm` needs to be build as part of the development process so that developers can run the node. We need to adjust `./scripts/build-dev-node`.

CodeSandwich:
Design proposal, I hope that I've covered all the quirks:

1. Create a Google Cloud bucket. It should be publicly readable, but writing should require credentials.
2. The bucket contains files with a strict naming convention
    3. The runtime file for spec 3 impl 2 is named `3_2.wasm`.
    1. The spec file for the ffnet is named `ffnet.json`. This file is managed manually.
    1. For the dev chain there are always 2 spec files per spec version:
        1. For the first impl, e.g. for spec 3 `dev_3_first.json`
        1. For the last impl, e.g. for spec 3 `dev_3_last.json`
3. Whenever the CI is running, it fetches necessary spec files from GC.
    1. If the latest version of the runtime file is available, it's downloaded, otherwise it's locally built.
    1. The spec file for the last impl is always built locally. In case of impl version 0, the spec file for the first impl is built too. 
    2. If an old version of the spec file is not available, the CI aborts. (It can't be built locally)
4. If the CI is running on a `master` branch (`$BUILDKITE_BRANCH`), it will upload files to GC.
    1. The GC credentials are available only for master branch builds to prevent theft.
    2. If a runtime for the current version had to be built locally, it's uploaded to GC.
[...]

PR #523, opened 2020-06-16
Build wasm runtime on CI and upload it as artifact

geigerzaehler:
As preparation for #492 we start building the wasm runtime on CI and uploading it as an artifact.

### DIFF
--- a/ci/run
+++ b/ci/run
@@ -58,6 +58,14 @@ echo "--- cargo doc"
 RUSTDOCFLAGS="-D intra-doc-link-resolution-failure" \
   cargo doc --workspace --release --no-deps --document-private-items
 
+# For non-master branch builds we rely on the file that is checked into
+# the repo. We will enable this for all builds once this file is removed.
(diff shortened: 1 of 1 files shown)


ci/run
```diff
@@ -43,6 +43,8 @@ time cargo fmt --all -- --check
 
 export RUSTFLAGS="-D warnings"
 
+touch runtime/latest.wasm
```
NunoAlexandre, line 46: is this necessary to bypass some caching?

geigerzaehler: Good catch! This is just an artifact from a previous experiment. Will remove it.

geigerzaehler:
@NunoAlexandre: could you have another look?

geigerzaehler:
> Is this useful in feature branches? Do we have a use case for those cases? If not, I'd prefer to only run this on the master branch to avoid increasing the build time.

It will become necessary to run this on every build once we remove `runtime/latest.wasm` from the repo (see #492). But at the moment this is not necessary and serves merely as a proof-of-concept. I will limit this to master branch builds.

PR #524, opened 2020-06-16
Expose the SecretStringError type as CryptoError

FintanH:
So that downstream users of the trait can also refer to the error.

Basically, I need access to the trait for the key storage, but I have no way of referring to the error unless I also add sp-core as a dep to proxy.

### DIFF
--- a/client/src/interface.rs
+++ b/client/src/interface.rs
@@ -24,7 +24,9 @@ pub use radicle_registry_core::*;
 pub use radicle_registry_runtime::{BlockNumber, Hash, Header, RuntimeVersion};
 
 pub use radicle_registry_runtime::{registry::Event as RegistryEvent, Balance, Event};
-pub use sp_core::crypto::{Pair as CryptoPair, Public as CryptoPublic};
+pub use sp_core::crypto::{
+    Pair as CryptoPair, Public as CryptoPublic, SecretStringError as CryptoError,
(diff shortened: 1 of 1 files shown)


FintanH:
omfg it doesnt implement display....

NunoAlexandre:
> omfg it doesnt implement display....

So this PR doesn't fix the issue for you? :disappointed_relieved:

FintanH:
> So this PR doesn't fix the issue for you?

It partly does. I'll just have newtype wrap it up.

FintanH:
> could you have your commit message follow the 50-char max, imperative-style convention as defined in the developing guide?

Done :+1:

NunoAlexandre:
@FintanH since the alias is only useful in your use case in the downstream, I'd prefer if we don't rename it here but only expose it and have you, when importing it, aliasing it to your preference. Does that sound good to you?

FintanH:
Sure thing, I was just trying to make it consistent :)

NunoAlexandre:
> Sure thing, I was just trying to make it consistent :)

Ah, good point, I overlooked that. It's good as is :+1:

FintanH:
Too late

NunoAlexandre:
> Too late

Undo please :laughing:

FintanH:
Killing me

Issue #497, opened 2020-06-08
CI tests client backward compatibility


CodeSandwich:
The CI should be running the e2e test against the current runtime and previous versions to ensure backward compatibility of the client with at least 1 old spec version of the runtime. See [this thread](https://radicle.community/t/runtime-updates-workflow-and-testing/224). We need to run against:
- the last implementation of the previous spec version runtime
- the previous implementation of the current spec version runtime
- the current implementation of the current spec version runtime

This needs to be implemented as a script, which can be run both locally by developers and by the CI.

Until https://github.com/radicle-dev/radicle-registry/issues/492 is in place it's fine to just store the WASM binaries needed for the CI in the repository as regular binary files.

We now we can use checked-in WASM blobs with the needed runtimes. The process of updating the runtimes needs to be documented.

geigerzaehler:
As part of this we should clearly document in the updates policy that we keep tests for the previous spec version that we are testing around to ensure that the new client works in that case, too.

geigerzaehler:
To make this effective we need to run a node with different runtimes without having to recompile the node. I propose we add a `--runtime` option that allows one to point to a Wasm file which is then used as the initial runtime for the dev chain. The option would only work for the `dev` and `local-devnet` chains.

The rest of the initial chain state would be provided by the chain spec for the selected chain. Otherwise we would need to store the initial chain state along side the runtime in the repo.

geigerzaehler:
@CodeSandwich: Is there something left to be done that wasn’t tackled in #536? I’m wondering why you reopend this?

CodeSandwich:
We still have the genesis creation issue: https://github.com/radicle-dev/radicle-registry/pull/536#issuecomment-648639519 Unless we resolve it, the CI backward compatibility tests are based on coincidence.

PR #529, opened 2020-06-18
Add node option to specify genesis runtime

geigerzaehler:
This will allow us to run tests against different runtimes. See #497

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -13,16 +13,12 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Provides [Chain] and [ChainSpec]s for various chains we want to run.
-//!
-//! Available chain specs
-//! * [dev] for runnning a single node locally and develop against it.
-//! * [local_devnet] for runnning a cluster of three nodes locally.
+//! Provides constructor functions to create [ChainSpec]s.
 use crate::pow::config::Config as PowAlgConfig;
 use radicle_registry_runtime::{
-    AccountId, BalancesConfig, GenesisConfig, SudoConfig, SystemConfig,
+    AccountId, Balance, BalancesConfig, GenesisConfig, SudoConfig, SystemConfig,
 };
-use sc_service::{ChainType, GenericChainSpec};
+use sc_service::{config::MultiaddrWithPeerId, ChainType, GenericChainSpec};
 use sp_core::{crypto::CryptoType, Pair};
 use std::convert::TryFrom;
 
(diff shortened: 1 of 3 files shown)


node/src/chain_spec.rs
```diff
@@ -13,11 +13,13 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Provides [Chain] and [ChainSpec]s for various chains we want to run.
+//! Provides constructor functions to create chain specs.
 //!
 //! Available chain specs
 //! * [dev] for runnning a single node locally and develop against it.
 //! * [local_devnet] for runnning a cluster of three nodes locally.
+//! * [devnet] for runnning a cluster of three nodes locally.
```
NunoAlexandre, line 21: `[local_devnet]` and `[devnet]` have the exact same description.

node/src/chain_spec.rs
```diff
@@ -32,6 +34,9 @@ pub type ChainSpec = GenericChainSpec<GenesisConfig>;
 const FFNET_CHAIN_SPEC: &[u8] = include_bytes!("./chain_spec/ffnet.json");
 const LATEST_RUNTIME_WASM: &[u8] = include_bytes!("../../runtime/latest.wasm");
 
+/// Chain for local development with an isolated node.
```
NunoAlexandre, line 37: To keep the terminology consistent with the above:

```suggestion
/// Chain for local development with a single node.
```

node/src/chain_spec.rs
```diff
@@ -32,6 +34,9 @@ pub type ChainSpec = GenericChainSpec<GenesisConfig>;
...
+///
+/// If `runtime` is given it is used as the genesis runtime. Uses dummy PoW that does not mine.
```
NunoAlexandre, line 39:
```suggestion
/// If `runtime` is given, it is used as the genesis runtime. Uses dummy PoW that does not mine.
```

One question here. If we are operating on a single node and it doesn't mine, wouldn't that mean that txs are never really included in a block?

node/src/chain_spec.rs
```diff
@@ -45,7 +50,9 @@ pub fn dev(runtime: Option<Vec<u8>>) -> ChainSpec {
     .into_chain_spec()
 }
 
-pub fn devnet(runtime: Option<Vec<u8>>) -> ChainSpec {
+/// Chain that is running on the cloud, frequently reset, and might not work but always uses the
+/// latest code.
```
NunoAlexandre, line 54: I would like to be explicit here. I read it and wondered if it was referring to the native code.

```suggestion
/// latest wasm code.
```

node/src/chain_spec.rs
```diff
@@ -56,13 +63,17 @@ pub fn devnet(runtime: Option<Vec<u8>>) -> ChainSpec {
...
         balances: dev_balances(),
         sudo_key: account_id("Alice"),
     }
     .into_chain_spec()
 }
 
+/// Chain for running a cluster of nodes locally.
+///
+/// If `runtime` is given it is used as the genesis runtime. Similar to [dev] but uses proper PoW
+/// consensus.
```
NunoAlexandre, line 76:
```suggestion
/// If `runtime` is given, it is used as the genesis runtime. 
/// Similar to [dev] but uses proper PoW consensus.
```

node/src/chain_spec.rs
```diff
@@ -13,16 +13,18 @@
...
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! Provides [Chain] and [ChainSpec]s for various chains we want to run.
+//! Provides constructor functions to create chain specs.
 //!
 //! Available chain specs
 //! * [dev] for runnning a single node locally and develop against it.
 //! * [local_devnet] for runnning a cluster of three nodes locally.
+//! * [devnet] for runnning a cluster of three nodes locally.
+//! * [ffnet] for runnning a cluster of three nodes locally.
```
NunoAlexandre, line 22:
```
//! * [ffnet] for runnning a cluster of three nodes locally.
```

This doesn't seem right

node/src/chain_spec.rs
```diff
@@ -32,98 +34,119 @@ pub type ChainSpec = GenericChainSpec<GenesisConfig>;
...
-fn devnet() -> ChainSpec {
-    GenericChainSpec::from_genesis(
-        "Radicle Registry devnet",
-        "devnet",
-        ChainType::Custom("devnet".to_string()),
-        dev_genesis_config,
-        // boot nodes
-        // From key 000...001
-        vec![
+/// Chain that is running on the cloud, frequently reset, and might not work but always uses the
```
NunoAlexandre, line 53: Shall we be more explicit on what "might not work" means?

node/src/chain_spec.rs
```diff
@@ -32,98 +34,119 @@ pub type ChainSpec = GenericChainSpec<GenesisConfig>;
...
+/// latest code.
+pub fn devnet() -> ChainSpec {
+    ChainParams {
+        id: String::from("devnet"),
+        chain_type: ChainType::Development,
+        boot_nodes: vec![
+            // From key 000...001
```
NunoAlexandre, line 60: Not following this comment? :thinking:

node/src/cli.rs
```diff
@@ -126,8 +133,24 @@ impl SubstrateCli for Cli {
...
-        let chain_spec = parse_chain(id)?.spec();
-        Ok(Box::new(chain_spec) as Box<dyn ChainSpec>)
+        let runtime = if let Some(runtime_path) = &self.runtime {
+            Some(
+                std::fs::read(runtime_path)
+                    .map_err(|e| format!("Error reading file {:?}: {}", runtime_path, e))?,
+            )
+        } else {
+            None
+        };
```
NunoAlexandre, line 143: This is difficult to read, lots of nesting here.

```
let runtime = &self
  .runtime
  .map(std::fs::read)
  .map_err(|e| format!("Error reading file {:?}: {}", runtime_path, e))?
```

node/src/cli.rs
```diff
@@ -126,8 +133,24 @@ impl SubstrateCli for Cli {
...
+
+        let chain_spec = match id {
+            "dev" => Ok(crate::chain_spec::dev(runtime)),
+            "local-devnet" => Ok(crate::chain_spec::local_devnet(runtime)),
+            "devnet" => Ok(crate::chain_spec::devnet()),
+            "ffnet" => Ok(crate::chain_spec::ffnet()),
+            _ => Err(format!("Invalid chain {}", id)),
+        }?;
```
NunoAlexandre, line 151: I usually prefer to have these in a separate function just to reduce the load of reading the outer function. I.e, now to understand `load_spec` the reader also _has_ to read to how each chain spec is resolved. Having it aside reduces that load. Just a suggestion.

node/src/chain_spec.rs
geigerzaehler: I’ve used the wrong word “mine” here. I’ve updated the comment to clarify it.

node/src/chain_spec.rs
geigerzaehler: This is going away in favor of the docs on functions.

node/src/chain_spec.rs
geigerzaehler: This is going away. See above.

node/src/cli.rs
geigerzaehler: Unfortunately your suggestion does not compile without the use of [`transpose`](https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose). I’d like to avoid using that method though because it’s hard to wrap your head around.

NunoAlexandre: It would make the code much more terse though, and - imo - easier to follow and read. As it is now is difficult to wrap _my_head around :laughing:

node/src/cli.rs
geigerzaehler: It think you need to understand how the chains are resolved and what functions are called in order to understand `load_spec`. I’d like to keep it like this.

node/src/cli.rs
geigerzaehler: Because of variable scope and borrowing this is the alternative code
~~~rust
let runtime = self
    .runtime
    .as_ref()
    .map(|runtime_path| {
        std::fs::read(runtime_path)
            .map_err(|e| format!("Error reading file {:?}: {}", runtime_path, e))
    })
    .transpose()?;
~~~
It’s not much terser. The only difference is that it replaces pattern matching with `transpose` and `&self.runtime` with `runtime.as_ref()`. Since `transpose` is quite obscure I’d like to go with the more explicit version. I’ve adjusted the code slightly to make it more readable.

node/src/chain_spec.rs
```diff
@@ -32,98 +28,119 @@ pub type ChainSpec = GenericChainSpec<GenesisConfig>;
...
+        runtime: runtime.unwrap_or_else(|| LATEST_RUNTIME_WASM.to_owned()),
+        balances: dev_balances(),
+        sudo_key: account_id("Alice"),
     }
+    .into_chain_spec()
 }
 
-fn dev() -> ChainSpec {
-    GenericChainSpec::from_genesis(
-        "Radicle Registry isolated development",
```
CodeSandwich, line 59: Did we loose a verbose node name?

node/src/pow/config.rs
```diff
@@ -17,7 +17,7 @@ use sc_service::{config::Configuration, Properties};
 use std::convert::{TryFrom, TryInto};
 
 /// Configuration of PoW algorithm, can be stored as chain spec property
-#[derive(serde::Deserialize, serde::Serialize)]
+#[derive(Clone, Copy, Debug, serde::Deserialize, serde::Serialize)]
```
CodeSandwich, line 20: Do we need `Copy`?

geigerzaehler: Nope. Will fix.

node/src/chain_spec.rs
geigerzaehler: Yes. This was used in the node output on startup. However, the node did not print the actual chain name so it was not super useful. Moreover the description doesn’t really add that much value so I decided to drop it.

node/src/chain_spec.rs
geigerzaehler: Fixed

node/src/chain_spec.rs
geigerzaehler: Fixed

PR #534, opened 2020-06-22
Offer dev mode for node with mem db

geigerzaehler:
We offer the `--dev` flag to run the node in development mode. This eliminates the need for manually resetting the chain. This is especially useful when we will run tests against multiple runtimes (#497).


node/src/cli.rs
```diff
@@ -100,6 +101,10 @@ pub struct Cli {
     /// has an effect for development chains and if the block database is empty.
     #[structopt(long)]
     runtime: Option<PathBuf>,
+
+    /// Run the dev chain with an in-memory database and mining
+    #[structopt(long, conflicts_with = "chain")]
+    dev: bool,
```
CodeSandwich, line 107: Would reusing `--chain dev` be a viable solution? I can't think of a scenario where I would run the `dev` chain without the `--dev` flag. But that would be a lot of implicit behavior.

node/src/cli.rs
```diff
@@ -213,6 +234,11 @@ impl Cli {
             other: execution_strategy,
         };
 
+        if self.dev {
+            let db: Box<dyn sp_database::Database<[u8; 32]>> = Box::new(sp_database::MemDb::new());
+            config.database = sc_service::config::DatabaseConfig::Custom(Arc::from(db));
```
CodeSandwich, line 239:
```suggestion
            let db = Arc::new(sp_database::MemDb::new());
            config.database = sc_service::config::DatabaseConfig::Custom(db);
```

ci/run
```diff
@@ -71,11 +71,12 @@ echo "--- scripts/build-release"
...
 echo "Starting radicle-registry-node"
-RUST_LOG=error ./target/release/radicle-registry-node \
-  --chain dev \
-  --mine 5HYpUCg4KKiwpih63PUHmGeNrK2XeTxKR83yNKbZeTsvSKNq \
-  --data-path /tmp/radicle-registry \
-  &
+
+# The home directory is not writable on CI
+export XDG_DATA_HOME=/tmp/xdg-data-home
+export XDG_CONFIG_HOME=/tmp/xdg-config-home
```
CodeSandwich, line 77: Why is this needed? I haven't found a single use of `XDG_DATA_HOME` in the whole substrate codebase. Wasn't `MemDb` supposed to leave nothing on the HDD? Won't this reintroduce the persistence problem?

node/src/cli.rs
geigerzaehler: Unfortunately rustc cannot infer the types properly with syntax. I’ll add a comment that explains this.

ci/run
geigerzaehler: Substrate uses the `app_dirs` crate on old versions and the `directories` crate on new versions. (See https://github.com/paritytech/substrate/commit/ffa32fa60767838b4ec779fb6ac42e7516600a7d). This code creates the directories recursively. Before we worked around this with the `--data` option.

node/src/cli.rs
CodeSandwich: I've run it before submitting and rustc was able to infer this just fine!

node/src/cli.rs
geigerzaehler: >  I can't think of a scenario where I would run the `dev` chain without the `--dev` flag or the `--dev` flag on a non-`dev` chain.

I find this unlikely, too.

> Would reusing `--chain dev` be a viable solution? [...] that would be a lot of implicit behavior.

The implicit behavior was my concern, too. I think it might be confusing (and also tricky to document) that using `--chain dev` has an effect on other parts of the node and that it is special when compared to the other chain specs. I also find it easier to explain in the developer docs to just use the `--dev` flag.

node/src/cli.rs
geigerzaehler: You’re right. Maybe I had something slightly different :thinking:

node/src/cli.rs
```diff
@@ -236,8 +236,8 @@ impl Cli {
 
         if self.dev {
             // We need to help the type checker with explicit annotations.
```
CodeSandwich, line 238: Now it's not needed anymore :)

PR #536, opened 2020-06-22
Test multiple runtimes on the CI

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/497


client/tests/end_to_end.rs
```diff
@@ -330,3 +330,180 @@ async fn insufficient_funds() {
...
+    assert_eq!(tx_included_once.result, Ok(()));
+
+    let tx_included_twice = submit_ok(&client, &author, register_org_message.clone()).await;
+    assert_eq!(
+        tx_included_twice.result,
+        Err(RegistryError::IdAlreadyTaken.into())
+    );
+
+    // Unregistration.
+    let fee = random_balance();
```
geigerzaehler, line 364: As I understand it the idea behind unregistering everything is to prevent state that is left on chain from interfering with other tests.

There are a couple of things to consider:

* We don’t clean up state consistently. Other tests don’t do it and may still interfere.
* We don’t clean up all the state. For example account balances remain.
* This is only an additional safe guard against interference from state. There’s no evidence that this is necessary. As far as we know it is enough to generate all the IDs randomly.
* The cleanup requires significant, non-trivial effort.

Given all this I would suggest not doing the cleanup. Instead we should focus on writing the tests so that they cannot interfere with each other.

NunoAlexandre: > * As far as we know it is enough to generate all the IDs randomly.

This.

CodeSandwich: > As far as we know it is enough to generate all the IDs randomly.
Right, no flimsy cleanup is needed :facepalm: Fixed.

CodeSandwich:
> we don’t actually run runtime/latest.wasm

> We should also adjust the execution strategy

One more thing is that the genesis is generated using the native runtime no matter what WASM we feed it. We were lucky, that the block format hasn't changed, but we need the WASM runtimes wrapped in JSON specs.

geigerzaehler:
> One more thing is that the genesis is generated using the native runtime no matter what WASM we feed it. We were lucky, that the block format hasn't changed, but we need the WASM runtimes wrapped in JSON specs.

I think you’re right. But I’m ok with merging this as is for now.

PR #537, opened 2020-06-22
Make runtime errors backwards compatible

geigerzaehler:
To make runtime errors compatible between runtime version 9 and 10 we restore the deleted errors and add the new error at the end. Otherwise the compiled with version 10 would be confused about the error.


CodeSandwich:
Maybe we should introduce hardcoded representation right away?

core/src/error.rs
```diff
@@ -50,31 +53,33 @@ pub enum RegistryError {
...
     #[cfg_attr(feature = "std", error("the user is already a member of the org"))]
-    AlreadyAMember,
+    AlreadyAMember = 8,
 
     #[cfg_attr(feature = "std", error("the provided fee is insufficient"))]
-    InsufficientFee,
+    InsufficientFee = 9,
 
     #[cfg_attr(feature = "std", error("the sender is not a project member"))]
     InsufficientSenderPermissions,
```
CodeSandwich, line 85: The discriminant is missing :grimacing:

PR #541, opened 2020-06-23
runtime: tx author pays for UnregisterOrg

### LINKED ISSUE #539: Tx author pays for UnregisterOrg
As reported by @CodeSandwich, now a tx author pays for `RegisterOrg` but to `UnregisterOrg` funds must first be transferred to the involved org so that said org can pay for that tx. This is odd from a user experience perspective.

Instead, have the tx author paying the tx fee associated with `UnregisterOrg`.

NunoAlexandre:
Closes #539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Upcoming
 
 ### Breaking changes
 
+* runtime: Tx author pays for `UnregisterOrg`
 * client: Offline transaction signing now requires users to manually specify the
   runtime spec version the transaction is valid for.
 * client: Renamed `ClientT::onchain_runtime_version` to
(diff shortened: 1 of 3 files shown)


geigerzaehler:
@NunoAlexandre: You forgot to recompile `runtime/latest.wasm`. Unfortunately that tests don’t catch that at the moment. We should also bump the spec version.

NunoAlexandre:
> @NunoAlexandre: You forgot to recompile `runtime/latest.wasm`. Unfortunately that tests don’t catch that at the moment. We should also bump the spec version.



> @NunoAlexandre: You forgot to recompile `runtime/latest.wasm`. Unfortunately that tests don’t catch that at the moment. We should also bump the spec version.

Followed up with https://github.com/radicle-dev/radicle-registry/pull/542

PR #543, opened 2020-06-24
Simplify `Message::Result`

geigerzaehler:
We replace the associated `Message::Result` type with `Message::Output`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Upcoming
 
 ### Breaking changes
 
+* client: Replaced the associated `Message::Result` type with `Message::Output`.
+  This should require no changes on the user’s side since
+  `TransactionIncluded::result` remains structurally equal.
 * runtime: Tx author pays for `UnregisterOrg`
 * client: Remove `TransactionIncluded::events` field.
 * client: Offline transaction signing now requires users to manually specify the
--- a/client/src/message.rs
+++ b/client/src/message.rs
@@ -28,22 +28,29 @@ type EventParseError = String;
 /// For every [RuntimeCall] that is exposed to the user we implement [Message] for the parameters
 /// struct of the runtime message.
 pub trait Message: Send + 'static {
-    /// Result of executing the message in the runtime that is presented to the client user.
-    type Result: Send + 'static;
+    /// Output of a successfully applied message.
+    ///
+    /// This value is extracted from the events that are dispatched when the message is executed in
+    /// a block.
+    type Output: Send + 'static;
 
     /// Parse all runtime events emitted by the message and return the appropriate message result.
     ///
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ Upcoming
 
 * runtime: Tx author pays for `UnregisterOrg`
 * client: Remove `TransactionIncluded::events` field.
+* client: Replaced the associated `Message::Result` type with `Message::Output`.
```
NunoAlexandre, line 11: This is misplaced, should go on top

client/src/message.rs
```diff
@@ -28,22 +28,28 @@ type EventParseError = String;
 /// For every [RuntimeCall] that is exposed to the user we implement [Message] for the parameters
 /// struct of the runtime message.
 pub trait Message: Send + 'static {
-    /// Result of executing the message in the runtime that is presented to the client user.
-    type Result: Send + 'static;
+    /// Output of a successfully applied message.
```
NunoAlexandre, line 31: we should use `included` instead of applied

client/src/message.rs
```diff
@@ -28,22 +28,28 @@ type EventParseError = String;
...
+    ///
+    /// This value is extracted from the events created by the message.
```
NunoAlexandre, line 33: `nit` the message itself doesn't create any events, it's the runtime/registry that issues events when handling the message

geigerzaehler:
> Maybe we could expose `type Result<T> = std::result::Result<T, TransactionError>` and use it whenever applicable?

I’m on the fence about adding type aliases because of the indirection. What’s you thought on this @NunoAlexandre? Alternatively we could use `type R<M> = Result<<M as Message>::Output, TransactionError>` which would look better on the call site.

NunoAlexandre:
Hm good question. I don't like the indirection that aliases create. If you guys are into it, I don't oppose. But if it's not a clear vote than I'd leave it as is and see if it becomes annoying in practice to justify the alias.

client/src/message.rs
geigerzaehler: The terminology is a bit tricky here. If a transaction is successfully included in a block it still may fail to execute properly and return a `TransactionError` instead of `Output`. I’m open to suggestions that make this clearer.

geigerzaehler:
> Hm good question. I don't like the indirection that aliases create. If you guys are into it, I don't oppose. But if it's not a clear vote than I'd leave it as is and see if it becomes annoying in practice to justify the alias.

I second this especially since the use of the result is very localized. I’ll keep it as is for the moment.

client/src/message.rs
CodeSandwich: I'm ok with the current description. It's clear, yet broad.

NunoAlexandre: we could add this comment to the docstring?

PR #545, opened 2020-06-24
Decouple client and substrate runtime modules

geigerzaehler:
The main goal of this change is to decouple the client from the third-party substrate modules. Specifically we want the client to only depend on the runtime crate and not to depend on dependencies of the runtime crate. The motivation behind this is to allow the client to link against multiple versions of the runtime crate.

A secondary objective that this change achieves is to be more disciplined in the public API of the runtime crate. We want to reduce
the number of things that are exposed. To this end we move the runtime construction that declares multiple public types and structs to a separate module from which we re-export selectively.


client/src/message.rs
```diff
@@ -66,15 +66,15 @@ impl Message for message::RegisterMember {
...
+                Event::registry(event::Registry::MemberRegistered(_, _)) => Some(Ok(())),
                 _ => None,
             }),
             Err(dispatch_error) => Ok(Err(dispatch_error)),
         }
     }
 
     fn into_runtime_call(self) -> RuntimeCall {
-        registry::Call::register_member(self).into()
+        call::Registry::register_member(self).into()
```
NunoAlexandre, line 77: :ping_pong:

client/src/message.rs
```diff
@@ -85,15 +85,15 @@ impl Message for message::RegisterOrg {
         let dispatch_result = get_dispatch_result(&events)?;
         match dispatch_result {
             Ok(()) => find_event(&events, "OrgRegistered", |event| match event {
-                Event::registry(registry::Event::OrgRegistered(_)) => Some(Ok(())),
+                Event::registry(event::Registry::OrgRegistered(_)) => Some(Ok(())),
```
NunoAlexandre, line 88: :ping_pong:

PR #546, opened 2020-06-25
Add transaction_index to runtime API

geigerzaehler:
We again reduce the coupling of the client to `frame_system` by internalizing access to the transaction index of an event.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -55,8 +55,6 @@ pub type Signature = ed25519::Signature;
 /// Same as  [sp_runtime::traits::Hash::Output] for [Hashing].
 pub type Hash = sp_core::H256;
 
-pub type EventRecord = frame_system::EventRecord<runtime::Event, Hash>;
-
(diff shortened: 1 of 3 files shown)


runtime/src/lib.rs
```diff
@@ -118,8 +116,21 @@ pub mod store {
...
 pub mod event {
+    pub use crate::runtime::Event;
+    pub type Record = frame_system::EventRecord<crate::runtime::Event, crate::Hash>;
     pub type Registry = crate::registry::Event;
     pub type System = frame_system::Event<crate::Runtime>;
+
+    /// Return the index of the transaction in the block that dispatched the event.
+    ///
+    /// Returns `None` if the event was not dispatched as part of a transaction.
+    #[cfg(feature = "std")]
```
CodeSandwich, line 127: Does it need to be feature gated? Why?

geigerzaehler: The function is not used when the runtime is executed on chain. To avoid generating the code when we compile the Wasm I added this.

PR #548, opened 2020-06-25
Switch CI tests from juggling runtime WASMs to spec JSONs

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/497.

The update flow is this:
1. change the runtime
1. run `update-chain-spec-new-impl` or `update-chain-spec-new-spec`
1. this updates `latest.wasm` based on the current runtime
1. the current runtime is now identical on both native and WASM, which are then used to safely generate the spec JSON

There's no need to have `--runtime` flag anymore, manual feeding of the runtime WASM is not needed in any use case:
- the developers use the `--dev` or `--chain <CHAIN>` flag, which runs the native runtime to build a genesis and then uses the same native runtime in the node
- the CI uses the `--dev --spec <SPEC_FILE>`, which has a prebuilt genesis and uses the prebuilt WASM runtime in the node
- the end users use `--chain ffnet`, which has a prebuilt genesis and uses both the prebuilt WASM and the native runtimes in the node

This PR has a fake `runtime/dev_prev_spec_last_impl.json`: it's just a copy of `runtime/dev_last_spec_last_impl.json`. In order to properly generate it I would have to backport this PR to an older version of the node. I think that it's not worth the effort, because we have a lot of runtime changes coming, which will correct this spec file.


CodeSandwich:
@NunoAlexandre 
I've renamed and consolidated `update-chain-spec-new-impl` and `update-chain-spec-new-spec` into `rebuild-chain-specs impl` and `rebuild-chain-specs spec`. I admit that it's cleaner now!

Issue #544, opened 2020-06-24
Don't allow users to claim unregistered ids


NunoAlexandre:
Suggested by @lftherios, we don't want users to be able to claim an id that had been used by another user, unregistered in the meantime. Once an id is unregistered, it should be forever unavailable.

@lftherios should this be the case for both orgs and users? I think the rationale applies to both but I think you were only referring to users?

lftherios:
The rationale in my opinion should apply to both.

CodeSandwich:
Now with a cost of 2 transactions one will be able to effectively destroy an ID. Having them single-use improves trust, but this allows cheap vandalism :thinking:

lftherios:
I think that's a fair point but in practice it depends on distribution. 

The way we distribute handles at launch can reduce the extend of that problem. Both ENS and Handshake at launch gave certain users the ability to claim a name or domain based on some centralized reference point.

xla:
You can also punish immedaite unregisteration by burning the deposit if the unregister happens too quickly after a register, e.g. within a year.

NunoAlexandre:
@xla @lftherios @cloudhead Currently, when a user tries to register a user or an org with an Id that is currently taken, it fails with `IdAlreadyTaken`. I think it would be useful to be granular in terms of error handling and have a separate error, something like `UnclaimableId`, that would be returned when said Id is not currently registered but it once was, therefore no longer claimable. 

I find the granularity useful for a user standpoint. If a user tries to register something with an Id they can see is not taken but that was unregistered and therefore no longer claimable, it'd be weird for them to get "Sorry, this one is already taken". In such a case, a message like "Sorry, that one is not claimable as it belonged to another user or org in the past".

What do you think?

geigerzaehler:
I second @CodeSandwich’s point. Unless we have a solution that prevents vandalism it seems dangerous to retire users on unregistering.

NunoAlexandre:
> I second @CodeSandwich’s point. Unless we have a solution that prevents vandalism it seems dangerous to retire users on unregistering. We need to think about this more carefully.

I second the importance of this issue. I think that @xla's suggestion is good and simple to implement.

PR #550, opened 2020-06-26
runtime: Forbid claim of unregistered ids

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/544

`TBD`

- [x] Find better way to accomplish a "set"-like store
- [x] Bump runtime spec


CodeSandwich:
I have a few issues with this PR, I'll prepare a PR into this branch fixing them, so we can discuss them.

runtime/src/registry.rs
```diff
@@ -91,6 +91,11 @@ pub mod store {
             // [Call::set_block_author] and not persisted.
             pub BlockAuthor: Option<AccountId>;
 
+            // The below map indexes each claimed id to data relative to the claim.
+            // We use the blake2_128_concat hasher so that the Id
+            // can be extracted from the key.
+            pub ClaimedIds1: map hasher(blake2_128_concat) Id => Option<()>;
```
geigerzaehler, line 97: I don’t think “claim” is an adequate term. I’d prefer “retired”.

CodeSandwich: Fixed

Issue #475, opened 2020-05-27
Current consensus implementation is not light client friendly


geigerzaehler:
Our implementation of the consensus is not light client friendly. To fully validate a block header a light client needs to be able to compute the target difficulty. The value of the target difficulty depends on the block timestamp which is only present in the block body or the chain state.

I’m not certain that we cannot work around this. However, to avoid surprises and keep the consensus close to proven algorithms we should put the block time stamp in the header.

This change would require either a fork of the block chain (i.e. the nodes need to be updated) or we need to reset the existing chains. The latter is less technical effort but requires coordination with existing participants. The former requires us to setup process, communication and infratstructure for the fork.

## Todo

- [ ] Decide on the strategy for releasing that change (fork or reset) and when we want to do it.
- [ ] Determine if there is an intermediate step that we can take that does not require a fork or reset.
- [ ] Validate that this works with a PoC PR

burdges:
It's proof-of-work?  We've fancier block production schemes in substrate like BABE aka Ouroboros Praos that ostensibly make this problem equally hard by requiring a validator list, but maybe a validator list fits better with many people's threat model.  We've no great way to distribute validator list to light clients yet, but we're slowly working on that problem, and we're not doing research for proof-of-work.

geigerzaehler:
Hey @burdges. Thanks for chiming in! Our consensus is indeed proof of work. One of the reasons why we chose PoW over what substrate champions is that it is more light client friendly. Of course you need to implement it properly ;). Fortunately this should be easy to accomplish.

While we’re using PoW now we’re also closely following the development on PoS and are excited about it’s possibilities. But the fact that it is still very much experimental and there are a couple unsolved problems along with governance concerns made us choose PoW. If you want to know more I’m sure @cloudhead is happy to provide more context.

burdges:
Yes governance is hard.  We've many reasons for governance, but some relevant ones include parameter changes, which PoW encounters too, and slashing refunds, which PoS avoids by binding the blockhash into the production slot assignment.  

You can build PoS without slashing.  Academic PoS schemes like Ouroboros Praos mostly never slash, well true Praos ignores equivocations like PoW does.  We're not exploring that either, but actually we should do so for parachains, and maybe it's rather small code changes.  If you've slashing then you kinda want slashing refunds partially for your own bugs, but also for operators who do engage in high availability stupidity and equivocate or whatever. 

[...]

PR #551, opened 2020-06-26
Move timestamps to header

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/475.

~It puts the timestamp in the seal of the block: it's the last piece of digest added to a block. This way the timestamp can be fully controlled by the PoW algorithm: it sets it, validates it and then uses it. Theoretically it could even alter it while mining (Bitcoin style).~


geigerzaehler:
I think it would be preferable to put the timestamp into a separate block header digest item so that it can be inspected by the runtime when a block executes.

CodeSandwich:
The timestamp is now a separate digest item in the header.

This is a hard-fork-worth breaking change, because now the PoW expects the timestamp to be present in the header. This is ok, because now we're gathering hard changes in master.

runtime/src/timestamp_in_digest.rs
```diff
@@ -0,0 +1,34 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use crate::Moment;
+use frame_system::{DigestItemOf, Trait};
+use parity_scale_codec::{Decode, DecodeAll, Encode, Error};
+use sp_runtime::{ConsensusEngineId, Digest, DigestItem};
+
+const CONSENSUS_ID: ConsensusEngineId = *b"time";
+
+pub fn load<Hash: Decode + Encode>(digest: &Digest<Hash>) -> Option<Result<Moment, Error>> {
```
geigerzaehler, line 23: This is not used by the on-chain runtime. Could we put it behind the `std` feature?

runtime/src/timestamp_in_digest.rs
```diff
@@ -0,0 +1,34 @@
...
+    digest
+        .log(|item| match item {
+            DigestItem::Consensus(CONSENSUS_ID, encoded) => Some(encoded),
+            _ => None,
+        })
+        .map(|encoded| DecodeAll::decode_all(encoded))
+}
+
+pub fn digest_item<T: Trait>(timestamp: Moment) -> DigestItemOf<T> {
```
geigerzaehler, line 32: Could we expand the return type and drop the generic `T`?
```suggestion
pub fn digest_item(timestamp: Moment) -> sp_runtime::DigestItem<crate::Hash> {
```

CodeSandwich:
> I would still prefer to have the timestamp put in the digest explicitly by the author to avoid coupling of runtime and consensus.

This isn't possible without forking the PoW consensus crate, there's zero possible configuration. But if we had full control, would we do it differently? it makes sense to check the timestamp and put it into the digest at the beginning of the inherent flow.

runtime/src/timestamp_in_digest.rs
CodeSandwich: Fixed

runtime/src/timestamp_in_digest.rs
CodeSandwich: Fixed

geigerzaehler:
>  But if we had full control, would we do it differently?

I imagined that we put the timestamp into the digest when the block header is initialized by the block proposer (not the runtime). Then we could remove the timestamp pallet altogether which would reduce the coupling.

> This isn't possible without forking the PoW consensus crate, there's zero possible configuration.

This would probably be a good idea in the long run anyways. But I’ll leave it up to you whether we want to go with this approach.

CodeSandwich:
We would need to fork the PoW, then we could drop the timestamp pallet with its inherent in favor of an in-house creation and verification mechanism of a digest item. That would definitely be cleaner, but I don't think that it's worth that kind of effort.

PR #552, opened 2020-06-29
Improve message event extraction

geigerzaehler:
We refactor the message event extraction to prepare for event compatibility and to make the code more maintainable in general.

* Structured `EventExtractionError` for event extraction
* Refactor helper functions for event extraction to reduce duplicate code.


geigerzaehler:
I don’t think it makes much sense to split this off from the substrate update. Will close

PR #553, opened 2020-06-29
runtime: Forbid claim of unregistered ids - Igor's review

CodeSandwich:
Review comment fixes for https://github.com/radicle-dev/radicle-registry/pull/550


runtime/src/registry.rs
```diff
@@ -433,35 +433,13 @@ decl_module! {
...
-    Taken,
-
-    /// The Id has been unregistered and is therefore unclaimable
-    Unclaimable,
-}
-
-fn id_availability(id: &Id) -> IdAvailability {
-    use IdAvailability::*;
-
+fn verify_id_availability(id: &Id) -> Result<(), RegistryError> {
```
geigerzaehler, line 436: Suggestion: `ensure_id_is_available` flows better in my opinion.

core/src/error.rs
```diff
@@ -134,9 +131,9 @@ pub enum RegistryError {
 
     #[cfg_attr(
         feature = "std",
-        error("the id has been unregistered and is no longer claimed")
+        error("the ID has been unregistered and can't be claimed again")
     )]
-    UnclaimableId = 18,
+    IdUnregistered = 18,
```
geigerzaehler, line 136: Suggestion: We could use the term “retired” as in `IdRetired`. This is [common terminology](https://en.wikipedia.org/wiki/Retired_number).

runtime/src/registry.rs
CodeSandwich: Fixed

core/src/error.rs
CodeSandwich: Love it, fixed

PR #554, opened 2020-06-29
Eliminate wrapper types for state objects in client

### LINKED ISSUE #521: Eliminate state wrapper types
I’m proposing to eliminate the wrapper types for state objects in the client and expose the state types directly.

---

The client has various methods to get state objects from the chain, for example `get_user()`. At the moment the return types for these functions  (e.g. `client::User`) don’t match the actual state types used by the runtime. Instead `client::User` wraps `core::state::User` and the translation happens as part of the client code. An alternative to wrapping the state types would be to expose them transparently through the client. For example, `get_user()` would return the `core::state::User` enum. 

The wrapper types serve two purposes. First, they enrich the state data with some additional data. At the moment this is only the ID for `User` and `Org`. After looking at the upstream code it turns out that this enrichment functionality is unused.

Second, they hide the different versions of the state data. While `state::User` is an enum with the different user versions in the state `client::User` is not. This effectively eliminates a match on the consumer side. However, it also means that we don’t expose the information about the state versions to the client which might be necessary at some point. In addition we can’t guarantee that we can always wrap the different versions of, say, the `User` state. In a future version this might require the `User` wrapper type to provide two versions as well.

In the bigger architectural picture `radicle-upstream` already has an [adapter](https://github.com/radicle-dev/radicle-upstream/blob/0b0d05d4da2f72f7c7bf998681c314a4a9f7cffc/proxy/src/registry.rs#L287) and [wrapper types](https://github.com/radicle-dev/radicle-upstream/blob/0b0d05d4da2f72f7c7bf998681c314a4a9f7cffc/proxy/src/registry.rs#L197-L202) for the client. If enrichment and abstracting the versions is indeed required this should be the place. The client API should be as unopinionated as possible.

geigerzaehler:
Closes #521

### DIFF
--- a/client/tests/end_to_end.rs
+++ b/client/tests/end_to_end.rs
@@ -48,28 +48,27 @@ async fn register_project() {
         let initial_balance = match &domain {
             ProjectDomain::Org(org_id) => {
                 let org = client.get_org(org_id.clone()).await.unwrap().unwrap();
-                client.free_balance(&org.account_id).await.unwrap()
+                client.free_balance(&org.account_id()).await.unwrap()
             }
             ProjectDomain::User(user_id) => {
                 let user = client.get_user(user_id.clone()).await.unwrap().unwrap();
-                client.free_balance(&user.account_id).await.unwrap()
+                client.free_balance(&user.account_id()).await.unwrap()
             }
         };
 
         let random_fee = random_balance();
         let message = random_register_project_message(&domain, checkpoint_id);
+        let project_name = message.project_name.clone();
         let tx_included = submit_ok_with_fee(&client, &author, message.clone(), random_fee).await;
         assert_eq!(tx_included.result, Ok(()));
 
         let project = client
-            .get_project(message.project_name.clone(), message.project_domain.clone())
+            .get_project(project_name.clone(), domain.clone())
             .await
             .unwrap()
             .unwrap();
-        assert_eq!(project.name.clone(), message.project_name.clone());
-        assert_eq!(project.domain.clone(), message.project_domain.clone());
-        assert_eq!(project.current_cp.clone(), checkpoint_id);
-        assert_eq!(project.metadata.clone(), message.metadata.clone());
+        assert_eq!(project.current_cp().clone(), checkpoint_id);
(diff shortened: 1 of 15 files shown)


client/tests/end_to_end.rs
```diff
@@ -48,28 +48,27 @@ async fn register_project() {
...
         assert_eq!(tx_included.result, Ok(()));
 
         let project = client
-            .get_project(message.project_name.clone(), message.project_domain.clone())
+            .get_project(project_name.clone(), domain.clone())
             .await
             .unwrap()
             .unwrap();
-        assert_eq!(project.name.clone(), message.project_name.clone());
-        assert_eq!(project.domain.clone(), message.project_domain.clone());
```
CodeSandwich, line 70: Why aren't we testing the `name` and `domain` anymore?

geigerzaehler: `name` and `domain` are not stored in the state and thus not part of the returned `project` object anymore. Before we just assigned them to the returned object from the input argument for convenience.

geigerzaehler:
> The scale of changes in the tests is staggering! I wonder if they could be avoided? We've introduced the bare bones client, but we haven't created a high level wrapper client for the CLI and the tests. The API of the old client was tailored for these two use cases, maybe we could've moved and reused it as a wrapper? Now every single change to the bare bones client will have exactly this magnitude of manual test adjustments.

Yes, I had to change a fair amount of code. (Although some of the changes just eliminate redundancies and inconsistencies.) Most of the changes are very mechanical though.

You’re right that a change in the client API multiplies in changes to the tests and that this could be alleviated by introducing an adapter. However, at this point I believe an adapter would create too much overhead to maintain. The test changes are still manageable and I don’t expect the client API to change that often.

Where I believe we can improve however is creating more abstraction for common scenarios like we’re already doing for registering an org with some funds.

CodeSandwich:
> Where I believe we can improve however is creating more abstraction for common scenarios like we’re already doing for registering an org with some funds.

This sounds exactly like a specialized, high level client adapter just for tests. I don't see much maintenance overhead here, it should be strictly private for registry, so any necessary changes can be imposed without caring about dependees.

PR #559, opened 2020-06-30
Expose frame_system in public runtime API

geigerzaehler:
We expose the `frame_system` crate from the runtime API. This allows the client to access the types defined by `frame_system` from old runtime versions to provide compatibility. In our case we specifically need it for `DisptachError`.


geigerzaehler:
Turns out this is not needed

PR #561, opened 2020-07-02
Add missing tip header error

geigerzaehler:
A step towards eliminating `Error::Other` in the client. Instead of returning an error if a block is missing we return an option. For the best chain tip we create a dedicated error variant.

### DIFF
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -203,16 +203,16 @@ impl backend::Backend for RemoteNode {
         Ok(keys.into_iter().map(|key| key.0).collect())
     }
 
-    async fn block_header(&self, block_hash: Option<BlockHash>) -> Result<BlockHeader, Error> {
+    async fn block_header(
+        &self,
+        block_hash: Option<BlockHash>,
+    ) -> Result<Option<BlockHeader>, Error> {
         self.rpc
             .chain
             .header(block_hash)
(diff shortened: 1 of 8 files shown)


client/src/backend/remote_node.rs
```diff
@@ -203,16 +203,16 @@ impl backend::Backend for RemoteNode {
...
             .chain
             .header(block_hash)
             .compat()
-            .await?
-            .ok_or_else(|| match block_hash {
-                Some(hash) => format!("Header not found for block hash {}", hash).into(),
-                None => "Header not found for the best chain tip".into(),
-            })
+            .await
+            .map_err(Error::from)
```
CodeSandwich, line 215: If `block_hash == None` and `.header(block_hash)` returns `None`, shouldn't we return `Err(Error::BestChainTipHeaderMissing)`?

geigerzaehler: We could but this wouldn’t change the behavior of the client. It would just make the contract for the backend more complicated: If the argument is `None` you’re guaranteed to get `Some` header or an error. In this case I’d opt for keeping the contract (and the implemenation) of the header simple and being forced to handle it in the client.

CodeSandwich: Ah yes, that's just the backend and in the client we do use it!

PR #562, opened 2020-07-02
Update substrate to v2.0.0-rc4

geigerzaehler:
*Blocked until we decide to [break ffnet compatibility](https://radicle.community/t/phasing-out-the-ffnet/238)*

Substrate v2.0.0-rc4 comes with the following changes.

* Added a `transaction_version` field to `sp_version::RuntimeVersion`.   This requires a runtime spec update and breaks compatibility between   client and node (but not compatibility between runtimes).
* Weight annotations for transactions have changed. We set everything to zero weight for now. We also don’t use substrate transaction fee system.
* On the node service side changed interfaces requires some refactoring.


scripts/build-runtime-wasm
```diff
@@ -8,6 +8,6 @@ if [ $# -ne 1 ] || [ "$1" = --help ] ; then
 fi
 
 export BUILD_WASM_BINARY_OUT_DIR="$(mktemp -d)"
-cargo build -p radicle-registry-runtime --release
+cargo build -p radicle-registry-runtime
```
CodeSandwich, line 11: Won't it result in debug builds of the runtime?

geigerzaehler: I dont’t think so. The build profile for the wasm runtime is set separately: https://github.com/radicle-dev/radicle-registry/blob/80faa28de30327766e1932f91f9b29d8a7aa09be/runtime/build.rs#L17-L19

However, on CI this will not be able to use the previously built artifacts. I will revert the change.

geigerzaehler:
@CodeSandwich Could you have another look?

runtime/src/lib.rs
```diff
@@ -77,11 +81,14 @@ pub type UncheckedExtrinsic = generic::UncheckedExtrinsic<AccountId, Call, Signa
...
 
+pub const SPEC_VERSION: u32 = 15;
+
 /// This runtime version.
 pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("radicle-registry"),
     impl_name: create_runtime_str!("radicle-registry"),
-    spec_version: 14,
+    spec_version: SPEC_VERSION,
+    transaction_version: SPEC_VERSION,
```
CodeSandwich, line 91: `transaction_version` bump implies `spec_version` bump, but `spec_version` bump doesn't imply `transaction_version` bump. That means that `spec_version` is decoupled from `transaction_version` and `spec_version` will always be always equal or greater than `transaction_version`. I think that it makes sense to set `transaction_version` to 0 or 1.

runtime/src/registry.rs
```diff
@@ -164,7 +166,7 @@ decl_module! {
             frame_support::traits::OnKilledAccount<AccountId>
     {
         fn deposit_event() = default;
-        #[weight = SimpleDispatchInfo::InsecureFreeNormal]
+        #[weight = (0, Pays::No)]
```
CodeSandwich, line 169: Why are we dropping the weight system? I know that it's a conscious decision, but I don't know the motivation.

scripts/update-substrate
```diff
@@ -10,16 +10,11 @@ if [[ -z "$new_rev" ]]; then
...
-  echo "Invalid revision format"
-  exit 1
-fi
-
 # Reads `Cargo.lock` and extracts the first revision of the `substrate`
 # repo that is used.
 function get_substrate_rev () {
   while read -r LINE; do
-    if [[ "$LINE" =~ ^\source\ .*git\+https://github\.com/paritytech/substrate.*\?rev=([0-9a-f]{40,40}) ]]; then
+    if [[ "$LINE" =~ ^\source\ .*git\+https://github\.com/paritytech/substrate.*\?rev=([^#]+) ]]; then
```
CodeSandwich, line 17: Why `#`? The `rev` in `Cargo.lock` ends with `"`.
```suggestion
    if [[ "$LINE" =~ ^\source\ .*git\+https://github\.com/paritytech/substrate.*\?rev=([^"]+) ]]; then
```

node/src/metrics.rs
```diff
@@ -16,19 +16,20 @@
...
-use substrate_prometheus_endpoint::prometheus::core::Atomic;
-use substrate_prometheus_endpoint::{Gauge, Registry, U64};
+use substrate_prometheus_endpoint::{prometheus::core::Atomic, Gauge, Registry, U64};
 
+use crate::blockchain::Block;
 use crate::pow::Difficulty;
+use crate::service::FullClient;
 
-pub fn register_metrics<S>(service: &S) -> Result<(), Error>
+pub fn register_metrics<S>(service: &S, client: Arc<FullClient>) -> Result<(), Error>
```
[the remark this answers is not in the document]
geigerzaehler: Using `service.client()` requires type constraints on `S::Client`. I wasn’t able to express them in a way that satisfied the type checker.

scripts/update-substrate
geigerzaehler: The script reads from `Cargo.lock`. There the source looks like this:
~~~
git+https://github.com/paritytech/substrate?rev=v2.0.0-rc4#00768a1f21a579c478fe5d4f51e1fa71f7db9fd4
~~~

runtime/src/registry.rs
geigerzaehler: We’re not really dropping the weight system. The new annotation is equivalent to the old one.

runtime/src/lib.rs
geigerzaehler: You’re right that a `spec_version` bump does not necessitate a  `transaction_version` bump. However, `transaction_version`is what is signed in a transaction and checked by `CheckTxVersion`. To keep the versioning scheme for the runtime simple we should move `transaction_version` in lock step.

We can discuss a more elaborate versioning scheme separately.

node/src/metrics.rs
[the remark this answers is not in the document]
CodeSandwich: We're hitting a very dumb spot of the compiler :facepalm: It requires to lay down all the constraints and all the relations between them in the `where` clause. I understand why there was such a generic boundary before. What was wrong with it, did it stop working after the update?

runtime/src/registry.rs
CodeSandwich: Ah, I misunderstood the PR description. Nothing really changed.

node/src/metrics.rs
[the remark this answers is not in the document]
geigerzaehler: > did it stop working after the update?

Yes indeed. Before the update `service.client()` returned `Client<S::Backend, S::CallExecutor, S::Block, S::RuntimeApi>`. But now it returns `S::Client`. For our code to work `S::Client` needs to implement multiple interfaces but adding those as a constrained does not work because they in turn conflict with other constraints.

If you can make it work without it being overly complex I’d love to include the change.

client/src/backend/remote_node.rs
```diff
@@ -240,9 +240,11 @@ impl backend::Backend for RemoteNode {
 }
 
 async fn check_runtime_version(rpc: &Rpc) -> Result<(), Error> {
-    match runtime_version(rpc, None).await?.spec_version {
-        9 | 10 | 11 | 12 | 13 | 14 => Ok(()),
-        other => Err(Error::IncompatibleRuntimeVersion(other)),
+    let spec_version = runtime_version(rpc, None).await?.spec_version;
+    if spec_version == VERSION.spec_version {
```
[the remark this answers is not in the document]
geigerzaehler: We can reintroduce the `match` when we need it again. For now, I think this is the safer version.

client/src/backend/remote_node.rs
```diff
@@ -240,9 +240,11 @@ impl backend::Backend for RemoteNode {
...
+        Ok(())
+    } else {
+        Err(Error::IncompatibleRuntimeVersion(spec_version))
```
CodeSandwich, line 247: I think that keeping the `match` is more welcome for future us trying to extend the list of supported versions:
```suggestion
    const CURRENT_SPEC_VERSION: u32 = VERSION.spec_version;
    match runtime_version(rpc, None).await?.spec_version {
        CURRENT_SPEC_VERSION => Ok(()),
        other => Err(Error::IncompatibleRuntimeVersion(other)),
```
Of course if we want to drop the explicit list of versions. I'm concerned that now it will be dangerously easy to accidentally remove perfectly fine support for the previous version if we introduce a new one.

client/src/backend/remote_node.rs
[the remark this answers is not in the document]
CodeSandwich: Continued in https://github.com/radicle-dev/radicle-registry/pull/562#discussion_r451638256

node/src/metrics.rs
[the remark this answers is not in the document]
CodeSandwich: Yes! https://github.com/radicle-dev/radicle-registry/pull/580

client/src/backend/remote_node.rs
geigerzaehler: This is great actually. I wasn’t aware that `VERSION` was constant and that you could easily get the field.

PR #563, opened 2020-07-02
Eliminate client::Error::Other variant

geigerzaehler:
Eliminate the `client::Error::Other` variant and replace it with dedicated variants for the different possible errors.

### DIFF
--- a/client/src/backend/remote_node.rs
+++ b/client/src/backend/remote_node.rs
@@ -25,13 +25,12 @@ use sc_rpc_api::{author::AuthorClient, chain::ChainClient, state::StateClient};
 use sp_core::{storage::StorageKey, twox_128};
 use sp_rpc::{list::ListOrValue, number::NumberOrHex};
 use sp_runtime::{generic::SignedBlock, traits::Hash as _};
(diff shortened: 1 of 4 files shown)


client/src/backend/remote_node.rs
```diff
@@ -77,12 +76,7 @@ impl RemoteNode {
             .await?;
         let genesis_hash = match genesis_hash_result {
             ListOrValue::Value(Some(genesis_hash)) => genesis_hash,
-            other => {
-                return Err(Error::Other(format!(
-                    "Invalid chain.block_hash result {:?}",
-                    other
-                )))
-            }
+            _ => return Err(Error::InvalidBlockHashResponse),
```
CodeSandwich, line 79: We were passing the full response into the error, do we want to stop that? On one hand it was probably a huge chunk of text, but on the other it was giving some debugging material.

geigerzaehler: Good point! I’ve add a `response` parameter but I don’t print it as part of the error message.

CodeSandwich: Brilliant solution, only debug prints will be littered :)

Issue #547, opened 2020-06-25
Remove panic from runtime code


geigerzaehler:
We have a `panic` in the runtime code that may be triggered by a transaction
https://github.com/radicle-dev/radicle-registry/blob/811fe3a13ea13c81f46bb302e3cb4c84afba4132/runtime/src/registry.rs#L372

PR #564, opened 2020-07-03
Fix runtime panic on user project checkpoint set

CodeSandwich:
Fixes https://github.com/radicle-dev/radicle-registry/issues/547.

The tests needed some cleanup, because there would be too many repetitions.


test-utils/src/lib.rs
```diff
@@ -141,6 +140,15 @@ pub fn random_register_user_message() -> message::RegisterUser {
...
 }
 
+pub fn random_create_checkpoint_message(
+    previous_checkpoint_id: impl Into<Option<CheckpointId>>,
+) -> message::CreateCheckpoint {
+    message::CreateCheckpoint {
+        project_hash: H256::random(),
+        previous_checkpoint_id: previous_checkpoint_id.into(),
+    }
+}
```
geigerzaehler, line 150: This function is only used once and only fills a random project hash. I’d suggest to remove it and inline in the one place.

CodeSandwich: Fixed. Good catch!

PR #565, opened 2020-07-06
Rename client Codec error to StateDecoding

geigerzaehler:
We rename the `Error::Codec` to `Error::StateDecoding` to make it explicit where this error originates. We also add the `key` parameter to simplify debugging.

### DIFF
--- a/client/src/error.rs
+++ b/client/src/error.rs
@@ -22,9 +22,14 @@ use crate::event::EventExtractionError;
 /// Error that may be returned by any of the [crate::ClientT] methods
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Decoding the received data failed
-    #[error("Decoding the received data failed")]
-    Codec(#[from] CodecError),
+    /// Failed to decode state value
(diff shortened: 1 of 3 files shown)


client/src/error.rs
```diff
@@ -22,9 +22,14 @@ use crate::event::EventExtractionError;
 /// Error that may be returned by any of the [crate::ClientT] methods
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Decoding the received data failed
-    #[error("Decoding the received data failed")]
-    Codec(#[from] CodecError),
+    /// Failed to decode state value
+    #[error("Failed to decode state value")]
```
CodeSandwich, line 26: Is the regular (CLI-ish) consumer of an error interested in the failed key?

geigerzaehler: I don’t think so. That’s why I’ve omitted it from the regular message.

CodeSandwich: Ok. I'm hesitant, but if we ever change our mind, we'll be able to easily add it.

PR #567, opened 2020-07-06
End to end tests can run twice

geigerzaehler:
We use randomly generate accounts instead of the “//Alice” root account throughout the tests. This enables us to run the end-to-end tests twice against the same state. This might also eliminate some conflicting transactions in the tests.


geigerzaehler:
So this definitely does not eliminate the confliciting transactions in the end to end tests.

CodeSandwich:
THE CI HAS SPOKEN

Issue #578, opened 2020-07-08
client: Provide call to check id availability


NunoAlexandre:
In https://github.com/radicle-dev/radicle-registry/pull/550, before my commit was modified and force-pushed against, I had introduced a type to describe the state of an Id: `Available`, `Taken`, or ~~`Unclaimable`~~ `Retired`. The goal of this type and the function returning it was to offer a client call to check the availability of given id.

We want to call `client.get_id_availability(an_id)` and get one of those 3 possible variants and provide the appropriate validation messages in the upstream.

PR #579, opened 2020-07-08
client: check id status

NunoAlexandre:
Closes https://github.com/radicle-dev/radicle-registry/issues/578


runtime/src/registry.rs
```diff
@@ -434,13 +434,33 @@ decl_module! {
...
+
+    /// The id is curently taken by a user or by an org
+    Taken,
+
+    /// The id has been unregistered and is now retired
+    Retired,
+}
+
+pub fn get_id_status(id: &Id) -> IdStatus {
     if store::Users1::get(id).is_some() || store::Orgs1::get(id).is_some() {
```
CodeSandwich, line 450: Nitpick:
```suggestion
    if store::Users1::contains_key(id) || store::Orgs1::contains_key(id) {
```

PR #582, opened 2020-07-09
Run master builds also on agent preview

geigerzaehler:
We build master also on agents that are running the preview version of our CI setup. This allows us to test the new setup more thoroughly

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -21,6 +21,13 @@ steps:
     concurrency: 1
     concurrency_group: master
     <<: *test
+  - label: "Test on preview"
+    branches: master
+    concurrency: 1
+    concurrency_group: master
+    agents:
+      queue: agent-preview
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yaml
```diff
@@ -21,6 +21,13 @@ steps:
     concurrency: 1
     concurrency_group: master
     <<: *test
+  - label: "Test on preview"
+    branches: master
+    concurrency: 1
+    concurrency_group: master
+    agents:
+      queue: agent-preview
+    <<: *test
```
CodeSandwich, line 30: Can it inherent from the regular master build? The only change is the agent and the label.

geigerzaehler: Yes, possibly. With YAML it’s a bit cumbersome to dry this up and we only save three lines. So in this case I’m prefering copy/paste

PR #588, opened 2020-07-14
Fix get_id_status

NunoAlexandre:
Currently, calling Client::get_id_status fails because it delegates to
runtime::get_id_status, which - in this context - runs outside an
externalities environment, making access to the storage fail to
happen. The solution here is about moving the logic to the client by
looking up the storage via the backend, just like we do for all other
storage lookups from the client, with the addition of a new fn to check
whether a given key exists in a specific store. Here we also move the
IdStatus definition to the client since it became the only placed where
it is used.


NunoAlexandre:
Thanks, @CodeSandwich :raised_hands:

PR #589, opened 2020-07-15
Initial implementation of link identity on chain attestation (WIP).

massimiliano-mantione:
This PR adds a blob of data to the user on chain state, `link_user`, which is meant to be a reference to a specific user identity revision in Radicle link.

This user revision is expected to be properly signed and to contain the registry user id, creating a two-way attestation of the Radicle link and registry user identities.

Parsing this blob and checking it against the provided Radicle link user is not the registry's responsibility, it is up to Upstream or whatever application will access the data to use it properly.

Todo:

- [x] Fix the build (I am just new to the registry, tests are passing locally but I am missing something).
- [x] I started making `link_user` and `Option` but it should not be so (I will remove the `Option` and leave the plain blob).
- [x] Complete the work on the registry CLI so that it is usable (likely with a base64 representation of the blob).
- [x] Give better names to properties (help welcome during the code review).
- [ ] Likely more: this is my 1st attempt at implementing something in the registry :-)


core/src/state.rs
```diff
@@ -381,6 +409,22 @@ pub struct UserV1 {
...
+///
+/// * `account_id` is immutable
+/// * `projects` is a set of all the projects owned by the User.
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct UserV2 {
+    /// Account ID that holds the user funds.
+    pub account_id: AccountId,
+
+    /// Reference to the radicle link user metadata last revision.
+    pub link_user: Bytes128,
```
CodeSandwich: Would it make sense to make it `Option<Bytes128>`? We obviously have a notion of a registry user not being linked to any link user, it probably makes sense to express it in the type system, we won't need to check if it's empty.

I think that it would simplify this PR, a lot of code was written for this case.

CodeSandwich (changes requested):

core/src/state.rs
```diff
@@ -381,6 +409,22 @@ pub struct UserV1 {
...
+
```
CodeSandwich, line 422: Should `link_user` be an opaque blob of undefined length? What's inside? Is it going to keep changing? Who will be taking care of supporting all versions?

CodeSandwich (changes requested):

core/src/state.rs
```diff
@@ -333,13 +333,15 @@ impl OrgV1 {
 #[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
 pub enum Users1Data {
     V1(UserV1),
+    V2(UserV2),
```
CodeSandwich, line 336: Do we need to extend `UserData`? Would it be feasible to have a separate store? I really don't know, but I think that it's worth considering.

- A con would be probably keeping data integrity, e.g. we haven't decided yet if users should be deletable.
- A pro would be that this is registry-specific data, but the `link_user` field is a part of integration with a separate, independent system, so they should be able to evolve separately.

core/src/state.rs
massimiliano-mantione: I initially implemented it with the `Option` :smiley: 
On the product side @cloudhead says this should be mandatory but IMHO in practice it could happen that there is not link identity.

On the other hand, if there _is_ a link user identity, the blob cannot be empty.
So with the `Option` the type system would allow the meaningless state of a `Some(Bytes128::empty()))`.
Without `Option` the "no identity" would be the empty vector.

I have no strong opinion, I'll do this however the registry team prefers.

core/src/state.rs
massimiliano-mantione: Inside there are a couple of hashes that ultimately refer to a specific git commit containing the "feed event" that defines a specific Radicle link user identity revision.
Therefore it cannot be "too large", the size depends on the hashing algos that will be used (they can change over time).
The encoding specifies the algos.

The registry cannot check anything about it because it would need the full blob of the user definition, which is too large to stay on chain, and the logic to parse it and verify it, which is even worse.
All the checks are up to Upstream (or any app that uses these data).

This blob is here as a link: it proves that this registry user claims to be that Radicle link user (which in turn claims to be this registry user).

The fact that "link" means both "_reference_ to a user" and "user in Radicle _link_" can be unfortunate when understanding the wording :smile:

core/src/state.rs
massimiliano-mantione: This piece of data correlated a user `ID` on registry to what is referred by the blob (see above).
For me it makes sense to put it in the user state.
If the user is deleted from the registry state it makes sense to also delete the association with their Radicle link identity, and this happens naturally keeping the reference inside the user state.

In a sense it is true that Radicle link is a "separated, independent system", but it should be the natural off-chain counterpart of Radicle registry.

massimiliano-mantione:
BTW, why is CI failing? all checks are passing here...

core/src/state.rs
CodeSandwich: > The encoding specifies the algos.

Would it be possible wrap possible encoding variants in an enum? This way it will be possible to have multiple consumer implementations. Now we're heading toward storing a blob which nobody will be able to understand except the newest version of radicle-link. If we want to store something in registry, it's the registry that should define and document encoding.

> The registry cannot check anything about it

Yes and no, the registry won't check if data is correct, but it can at least tell if it's sane and properly encoded, sized and versioned. It's a data store like a JSON file: it doesn't know if number `3000` makes sense in a given context, but it can tell that number `30zero0` is surely invalid.

CodeSandwich:
> BTW, why is CI failing? all checks are passing here...

You haven't run `scripts/rebuild-chain-specs`, see `DEVELOPING.md`. All in good time, first let's discuss the overall design :wink:

core/src/state.rs
CodeSandwich: It's not mandatory, at least not yet, we're throwing `Bytes128::empty()` left and right. The consumer of this data must check if it's empty and support cases when it is. If it's going to be mandatory, we need a more sweeping change to the runtime and its API.

NunoAlexandre: Yeah, I also think it should be Optional, if it might not be present in some cases. Then we know whether there actually is a link user associated or not. Will this field be mutable through some new tx we will add?

kim:
For posterity: the value of this feature is very limited if the transaction is not also validated such that the sequence of on-chain attestations corresponds to those made off-chain (iow: each revision of a checkpointed identity must refer to the previous one, and the transaction be rejected if they don't match). Otherwise, an attacker who gains control over the transaction-signing key can _fork_ the off-chain identity without also having to prove ownership of the `radicle-link` keys. The feature would thus be approximately as useful as publishing the `link` revision hash on Twitter, except for censorship-resistance.

CodeSandwich:
> each revision of a checkpointed identity must refer to the previous one, and the transaction be rejected if they don't match

Would it be enough if the link update message had the previous hash and compare it with the old one before accepting? Do the users (clients, not the node) need access to the full chain of checkpoints in order to gain trust?

core/src/state.rs
```diff
@@ -348,13 +350,23 @@ impl Users1Data {
...
             Self::V1(user) => user.account_id,
+            Self::V2(user) => user.account_id,
+        }
+    }
+
+    /// Latest attested user identity on radicle link.
+    pub fn link_user(&self) -> &Option<Bytes128> {
+        match self {
+            Self::V1(_) => &None,
+            Self::V2(user) => &user.link_user,
```
CodeSandwich: How about this? It would make working with an API easier.
```suggestion
    pub fn link_user(&self) -> Option<&Bytes128> {
        match self {
            Self::V1(_) => None,
            Self::V2(user) => user.link_user.as_ref(),
```

massimiliano-mantione: :facepalm: Done, thanks!

core/src/state.rs
massimiliano-mantione: Thanks for the input, now it is optional again.

cloudhead: Note that if it’s made optional, it should only be allowed to be set once when None. Once it is Some, it is immutable.

cloudhead: Hm I think I’m misunderstanding: the behavior we need is that the urn can never change. However, it’s okay to update the latest revision hash.

cloudhead: I was also under the impression that we decided against having the hashes in there since a key compromise would destroy the history. In any case, the functionality that is needed is an immutable link to a user identity/URN.

massimiliano-mantione: So, just to be sure, the `link_user` property should be an `Option<Bytes128>` in the state but a plain `Bytes128` in the message, and `set_link_user` must fail if the link info in the state  is not `None`.

Correct?

core/src/state.rs
CodeSandwich: @massimiliano-mantione What's your opinion? Do you agree that the ecosystem would benefit from exchanging a blob for something with a well defined format?

massimiliano-mantione: The ecosystem would indeed benefit from having a well-known format for that blob.
It's not that I just agree on this, I am convinced of this.

What I am not sure about is whether this format should be enforced by the registry or not.

This blob is something that Radicle Upstream will handle (through Radicle Link), and Its format could evolve over time.

What the registry is attesting is "this is the registered Radicle Link identity".
It could be valid or invalid for reasons beyond its format, the registry cannot know.

Therefore I am not sure that adding a purely "format" verification to the registry adds a real value: the identity could be invalid anyway.

But adding that "format" verification will tie the registry code to the Radicle Link evolution of that format.
It will make evolving that format in Radicle Link impossible without a registry core upgrade (adding the logic to parse the evolved Radicle link identity format).

So this is the tradeoff: the value of a "format" verification at on chain transaction time vs the burden of having to update the transaction logic every time that format changes.
My _instinct_ says not to add this code to the registry but in truth I am not able to weight this tradeoff properly.

CodeSandwich: You've got a point, the evolution of the link identity would be totally crippled with mandatory runtime updates and unnecessarily tight coupling with registry.

core/src/state.rs
massimiliano-mantione: It turns out that this PR was not distinguishing between two distinct features:

1. Cross linking an immutable radicle link user identity (a URN) and a radicle registry user ID.
2. Certifying the evolution of a radicle link user identity, storing a reference to its current _revision_, which can change over time.

This PR will stick to (1) and leave point (2) open.
For clarity the user property will be renamed `link_urn`.

massimiliano-mantione: Commit 596d6129f5d972add5bdd0bc4ad46f0877aab9c8 implements the above comment.

CodeSandwich: ~It seems like there is absolutely no data validation. This leads to a situation when a single link user can be linked to hundreds of registry users, many of which can be controlled by evil people or even none controlled by the actual project owner. Is it ok? If yes, how is the registration useful and what does it prove?~

CodeSandwich:
We had a short chat with @lftherios @geigerzaehler and @NunoAlexandre and it seems that the attestation is just one use case of general identity linking feature. Ideally every user should have a list or a dictionary with different IDs or keys. It looks very much in scope of this PR to generalize the current design, what do you think?

massimiliano-mantione:
> We had a short chat with @lftherios @geigerzaehler and @NunoAlexandre and it seems that the attestation is just one use case of general identity linking feature. Ideally every user should have a list or a dictionary with different IDs or keys. It looks very much in scope of this PR to generalize the current design, what do you think?

That the concept is perfect, what I do not know is whether this collection should be on chain or off chain.

I mean, I see two extremes in this:

1. The collection of "external attestations" (IDs or keys) is entirely on chain.
2. On chain we store a hash that has been obtained by hashing a canonical storage form of the collection of attested IDs or keys. This collection is stored off chain, likely in a Radicle Link event feed, which ultimately means in a specific blob in the git repositories that participate in the Radicle Link replication process.

In any case each attested ID or key would need a semantics attached to it.
For instance, if I wanted to attest that I am `M_a_s_s_i` on Twitter, storing `M_a_s_s_i` in the collection would not be enough, the collection entry would need to be more like `"twitter": "M_a_s_s_i"`.
And this "semantic tag" would need to be extensible (we start with `"rad", "github", "gitlab", "twitter"... but more will pop up).

The tradeoff between (1) and (2) is cost of on chain storage vs content availability: with (2) we only know that the user attests the hash of a collection, then we have to go get it.
In any case the collection itself should be mutable (in solution (2) the user would attest different hashes over time, the user on chain state would contain the last one).

And this, BTW, intersects with the "user profile" feature: is this collection of external identities part of the user profile? (I would say yes, but we have to settle on this).
And where is the profile stored?
Also note that the user profile is also supposed to be mutable.

In the rest of this comment I will explore path (2) a bit more.

One idea would be to store this collection of attestations inside the Radicle Link user profile, and, to extend this feature so that we store a reference to a specific user identity revision, in the form of a pair of hashes.
One is the current `link_urn`, referring to the identity as a whole.
The other would be something like `link_rev`, referring to the last (which means "current") certified revision of the user profile, and it is meant to change over time.

I do not think it should be _mandatory_ to certify every revision of a user identity.
It would be _desirable_, but I still see a value in certifying only _some_ of them, probably the ones that contain changes that the user really cares about and not just minor changes.

Of course, my 2c :-)

geigerzaehler:
From my perspective the implementation of this looks good. The CI build should fix itself if you wait for #601.

It is difficult to judge the design of this though. I’m lacking context on what we want to achieve here in the scope of attestation. I’m also wondering what the next steps are since it seem that we need to add additional functionality to realize attestation. A document that explains all this would be highly appreciated.

In summary I’d prefer to wait until we have some more insight into the problem and design. But I’d also be ok with merging it.

NunoAlexandre:
@massimiliano-mantione, thank you so much for your patience and sorry for this slow and tedious first experience in the Registry.

> From my perspective the implementation of this looks good. The CI build should fix itself if you wait for #601.
> 
> It is difficult to judge the design of this though. I’m lacking context on what we want to achieve here in the scope of attestation. I’m also wondering what the next steps are since it seem that we need to add additional functionality to realize attestation. A document that explains all this would be highly appreciated.
> 
> In summary I’d prefer to wait until we have some more insight into the problem and design. But I’d also be ok with merging it.

We should wait. There is no objective value in merging code that won't be used or that will be changed shortly after. @massimiliano-mantione would that work for you?

lftherios:
I think that the trade-offs that @massimiliano-mantione is outlining are the right ones. Let's have a kick-off meeting for attestations today or tomorrow and we can provide context for beta-1, main-net and beyond.

massimiliano-mantione:
> We should wait. There is no objective value in merging code that won't be used or that will be changed shortly after. @massimiliano-mantione would that work for you?

Of course I am, we need to get this right :-)

And BTW, this will need to be rebased onto master anyway, and I'll actually need to redo all the commits because I forgot to setup this cloned repo to also sign them (no commit is signed on this branch).

Let's do the meeting, document this, and get it right.

massimiliano-mantione:
I added a document explaining what this feature does in general, comments are welcome so we can define it properly.
In the context of that document the goal of this PR is to implement the registry side of the identity attestation MVP for users.

BTW, after the review is complete I'll have to redo all these commits anyway because they are not signed.
I was planning on rebasing everything on master but this does not play nice with github PRs (comments lose their pointers to the code).

Should I instead do a new branch and open a new PR that will link to this? This should preserve the discussion properly.
Otherwise I can just rebase :-)

CodeSandwich:
Don't worry about rebasing after the PR is accepted. You'll probably need to rebase anyway to squash the comment-fixing commits, but by then the comments probably won't be relevant anyway :)

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
...
+One of them is the "security model" of this attestation: what it proves when everything works as intended is clear, but what to deduce is anything goes wrong is not (like, how to handle user key theft or compromise, or how to show partial attestations where only one of the two links is in place).
+
+Related to the security model is the problem of handling key rotation on both sides of the link (both on the Registry and Link).
+
+### Registry storage cost
+
+Then there is the "registry storage cost" problem: in principle the registry state should contain links to every supported external identity (Link, Github, Twitter...) but this list can be quite long.
+
+An easy solution to this problem is to only store a single link inside the registry, unambiguously referring to an external piece of metadata that in turn will contain all other external links.
+To guarantee absence of ambiguity this link should be a cryptographic hash of the external metadata.
```
CodeSandwich, line 46: One thing to keep in mind is that every update of the hash is stored on the blockchain too. The data savings here are much smaller than what it seems on the first sight, but simplicity and accessibility decrease significantly.

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
...
+### Data format and verification
+
+Let's say we decide to store the Link user identity in the Registry user state; we should then decide how to store it.
+In Link identities are represented by Radicle URNs, as described in the (identity resolution)[https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/identity_resolution.md] and (identities)[https://github.com/radicle-dev/radicle-link/pull/248] specifications.
+
+However we have the following issues:
+
+- Registry does not depend on Link.
+- The URN format is not compact, while in the registry state every byte counts.
+- The URN format could evolve, or this feature (identity attestation) could evolve, allowing extensions to store something more than Link URNs, or something different; if we wanted Registry to be able to verify the format of this piece of data the core Registry code would need to be updated before each data format change.
```
CodeSandwich, line 60: It's worth writing down that the data format should be versioned, this CAN NOT be forgotten in the data design process and it MUST be an available feature from day 1

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
+# External Attestations
+
+## Feature goals
+
+The Radicle Registry state contains data about users, organizations and software projects.
+However, the same entities will also have an identity off chain, in other systems.
+For instance, a Registry user could also have an identity on Radicle Link (hereinafter "Link"), or on Github, Gitlab, Twitter, or they will have an email address...
```
CodeSandwich: This is probably not too early to make this design decision: can a user be linked to multiple identities of the same kind? For example can I have 2 GitHub accounts (1 professional and 1 private) and 3 email addresses (1 from company A, 1 from B and 1 personal)? Should I be able to have multiple link identities?

massimiliano-mantione: This is a very good point.

The design of the "external identities collection", where "external" means "outside of Radicle Link and Registry" is probably entirely up to Upstream, and I would not see a problem allowing several email addresses or something similar.
That data would be part of the user profile, there could be anything there.

About Registry and Link identities, of course a user could create more than one
IMHO this would be a bit against the spirit of the platform but there is nothing we can do to prevent it.

However we are not required to make it _easy_ to work with multiple Link and Radicle identities, and this feature should stick to one-to-one attestation of Radice and Link IDs.

Unless the product team thinks this is a scenario we need to actively support...

docs/external-attestations.md
massimiliano-mantione: Does line 80 of this document suffice?
It is part of the list of requirements for the MVP.

```
* Specify and implement the compact (and extensible) URN binary encoding format (in Link).
```

I could remark "versioned and extensible" but the intention behind "extensible" is "that can evolve in a formalized way".

docs/external-attestations.md
massimiliano-mantione: True, but this growth problem would exist for whatever mutable content we would store.

I mean, if what we store is a hash, the blockchain history will grow containing all the registered hashes, but if we store a full metadata blob it would grow containing all the blobs.

The growth is there anyway, and the space vs data accessibility tradeoff is the same.

One of the reasons why I keep insisting that we should not require users to register every revision of their Link metadata is to mitigate this "growth problem": only revisions containing info worth of being attested should be stored, others can be skipped.

docs/external-attestations.md
CodeSandwich: Yes, sorry, I haven't noticed it :)

docs/external-attestations.md
xla:
```suggestion
For instance, a Registry user could also have an identity on Radicle Link (hereinafter "Link"), or on Github, Gitlab, Twitter, or email address.
```

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
...
+In the same way, a registered project could also exist on Link, Github, or elsewhere.
+
+When relevant it is desirable to be able to link these identities together so that a registry use can be confident that they do in fact represent the same identity.
```
xla:
```suggestion
When relevant it is desirable to be able to link these identities together so that a user of the registry can be confident that they do in fact represent the same identity.
```

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
...
+The second builds on the first but is an optional extension to it.
+
+## General principle of operation
+
+The feature builds on the ability to refer to one identity from another one.
+
+For instance, it should be possible to insert a reference to a Registry user identity inside a Link user identity (by writing the Registry user ID inside the Link user metadata).
+It should also be possible to do the reverse: insert the Link user ID (their URN) inside the Registry user state.
+
+The key point is that only the legitimate user can do each of these operations.
```
xla:
```suggestion
The key point is that only a legitimate user can do each of these operations.
```

docs/external-attestations.md
xla: @radicle-dev/product

docs/external-attestations.md
```diff
@@ -0,0 +1,88 @@
...
+An easy solution to this problem is to only store a single link inside the registry, unambiguously referring to an external piece of metadata that in turn will contain all other external links.
+To guarantee absence of ambiguity this link should be a cryptographic hash of the external metadata.
+
+A practical approach to this would be storing the Link identity on the Registry, and putting every other identity references inside the Link identity metadata, reducing the registry state requirements to a single link, which is acceptable.
+Since Link identities are versioned with cryptographic hashes the system would be unambiguous, as required.
+
+### Data format and verification
+
+Let's say we decide to store the Link user identity in the Registry user state; we should then decide how to store it.
+In Link identities are represented by Radicle URNs, as described in the (identity resolution)[https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/identity_resolution.md] and (identities)[https://github.com/radicle-dev/radicle-link/pull/248] specifications.
```
xla:
```suggestion
In Link identities are represented by Radicle URNs, as described in the [identity resolution](https://github.com/radicle-dev/radicle-link/blob/master/docs/rfc/identity_resolution.md) and [identities](https://github.com/radicle-dev/radicle-link/pull/248) specifications.
```

xla (changes requested):

docs/external-attestations.md
massimiliano-mantione: :facepalm: :ok_hand:

kim:
I’ve been chatting with @geigerzaehler about the topic, and I think I can now formulate what I think makes this less practical than it could be:

The idea, iiuc, is to simply publish a claim of identity on both sides. It is true that this proves the link, but it also means that, in order to verify that proof, one **must** verify both sides (ie. run a light client _and_ obtain the `link` identity chain and run the verification procedure).

By contrast, consider a scheme similar to how keybase proofs work: you add a claim of your chain identity to the `link` document, sign that, and then publish the signature over that revision along with the public key used on-chain (32 + 64 = 96 bytes). Now, I can verify that you owned that key at this point in “time” _without even looking at the `link` side_.

This is valuable in the obvious way: say @giverofrad sends me a message on telegram saying “yo I’m @lftherios on radicle, track me at lftherios@hwdeafbeef... to receive 10 RAD!”, I can discard this immediately without establishing the trust relationship implied by tracking on `link`.

Once I’ve established a tracking relationship, verifying updates on `link` allows me to get notified of key revocations without incurring transaction fees. I can also extend trust to additional key delegations once I have verified one anchor.

As usual, the economics are what determines whether this is a viable product or not: if the `link` identity, like the keybase “sigchain”, is the canonical side, blockchain anchoring can be offered for free — just like keybase, some entity would just aggregate all `link` identities in a single merkle tree (ie. merge all known `rad/id` branches into a single one), and anchor the root on-chain. Note that there’s nothing “centralised” about this, it’s just a service anyone could run, even charge a small fee for it.

If, on the other hand, the `registry` identity is canonical, I pay a fee every time I add a new proof to an external system (because I have to include the claim on-chain).

/cc @cloudhead

cloudhead:
> By contrast, consider a scheme similar to how keybase proofs work: you add a claim of your chain identity to the link document, sign that, and then publish the signature over that revision along with the public key used on-chain (32 + 64 = 96 bytes). Now, I can verify that you owned that key at this point in “time” without even looking at the link side.

I'm trying to understand this:

```rust
let link_key = rad_link_keypair()
let identity = self()
let identity = identity { regitry_id = "0x123456", rev = identity.rev + 1 }
let signature = sign(hash(identity), link_key.secret)

// publish my signature over the revision containing the registry id
submit_tx(AnchorTx { rev = identity.rev + 1, sig = signature, key = link_key.public })
```

What did I (the tx author) prove?

I think I'm missing something in the mechanics of how this works. How do I verify that the key actually controls this identity without replaying the link history?

cloudhead:
Thinking about this more outloud..

What's the difference between how keybase does it, and our initial solution:

Radicle:
```
chain: link-id = abcdef
link: chain-id = 0x1234
```

Keybase
```
chain (twitter): signature(hash(abcdef), abcdef)
link: chain-id: 0x1234
```

What's the difference? Well with Keybase, I can verify just by looking at twitter, that the twitter handle owns the `link` key (`abcdef`).

So I can see an immediate benefit here, since the proof only requires looking in one place. But this is *given that I already know the key and trust the key (`abcdef`)*.

Now this is where I get stuck: how does this help us in practice, when the key controlling an identity can change?

kim:
@cloudhead sorry, missed your follow-up

> given that I already know the key and trust the key (abcdef)

Well, one side is known and trusted in either case, no? The point is just to get a proof that the key of the _other_ side is also held.

> how does this help us in practice, when the key controlling an identity can change?

The `link` identity is a "sigchain", so if one revision was checkpointed, you can assert that its _current_ set of keys is still linked to the registry identity by running the verification procedure from the checkpoint to the current tip. That's a no-op if the revisions match, ie. the chain anchor is refreshed. 

So yes, if you want to verify a signature in `link`, you may need the `link` "sigchain". If you want to verify that a `link` identity is linked (sic) to a registry identity, you look at the `link` doc and ask the registry for a signature made by one of they keys in the `link` identity's history (you know the `registry_id`). 

If, however, you only know the registry side -- yeah, what's missing is that the stable identifier (the root revision) also needs to be stored on-chain. That's an extra 32 bytes, but never changes.

NunoAlexandre:
No longer applicable, closing.

PR #592, opened 2020-07-15
Name 'amount' the Balance field of Transfer txs

### LINKED ISSUE #591: Make transfer message fields consistent
Reported by the upstream, one of the transfer tx message field for the balance is `amount` while the other one is `value`.

NunoAlexandre:
Closes #591


core/src/message.rs
```diff
@@ -213,33 +213,33 @@ pub struct SetCheckpoint {
...
 ///
-/// The org account must have a balance of at least `value`.
+/// The org account must have a balance of at least `amount`.
 ///
 #[derive(Decode, Encode, Clone, Debug, Eq, PartialEq)]
 pub struct TransferFromOrg {
     pub org_id: Id,
     pub recipient: AccountId,
-    pub value: Balance,
+    pub amount: Balance,
```
CodeSandwich, line 222: Shouldn't `Balance` be renamed then too?

NunoAlexandre: Don't think so. `Balance` is the type, `amount` is the field. `Amount` is "the amount being transferred", `Balance` represents a an amount in a specific currency. If you would like to discuss that further feel free to follow it up on a new issue, here I am just really concerned with making these fields consistent between themselves and with the CLI :)

NunoAlexandre:
@CodeSandwich, I have bumped the `impl_version` and ran `./scripts/rebuild-chain-specs impl`, pushed in https://github.com/radicle-dev/radicle-registry/pull/592/commits/052aec415890215fa928881e56e4bed742e31773.

Will wait for your signal to merge.

runtime/Cargo.toml
```diff
@@ -3,7 +3,7 @@ edition = "2018"
 name = "radicle-registry-runtime"
 description = "Substrate chain runtime for the Radicle Registry"
 authors = ["Monadic GmbH <radicle@monadic.xyz>"]
-version = "0.16.0"
+version = "1.16.0"
```
CodeSandwich, line 6: Wrong way :scream: 
```suggestion
version = "0.16.1"
```

NunoAlexandre: :man_facepalming: what was I thinking about

Cargo.lock
```diff
@@ -3805,7 +3805,7 @@ dependencies = [
 
 [[package]]
 name = "radicle-registry-runtime"
-version = "0.16.0"
+version = "1.16.0"
```
CodeSandwich, line 3808: The lock needs a fix too. I think that it's the last place.
```suggestion
version = "0.16.1"
```

NunoAlexandre:
Thanks for the patience, @CodeSandwich!

PR #595, opened 2020-07-17
client: Export Ss58Codec

NunoAlexandre:
Tell me if I'm mistaken, but this seems to be the Rusty way of doing it. I'd prefer to have a `parse_ss58_address(&str) -> AccountId` function that would encapsulate that logic instead of exposing the internals and have third clients using the underlying trait to get it right. What do you recon @CodeSandwich?

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -344,6 +344,11 @@ impl ClientT for Client {
     }
 }
 
+/// Parse an [AccountId] from str expected to be in the ss58 format, failing otherwise.
+pub fn parse_ss58_address(address: &str) -> Result<AccountId, sp_core::crypto::PublicError> {
+    sp_core::crypto::Ss58Codec::from_ss58check(address)
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,7 @@ Upcoming
 
 ### Addition
 
+* client: Add `parse_ss58_address` to parse an `AccountId` from a ss58 formatted string
 * client: Add `account_exists` to check whether an account exists on chain
 * client: Add `fn get_id_status` to get the status of an id (available, taken or retired)
 * Support user project registration


CodeSandwich:
> I'd prefer to have a `parse_ss58_address(&str) -> AccountId` function

Yeah, me too, it would hide an otherwise useless trait, which we don't control from our API. Why does this PR expose this trait instead?

NunoAlexandre:
> > I'd prefer to have a `parse_ss58_address(&str) -> AccountId` function
> 
> Yeah, me too, it would hide an otherwise useless trait, which we don't control from our API. Why does this PR expose this trait instead?

Ok great! Because I thought I would be asked to expose the trait instead of doing that :b will change

client/src/lib.rs
```diff
@@ -357,3 +357,8 @@ mod test {
         is_sync_send(Client::new_emulator().0);
     }
 }
+
+/// Parse an [AccountId] from str expected to be in the ss58 format, failing otherwise.
+pub fn parse_ss58_address(address: &str) -> Result<AccountId, sp_core::crypto::PublicError> {
+    sp_core::crypto::Ss58Codec::from_ss58check(address)
+}
```
CodeSandwich, line 364: The tests should be the last thing in a file, people won't be able to find a production function here.

client/src/lib.rs
```diff
@@ -357,3 +357,8 @@ mod test {
         is_sync_send(Client::new_emulator().0);
     }
 }
+
+/// Parse an [AccountId] from str expected to be in the ss58 format, failing otherwise.
+pub fn parse_ss58_address(address: &str) -> Result<AccountId, sp_core::crypto::PublicError> {
```
CodeSandwich, line 362: I wonder if the error type here is right. All the APIs exposed here operate on a single error type: `Error`. This is very convenient for the consumers, because it requires only one error routine for all client errors. Maybe we should add a new variant to `Error` for SS58 decoding failure?

NunoAlexandre: I thought of just passing the underlying error otherwise the caller of this fn needs to handle all other client `Errors` even though they can not occur here. And this is not a Client fn per se, it's just a utility provided by the crate.

CodeSandwich: What's the use case of this function? I'd assume that the caller is already a user of this crate and can support `Error`s and would be happier if he didn't have to support one more error type.

NunoAlexandre: not really. the use case is to deserialize an api request param in the upstream, there is no broader `error` handling. If it's error we return 400, otherwise we process it. We could also makes this fn return `Option` as we don't need the granular error info but I'd rather leave it as is.

CodeSandwich: So it's going to be used in the upstream as part of running the client? Then it would make a whole lot sense to have `Error` here. Imagine upstream function parsing the ID and sending a request: if these steps produce different errors, the returned error type will be super messy. If handling of decoding failure on upstream side should be similar to handling a bad request coming from the node, it would be convenient to have one error flow.

NunoAlexandre: > So it's going to be used in the upstream as part of running the client?

> No. I just really need a function that offers this functionality without having to add the sp_core trait to the upstream.

It's me :b see https://github.com/radicle-dev/radicle-upstream/pull/681/commits/d933b7563bfa4da637c79af0f07176c99a897541#diff-00f1ac2eadf7c48a5661bac5f74a7961R125 to see this code in action.

Issue #585, opened 2020-07-10
Eliminate checkpoints


geigerzaehler:
As @cloudhead confirmed the checkpointing system does not have a use in the current design for the product and it will not be used in the form that it is implemented in. It is yet unclear how a checkpointing system will look like and when it will be needed. Since checkpointing requirements complicate things unnecessarily and provide a potential attack vector (especially for storage attacks) we want to ged rid of them.

CodeSandwich:
If we have just the initial state hash, we can still prove genuineness of the repository version. If forks occur, we can check which one has the right initial state and contains a commit signed by the owner with the newest timestamp: that's our canonical branch.

That's why in my opinion we should keep the initial state hash in the project regardless of the volatile checkpoints removal. I think that we should add a special field to the project storage for it. An alternative would be using the opaque blob, which is already present there, but in my opinion the initial hash is too important from security perspective to accept registration of a project without it.

NunoAlexandre:
@CodeSandwich is your suggestion to essentially have an `initial_checkpoint: CheckpointId` field in the `Project`?

geigerzaehler:
> That's why in my opinion we should keep the initial state hash in the project regardless of the volatile checkpoints removal.

Until we have a properly designed solution for attestation that includes link and upstream I would suggest we not implement anything related to that. The probability that what we implement now is not sufficient or has to be modified later is very high. This is exactly how we ended up in this situation where we remove the checkpoints.

geigerzaehler:
@lftherios mentioned that @kim had ideas about using a similar system for identites or attestation. Nevertheless I’d like to move forward with this since @kim is out. This feature is unused has [some problems](https://github.com/radicle-dev/radicle-registry/issues/474) and I doubt that we can easily repurpose it without major adjustment. In the spirit of yesterdays discussion we should only have code in the runtime that we know has a purpose.

lftherios:
> @lftherios mentioned that @kim had ideas about using a similar system for identites or attestation. Nevertheless I’d like to move forward with this since @kim is out. This feature is unused has [some problems](https://github.com/radicle-dev/radicle-registry/issues/474) and I doubt that we can easily repurpose it without major adjustment. In the spirit of yesterdays discussion we should only have code in the runtime that we know has a purpose.

Good with me.

PR #598, opened 2020-07-20
Drop checkpoints

NunoAlexandre:
`WIP`

Closes https://github.com/radicle-dev/radicle-registry/issues/585


runtime/src/registry.rs
```diff
@@ -350,14 +341,8 @@ pub fn org_has_member_with_account(org: &state::Orgs1Data, account_id: AccountId
...
-    pub enum Event {
-        MemberRegistered(Id, Id),
-        OrgRegistered(Id),
-        OrgUnregistered(Id),
-        ProjectRegistered(ProjectName, ProjectDomain),
-        UserRegistered(Id),
-        UserUnregistered(Id),
-    }
+    /// We don't make use of Events
+    pub enum Event {}
```
geigerzaehler, line 345: We can remove this completely

NunoAlexandre: Yeah. I get over 1k compilation errors when doing that so I gave this smaller step first. I have this parked in favour of #600.

PR #600, opened 2020-07-21
Implement registration fees

### LINKED ISSUE #599: Implement Registration fees
### User journey

When registering a user or an org in the registry, the tx author will be charged for a "registration fee" on top of the ordinary tx fee. This registration fee covers for the weight that such a transaction causes on the chain stores. If the author does not have enough funds, an error will be returned.

### Specification

1. The author will pay for a "registration fee" for `RegisterUser` and `RegisterOrg`
2. The fee is a static value defined in the runtime by us, equal for both user and org registration, and independent of the contents of those transactions.
3. It's a one time, 100% burn fee.

NunoAlexandre:
Closes #599 

**Note:** This could have been done with less code but introducing sparse fees in the codebase seems 
dangerous and the antithesis of self-documenting. Having a type describing the fees we impose and their 
properties in a unified place means that a reader can know everything there is to know about fees 
in the registry by checking the `fees` module. This also makes some code to be easier to reuse.


CodeSandwich:
> introducing sparse fees in the codebase seems dangerous and the antithesis of self-documenting

What's the "sparse" alternative?

NunoAlexandre:
> > introducing sparse fees in the codebase seems dangerous and the antithesis of self-documenting
> 
> What's the "sparse" alternative?

To have the tx fee being withdrawn from the signed extension and the registration fee being withdrawn from the handlers (as they are now) without any common touch between these two fees. In this case, the common ground is the enum self-documenting the two types of fees we now support and how they are handled: how they are paid, the technical withdrawn reasons they report, etc.

runtime/src/fees/mod.rs
```diff
@@ -27,12 +27,43 @@ use sp_runtime::transaction_validity::{
...
 mod payment;
 
-/// The minimum fee required for a transaction to be considered
-/// for inclusion in the blockchain.
-pub const MINIMUM_FEE: Balance = 1;
+pub use payment::pay;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum Fee {
+    /// A transversal transaction fee, defined by the author.
```
geigerzaehler, line 34: What does “transversal” mean?

runtime/src/fees/payment.rs
```diff
@@ -13,43 +13,49 @@
...
-    drop(burn);
-
-    // The block author is only available when this function is run as part of the block execution.
-    // If this function is run as part of transaction validation the block author is not set. In
-    // that case we don’t need to credit the block author.
-    if let Some(block_author) = store::BlockAuthor::get() {
-        crate::runtime::Balances::resolve_creating(&block_author, reward);
-    }
+pub fn pay(author: &AccountId, fee: Fee) -> Result<(), DispatchError> {
+    match fee.clone() {
```
geigerzaehler, line 31: I don’t think a clone is necessary here

runtime/src/fees/mod.rs
NunoAlexandre: it means that it applies to all transactions.

runtime/src/fees/mod.rs
```diff
@@ -27,12 +27,43 @@ use sp_runtime::transaction_validity::{
...
+    /// The higher the fee, the higher the priority the tx receives.
+    TxFee(Balance, Call),
+
+    /// A mandatory fee, defined by us, meant to cover the storage costs of
+    /// registering a user or an org and its assets on chain.
+    RegistrationFee,
```
geigerzaehler, line 40: `Fee::RegistrationFee` and `Fee::TxFee` seem redundant. I’d prefer `Fee::Registration` and `Fee::Tx`.

NunoAlexandre:
> I’d prefer a simpler version. The `Fee` enum and the generic `pay` function look like unnecessary abstractions to me. `fees::pay` is called from two places (`registry::pay_registration_fee` and `PayTxFee::validate`). Both call sites construct the `Fee` variant which is immediately destructed again in `fees::pay`. Moreover `Fee` only implements two methods so that `withdraw` can work with any fee.
> 
> I think it would be much simpler to just have a `pay_registration_fee` in the `fee` module that holds all that logic and keep it separate from the transaction fee. This would still ensure that everything related to fees is in the `fee` module.

Ok!

NunoAlexandre:
@geigerzaehler simplified the solution, could you have another look? :seedling:

runtime/src/fees/mod.rs
```diff
@@ -27,12 +26,16 @@ use sp_runtime::transaction_validity::{
...
-/// The minimum fee required for a transaction to be considered
-/// for inclusion in the blockchain.
-pub const MINIMUM_FEE: Balance = 1;
+pub use payment::{pay_registration_fee, pay_tx_fee};
+
+/// The minimum acceptable tx fee
+pub const MINIMUM_TX_FEE: Balance = 1;
+
+/// The registration fee
+pub const REGISTRATION_FEE: Balance = 10;
```
CodeSandwich, line 35: Is it the same thing as `REGISTRATION_FEE` in `runtime/src/registry.rs`?

NunoAlexandre: indeed. it must have been re-introduced in a rebase, dropped it.

PR #601, opened 2020-07-22
Build runtime on the CI

### LINKED ISSUE #492: Build runtime Wasm artifacts on CI

CodeSandwich:
Fixes #492 except runtime source code hash validation.


scripts/rebuild-runtime-cache
```diff
@@ -0,0 +1,46 @@
...
+impl_version="$(echo "$runtime_version" | awk -F . '{print $3}')"
+prev_spec_version=$(($spec_version - 1))
+latest_runtime_name="v${spec_version}_${impl_version}.wasm"
+latest_spec_latest_impl_name="dev_v${spec_version}_latest.json"
+latest_spec_first_impl_name="dev_v${spec_version}_0.json"
+prev_spec_latest_impl_name="dev_v${prev_spec_version}_latest.json"
+
+# fetch or build the latest runtime
+new_runtime=0
+gsutil cp "$STORAGE_PREFIX$latest_runtime_name" "$LATEST_RUNTIME_PATH" || new_runtime=1
```
geigerzaehler, line 29: I think it would make sense to use curl here. This would allow devs to use this without having to install `gsutil`.

scripts/rebuild-runtime-cache
```diff
@@ -0,0 +1,46 @@
...
+if [ "$new_runtime" == "1" ]
+then
+  echo "building the latest runtime"
+  ./scripts/build-runtime-wasm "$LATEST_RUNTIME_PATH"
+fi
+
+echo "building latest impl of current spec"
+RUST_LOG=off ./target/release/radicle-registry-node build-spec --chain dev > "$RUNTIME_CACHE$latest_spec_latest_impl_name"
```
geigerzaehler, line 37: We need to rebuild `radicle-registry-node` if the latest runtime wasm was changed.

DEVELOPING.md
```diff
@@ -153,32 +143,41 @@ affect the semantics.
...
 `VERSION` and the patch version of the crate.
 
-After all the changes are applied it must also update the CI artifacts:
+### Semantic updates
+
+Semantic changes must increment the `VERSION.spec_version` and
+`VERSION.transaction_version` fields and the `runtime` crate minor version and
+reset the `impl_version` field and the crate patch version to `0`.
+
+### Testing after runtime update
```
geigerzaehler, line 152: Can we move the content of this section to “Running checks and tests”? We should also point out that the CI runs tests against three versions of the chain spec.

ci/run
```diff
@@ -128,6 +107,26 @@ declare -r artifact_prefix="https://builds.radicle.xyz/radicle-registry/${artifa
   done
 } | buildkite-agent annotate --context node-binary --style success
 
+if [[ "${BUILDKITE_BRANCH}" == "master" ]]
+then
+  echo "--- Upload runtime and specs"
+
+  STORAGE_PREFIX='gs://radicle-registry-runtime/'
+  awk '$1 == "RUNTIME_STORAGE_GCP_CREDENTIALS:" {$1 = ""; print $0; exit}' .secrets | tr -d \' > gcp_cred.json
```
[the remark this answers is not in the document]
CodeSandwich: I don't understand this solution, could you break it down? What does `.source` command do? `secrets` file doesn't exist, did you mean `.secrets`? `RUNTIME_STORAGE_GCP_CREDENTIALS` isn't a variable, it's just a name of a field in decoded from `.secrets`?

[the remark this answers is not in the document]
geigerzaehler: Sorry, I had a typo in the suggestion. `source .secrets` interprets `.secrets` as a bash script. Since the file includes one variable assignment per line the `RUNTIME_STORAGE_GCP_CREDENTIALS` variable is then available.

geigerzaehler:
Could you also open a PR on the infra repo that describes the infrastructure reuqired for this?

CodeSandwich:
> infrastructure reuqired for this

You mean the GCP storage? It's the only thing outside of registry touched by these changes.

ci/run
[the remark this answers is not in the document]
CodeSandwich: I don't think that it's a good solution, `.source` is a YAML file, its first line says `data: |` and the key-values are written as `NAME: VALUE`, which don't match bash's `NAME=VALUE`.

geigerzaehler:
> You mean the GCP storage? It's the only thing outside of registry touched by these changes.

Yes, the GCS bucket and the service account that can write to it and is used by the build.

scripts/rebuild-runtime-cache
CodeSandwich: We do, in `ci/run`. But you've shown a different problem: the CI reuses the old runtime from previous build to create a spec :D Fixed.

DEVELOPING.md
CodeSandwich: Fixed

scripts/rebuild-runtime-cache
CodeSandwich: Fixed

ci/run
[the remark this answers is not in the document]
CodeSandwich: Fixed, no more secrets in file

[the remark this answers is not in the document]
geigerzaehler: > I don't think that it's a good solution, `.source` is a YAML file, its first line says `data: |` and the key-values are written as `NAME: VALUE`, which don't match bash's `NAME=VALUE`.

I missed that. But this will break `ci/deploy` which relies on the old format. We need to fix that too but I’d prefer the old format: Yaml with one key per secret. This aids readability and makes it easier to extract.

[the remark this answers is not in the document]
CodeSandwich: This PR doesn't change the format of the secret. It's still YAML, just with 1 more entry.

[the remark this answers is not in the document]
CodeSandwich: Oh no, `ci/deploy` expects `.secrets` to be a bash script, NOT a YAML, but on `master` it IS YAML! I'll fix that!

NunoAlexandre:
I had a look and don't have anything to add on top of Thomas' comments. Dope to move this to the CI, thanks for the effort @CodeSandwich, will be great to have this on master :1st_place_medal:

ci/run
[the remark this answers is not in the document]
CodeSandwich: Nope, `master` has a perfectly fine secrets file, it was just me misusing sops. Anyway, fixed, now `.buildkite/secrets.yaml` decodes to a valid dotenv file.

CodeSandwich:
@geigerzaehler could you take another look? I've added logic to skip fetching a spec when the current version of a runtime causes a historical compatibility break in the client. Not very elegant, but it works.

scripts/rebuild-runtime-cache
```diff
@@ -0,0 +1,72 @@
...
+latest_runtime_name="v${spec_version}_${impl_version}.wasm"
+latest_spec_latest_impl_name="dev_v${spec_version}_latest.json"
+latest_spec_first_impl_name="dev_v${spec_version}_0.json"
+prev_spec_last_impl_name="dev_v${prev_spec_version}_latest.json"
+
+# params: <FILE_NAME_IN_BUCKET> [<FILE_NAME_IN_CACHE_IF_DIFFERENT>]
+download_cache_file()
+{
+  out_file="$runtime_cache${2-$1}"
+  http_code=`curl -s -o "$out_file" -w "%{http_code}" "https://storage.googleapis.com/radicle-registry-runtime/$1"`
```
geigerzaehler, line 28: A common pattern is `curl -fS` wchich avoids the manual code check and file cleanup

scripts/rebuild-runtime-cache
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+skip_prev_spec=0
+for breaking_spec in "17"
+do
+  if [ "$spec_version" == "$breaking_spec" ]
+  then
+    echo "spec version $spec_version is designed for no backward compatiblity, \
+skipping fetching spec file for spec $prev_spec_version"
+    skip_prev_spec=1
```
geigerzaehler, line 64: You could just `exit 0` here and avoid using `$skip_prev_spec`.

scripts/rebuild-runtime-cache
CodeSandwich: Nice! Fixed

scripts/rebuild-runtime-cache
CodeSandwich: Good idea. I was trying to make the flow non-surprising if someone tried to add more functionality in the end of this script. But it's a heavy solution to a non-existent problem and the script is really easy to follow anyway. Fixed.

PR #606, opened 2020-07-28
runtime: fix user unregistration 

### LINKED ISSUE #584: Users can be unregistered even if they have projects
At the moment a user (unlike an org) can be unregistered even if they still have projects registered under them. We should require that a user  can be unregistered only if they have no projects.

NunoAlexandre:
Closes #584


CodeSandwich:
We don't need to store `runtime/latest.wasm` in the repository anymore

runtime/src/registry.rs
```diff
@@ -439,10 +440,8 @@ fn ensure_id_is_available(id: &Id) -> Result<(), RegistryError> {
 // TODO(xla): This is a naive first version of the check to see if an account is
 // already associated to a user. While fine for small dataset this needs to be reworked
 // in the future.
-pub fn get_user_id_with_account(account_id: AccountId) -> Option<Id> {
-    store::Users1::iter()
-        .find(|(_, user)| user.account_id() == account_id)
-        .map(|(id, _)| id)
+pub fn get_user_with_account(account_id: AccountId) -> Option<(Id, state::Users1Data)> {
+    store::Users1::iter().find(|(_, user)| user.account_id() == account_id)
```
CodeSandwich, line 444: Maybe the changes would be less sweeping if there was a new function `get_user_with_account`, which was internally used by `get_user_id_with_account`? The API of `get_user_with_account` seems a bit too verbose, in most places only the ID is needed.

NunoAlexandre: Agree, fixed it!

